PR #1, opened 2016-04-27
fix missing favicon from npm package

ryanflorence:
now the favicon makes it to the npm package and the server actually runs

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,8 @@
     "README.md",
     "LICENSE",
     "dist",
-    "index.js"
+    "index.js",
+    "assets"
   ],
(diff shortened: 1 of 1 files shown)


calebmer:
Thanks! Totally my bad 😊 👍

calebmer:
I've started a [Gitter room](https://gitter.im/calebmer/postgraphql) if you want to talk about your experiences with PostGraphQL. If you'd like to contribute I also want to adopt an open-open source governance model so core collaborators have direct commit rights on the repo (only makes sense as PostGraphQL is definitely a critical dependency).

PR #13, opened 2016-04-28
Cast a.attnum to int

rattrayalex:
Addresses https://github.com/calebmer/postgraphql/issues/12

### DIFF
--- a/src/postgres/getCatalog.js
+++ b/src/postgres/getCatalog.js
@@ -67,7 +67,7 @@ const getRawColumns = memoize(client =>
       left join pg_catalog.pg_constraint as cp on
         cp.contype = 'p' and
         cp.conrelid = a.attrelid and
-        cp.conkey @> array[a.attnum]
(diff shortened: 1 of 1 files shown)


calebmer:
Per chance, what PostgreSQL version are you using?

rattrayalex:
9.4.5

calebmer:
Ok cool, I've been testing on 9.5 myself 😊

I definitely want to merge and release this, I just need to setup CI first, so that might take me a bit.

calebmer:
I decided to merge and test locally so you wouldn't have to rebase my CI commit and got the following error: `operator does not exist: smallint[] @> integer[]`. I rolled back, so could you submit the PR again so it can hit CI and we can debug together?

rattrayalex:
Huh, interesting. Yes, happy to - unsure if I'll be able to w/in 24h.

First guess is that casting the other side of the expression go int might
do the trick...

PR #24, opened 2016-04-29
Typo in anchor

jazzido:
Correct typo in anchor link


calebmer:
Thanks for the catch 🎉

PR #39, opened 2016-05-04
Postgres is now 20 years old. :)

decibel:


calebmer:
Nice! I was going off if the [about](http://www.postgresql.org/about/) docs on the website 😊

Issue #40, opened 2016-05-05
Authentication endpoint to generate JWT tokens


pvdvreede:
Great project!

I am having trouble working out where I generate the JWT token. Your docs provide a really good description of how you can see the details of a passed in token, so I assume that you are meant to create a JWT token in another service and just have the same secret as you have in postgraphql. Is this correct?

If so, is it possible to add an endpoint to postgraphql to allow users to authenticate with a postgresql login role and password and have postgraphql generate and return a JWT token for use with graphql queries?

Thanks.

calebmer:
We are still working out the best way to create tokens in PostGraphQL, if you have any ideas, we would love to here them! 😊

For now it's recommended to create a single HTTP endpoint where you can (temporarily) manually create tokens.

We'll probably have a solution soonish with the implemention of procedure execution, so stay tuned! (see my notes on #31)

rahult:
This is an awesome project. Discovered it today and we already have a working example for implementation.

Is it possible to implement something like PostgREST

JWT tokens - http://postgrest.com/admin/security/
Users - http://postgrest.com/examples/users/

calebmer:
@rahult Authentication currently works very similar to PostgREST (actually hoping they will rename where they mount claims to be compliant with [this](https://github.com/calebmer/postgraphql/blob/master/docs/pg-jwt-spec.md) spec so we can be compatible in that regard).

I will also probably do authentication that way, it really seems like the best idea. That is blocked by the implementation of procedures though so stay tuned!

I'm going to leave this open until we have an authentication implementation.

pvdvreede:
Thanks @calebmer! For what its worth, I do like that I can create my own authentication service to create JWT tokens for any special needs or authentication systems I might have, so it would be good to continue to be able to do that even once authentication is built in. It would just be nice to have the option of a built in authentication as well. So hopefully you can architect it to keep both options available.

Good luck getting the stored proc stuff sorted!

calebmer:
Thanks! Yeah, opt in authentication is super important for us 😊

meglio:
Now that procedure support has been released, chances are there will be an authentication point done any time soon as well?

calebmer:
@meglio I’m going to look into adding this over the weekend.

meglio:
This is awesome! Are you going to also handle [replay attacks](https://github.com/auth0/node-jsonwebtoken/issues/36) in any way?

calebmer:
[...]

Issue #31, opened 2016-05-01
forum example: how to use search operators


mrbarletta:
thanks for the kool project! looks primising.

I was wondering how to use the search operators (e.g. like givenName: '%partOfName%')

Do we have to extend the exposed methods? is that even possible? 

thanks!

calebmer:
I haven't thought of a good way to represent complex operators in the GraphQL type system. However, I am planning to finish support for calling PostgreSQL procedures soon. You can actually see the branch [here](https://github.com/calebmer/postgraphql/tree/feat/procedures) where I currently have mutative procedures implemented.

Relevant parts from that branch include [this](https://github.com/calebmer/postgraphql/blob/feat/procedures/examples/forum/schema.sql#L117-L123) new example procedure and [this](https://github.com/calebmer/postgraphql/blob/feat/procedures/tests/integration/fixtures/procedure-mutation.graphql) passing test showing off procedures that perform mutations.

gilesbradshaw:
have you any thoughts about how the existing graphql schema produced by postgraphql could be extended to add aditional parameters?

calebmer:
Not too many yet. After the procedure stuff gets implemented and we have a good story for authentication, I want to work on the node API side to allow people to extend their schema however they want.

calebmer:
@mrbarletta check out version `1.3.0-0` and #41 and tell me how that works for you 👍

calebmer:
Procedure support is released in `1.3.0`. I wrote [this](https://github.com/calebmer/postgraphql/blob/master/docs/advanced-queries.md) new article with your issue in mind 😉

If it's inadequate in explaining the new feature, please tell me. Also you can follow #47 where simple free-form filtering using conditions will be possible.

PR #41, opened 2016-05-06
Add procedure support

calebmer:
This PR adds support for PostgreSQL procedure execution. This will lead to a lot of awesome things like custom search functionality (#31) and authentication (#40) to name a few things that currently have open issues. This also adds a PostgreSQL native extension mechanism.

This PR is almost done, it has support for mutative procedures and query procedures, all that's missing right now is support computed column procedures.

There are a whole bunch of edge cases I'm probably missing so if you guys could try this out in your projects and report back the results that would be awesome 👍. I've published a prerelease on npm at version `1.3.0-beta.0`.

Relevant files to get an understanding of how this include:
- [Example procedures](https://github.com/calebmer/postgraphql/blob/feat/procedures/examples/forum/schema.sql#L78-L157)
- [Mutation test](https://github.com/calebmer/postgraphql/blob/feat/procedures/tests/integration/fixtures/procedure-mutation.graphql)
- [Query test](https://github.com/calebmer/postgraphql/blob/feat/procedures/tests/integration/fixtures/procedure-query.graphql)


calebmer:
Ok, the code for this is pretty much done! 🎉

Computed columns work great. Before I merge and publish this I'm going to write for comments from anybody, and write some docs.

If anyone is interested in trying this out, the code is published under `1.3.0-beta.1`.

calebmer:
One thing I haven’t decided on, is should procedures like `forum_example.search_posts` return a connection like `postNodes` or not? If so a user could paginate as normal over a set returned by the procedure which could be nice.

I'm also thinking about deprecating arguments testing equality in fields like `postNodes` (so deprecate the `headline`, `topic`, etc. arguments on that field). Thoughs?

burkhardr:
Re: whether procedures should return a connection: Yes, definitely, but how do you distinguish between procedures that behave like tables / return rows and others that are helper functions?

calebmer:
`returns setof` is the differentiator I think.

calebmer:
Ok, implementing connections from `setof` procedures was more difficult than I thought. I'm going to leave this open so people can review if they want and I can write docs. Let’s get this merged quickly so we can iterate 👍

calebmer:
Just finished writing documentation, so if you want to read to learn more about this feature/spell check take a look!

I'm super excited about this feature guys 😊

README.md
```diff
@@ -72,6 +72,43 @@ Can query relations like so:
...
+And queried through GraphQL like this:
+
+```graphql
+{
+  searchPosts(search: "Hello world", first: 5) {
+    pageInfo {
+      hasNextPage
+    }
+    totalCount
+    nodes {
```
burkhardr: Should this be `edges { node { headline, body } }`?

calebmer: `nodes` is a little simpler and less intimidating for the `README` I think.

calebmer:
Released in `1.3.0` 🎉

README.md
burkhardr: Ah, but returns the same? I've not checked the implementation yet

calebmer: Yep, it only returns the list, it's a shortcut if you're not interested in the cursors.

PR #46, opened 2016-05-11
Deprecate nodes filter

calebmer:
I want to deprecate being able to do queries like `postNodes(help: TOPIC)` which uses an equality test to filter the nodes. I feel like it adds too much noise to the query arguments list for very little gain, it especially gets weird when you want to start matching `body` or long form text fields. Procedures also are a better more flexible way to implement this feature on a case by case basis.

I feel like we should reconsider this feature and try to find a solution that works better with the GraphQL type system and allows us to use many different operators outside of just `=`.

### DIFF
--- a/src/graphql/createConnectionArgs.js
+++ b/src/graphql/createConnectionArgs.js
@@ -50,6 +50,7 @@ const createConnectionArgs = (table, ignoreColumnConditions = []) => ({
       'items will be in ascending order. `false` by default.',
     defaultValue: false,
   },
+  // DEPRECATED: Remove this entirely
   ...fromPairs(
     table
     .getColumns()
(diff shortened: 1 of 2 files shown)


calebmer:
Tell me what you think, I won't merge this without a response from users 👍

burkhardr:
I always considered the fact that you could create free-form queries on your client-side as required, so you could just add/remove query arguments as and when needed. If I understand this right, delegating any kind of filtering in queries to procedures would mean having to change those procedures with your client queries?

If so, I'd probably vote against.

artem-barmin:
Agree with @burkhardr. Main use case for auto-generated graphql schema - ability to create queries purely on client. 

I vote against

calebmer:
Alright, sounds reasonable. I do want a better way to filter things, but I figured with procedures it wasn't a high priority.

Issue #44, opened 2016-05-10
Embedded / Library Usage


grahamb:
Stemming from a [gitter discussion](https://gitter.im/calebmer/postgraphql?at=57324f9df36daf63798e718f). 

I'm wondering what @calebmer and others thoughts are surrounding using postgraphql in an embedded or library fashion. I'm working on an express app right now that will be querying an external graphql server for its user-facing functionality. Internally, the app has a very simple postgres DB for storing users, and I'd like to expose a graphql interface to it for building an admin dashboard. I've been able to make it work by basically copying the logic from [createServer](https://github.com/calebmer/postgraphql/blob/1388542088f2d7bb783ac245550127ab6764c91d/src/createServer.js#L40) and implementing it in an [Express route](https://github.com/sfu/student-dashboard/commit/73b2d9e3ffe95523ece5aa0c45cb3c1bf33a9161). 

One potential thing I could see being useful it the ability to specify the client. I'm using [knex](knexjs.org) everywhere else in the app, but for this portion I need to manually import pg and specify the client.

burkhardr:
Hey @grahamb. I'm in the exact same situation as you (use as a library, not being able to get the pg client out of knex / pass it into knex). There's 2 things I can immediately think about when it comes to using postgraphql as a library:

* documentation, documentation, documentation (on library usage, the rest is excellent)
* authorisation (at the moment, you'd have to manually wrap it in a transaction and set config vars)

hardchor:
Hey @grahamb. I'm in the exact same situation as you (use as a library, not being able to get the pg client out of knex / pass it into knex). There's 2 things I can immediately think about when it comes to using postgraphql as a library:

* documentation, documentation, documentation (on library usage, the rest is excellent)
* authorisation (at the moment, you'd have to manually wrap it in a transaction and set config vars)

calebmer:
Check out #48 and tell me if that's what you're looking for.

calebmer:
This has been added in `1.4.0` 🎉

avkonst:
Letting you know that I published starter project for backend development. It uses postgraphile as a library, based on Apollo Server 2.0 and leverages automated code generation and database schema migrations. Check it out here https://github.com/avkonst/graphql-postgraphile-typeorm-starter

PR #48, opened 2016-05-11
Add library usage

calebmer:
Addresses #44. The current state of this PR is the simplest translation of what already exists to a simple interface. @grahamb, @hardchor what other things would you like to see in this interface?

I wrote a documentation article explaining how to use the new interface.


calebmer:
Released in `1.4.0`! 🚀

burkhardr:
Sorry, super busy with work - this looks great! In a future addition, we could also document how to use this without connect-style middleware (i.e. `createGraphqlSchema`) - unless you'd prefer to leave this bit undocumented because it's hard to support?

calebmer:
If we go in the direction mentioned in #50 it may be simpler to support and document.

Issue #43, opened 2016-05-08
Relay mutation types support


msimulcik:
Hi, I'm trying to write Relay client backed by postgrqphql, but I'm having trouble with RANGE_ADD mutation type, which I want to use for adding a new person to the list of persons.

`InsertPersonPayload` type only contains `person: Person` and `clientMutationId: String` fields, but according to [relay documentation](https://facebook.github.io/relay/docs/guides-mutations.html#range-add) this mutation type requires a parent, a connection, and the name of the newly created edge from the  payload to be specified.

I'm new to GraphQL and Relay and maybe I'm missing something here, but it seems to me that with current mutation payloads it is not possible to define all Relay mutation types. I checked how [Reindex](https://www.reindex.io/docs/tutorial/adding-mutations/#addtodomutation) solves this and I think we need to add a field containing newly inserted edge (personEdge) and a `viewer` field that will act as a single entry node allowing us to access connections (like personNodes) that are currently available only in Query type. The single entry node field was already mentioned in [#38](https://github.com/calebmer/postgraphql/issues/38) but in the context of queries.

calebmer:
Could you draft out a quick schema of what you'd need? This shouldn't be too hard to add, I just want to see exactly what you have in mind.

Maybe this is related to #38? (/cc @hardchor)

msimulcik:
Sure, I'm going to put something together after work. In the meantime you can check doc pages on [mutations](https://www.reindex.io/docs/graphql-api/mutations/) and [queries](https://www.reindex.io/docs/graphql-api/queries/) from Reindex project, which is very similar to postgraphql. While Postgraphql generates the GraphQL schema (queries and mutations) based on the DB schema, Reindex generates it based on a custom JSON schema, that describes entities and relations. I think it could be a good inspiration for postgrqaphql as they have already solved these issues with the structure of query and mutations payload types.

msimulcik:
I will demonstrate the problem on simple todo list. First create a todo table with few rows
```sql
create table todo(todo_id serial primary key, todo_text varchar(255), complete boolean);
insert into todo(todo_text, complete) values ('foo', false), ('bar', false);
```
Now I want to list all todos in Relay
```javascript
// App.js
class App extends React.Component {
  render() {
    return (
      <div>
        <h1>Todo list</h1>
        <ul>
          {this.props.todos.edges.map(edge =>
            <li key={edge.node.id}>{edge.node.todoText}</li>
          )}
        </ul>
      </div>
    );
[...]

PR #49, opened 2016-05-11
Add Relay viewer field

### LINKED ISSUE #38: Single entry node
Relay (currently) doesn't support connections / array nodes without arguments as entry nodes (see https://github.com/facebook/relay/issues/112 for the rationale).

In terms of the classic use case for relay, where it gets used to resolve the data structure for a given page, that kind of structure (`viewer`) actually makes sense, as the `viewer` would be representative of the page you're currently on and would then resolve data starting from that node.

So the current approach for most relay-compatible GraphQL endpoints is to implement a single(ton) `viewer` node (which sits alongside the `node` node as the only entry nodes), onto which all other nodes get joined.

calebmer:
Adds a Relay viewer field to the root query type and to all of the mutation payloads. @hardchor, @msimulcik is this what you’re looking for?

Closes #38 and maybe #43.

### DIFF
--- a/src/graphql/createViewerField.js
+++ b/src/graphql/createViewerField.js
@@ -1,6 +1,6 @@
 import { memoize } from 'lodash'
 import { GraphQLObjectType, GraphQLNonNull } from 'graphql'
-import createQueryFields from './createQueryFields.js'
+import createQueryFields from './query/createQueryFields.js'
 
 const createViewerField = memoize(schema => ({
   type: new GraphQLNonNull(new GraphQLObjectType({
--- a/src/graphql/createViewerField.js
+++ b/src/graphql/createViewerField.js
@@ -1,20 +1,35 @@
 import { memoize } from 'lodash'
-import { GraphQLObjectType, GraphQLNonNull } from 'graphql'
+import { GraphQLObjectType, GraphQLNonNull, GraphQLID } from 'graphql'
+import { $$isViewer } from '../symbols.js'
+import { NodeType } from './types'
 import createQueryFields from './query/createQueryFields.js'
 
 const createViewerField = memoize(schema => ({
   type: new GraphQLNonNull(new GraphQLObjectType({
     name: 'Viewer',
-    description: '',
-    fields: createQueryFields(schema),
+    description: 'The viewer type, provides a “view” into your data. To be used with Relay.',
+    interfaces: [NodeType],
+    isTypeOf: value => value[$$isViewer],
+    fields: {
+      ...createQueryFields(schema),
+      id: {
+        type: GraphQLID,
+        description:
+          'An identifier for the viewer node. Just the plain string “viewer.” ' +
+          'Can be used to refetch the viewer object in the `node` field. This ' +
+          'is required for Relay.',
+
+        resolve: () => 'viewer',
+      },
+    },
   })),
 
   description:
-    'A single entry query for advanced data clients like Relay. Nothing ' +
+    'A single entry query for the advanced data client Relay. Nothing ' +
     'special at all, if you don’t know what this field is for, you probably ' +
     'don’t need it.',
 
-  resolve: source => source || {},
+  resolve: () => ({ [$$isViewer]: true }),
 }))
 
 export default createViewerField
--- a/src/graphql/mutation/createDeleteMutationField.js
+++ b/src/graphql/mutation/createDeleteMutationField.js
@@ -3,7 +3,9 @@ import { $$rowTable } from '../../symbols.js'
 import SQLBuilder from '../../SQLBuilder.js'
 import getType from '../getType.js'
 import createTableType from '../createTableType.js'
-import { inputClientMutationId, payloadClientMutationId } from './clientMutationId.js'
+import getPayloadInterface from './getPayloadInterface.js'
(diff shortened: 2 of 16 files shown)


msimulcik:
Great work, it's almost perfect! However, Viewer has to implement Node interface to be used with relay mutations. The open question here is what should be the id field of viewer node. I was thinking about id of current postgresql role.

Another thing, that is much easier to add, is missing edge field in mutation payload. The edge type definition is already there, it just needs to be attached to mutation payload.

For example, currently InsertTodoPayload contains:
```
todo: Todo
clientMutationId: String
viewer: Viewer!
```
The desired structure is:
```
todo: Todo
todoEdge: TodoEdge
clientMutationId: String
viewer: Viewer!
```

calebmer:
So Relay wants the edge with the cursor?

calebmer:
Also, I really can't believe Relay wants `Viewer` to be a `Node`, if so this might be the wrong implementation. Do you have a source for that?

msimulcik:
Answers for both questions can be found in [Relay docs for RANGE_ADD mutation](https://facebook.github.io/relay/docs/guides-mutations.html#range-add)

>So Relay wants the edge with the cursor?

This is required for `edgeName` field in mutation configuration (see linked docs).

>Also, I really can't believe Relay wants Viewer to be a Node, if so this might be the wrong implementation. Do you have a source for that?

Relay documentation doesn't directly mention Viewer field much, so this requirement is rather implicit. Take a look on the `ParentID` field in docs. Description says: "The DataID of the parent **node** that contains the connection". In our case the parent of all connections is Viewer so it has to be node and it has to have ID.

[Reindex docs on mutations](https://www.reindex.io/docs/graphql-api/mutations/) also contain good examples and explanations.

burkhardr:
@calebmer Yes, it does need to be a node, and it does need to implement the `NodeInterface`.

Re: the discussion around the viewer ID. In all my previous apps and others I've seen, this was always a singleton, so the ID doesn't really matter. The viewer node is simply a static entry point. I've implemented this as `id: viewer`, facebook use `id: me` (afaik).

calebmer:
Ok, so I'd also need the `node` field to return viewers?

burkhardr:
That's correct. The viewer field type itself is really simple, but it does require a bit of wiring up, unfortunately.

calebmer:
After some work with Relay I better understand the need for this. With that knowledge, do you think `viewer` is the right name? What do you think about making it called `root` or `query`?

Then the name `viewer` can then be reserved for a procedure returning a `User` type or something.

ferdinandsalis:
@calebmer see this issue https://github.com/facebook/relay/issues/820. However `viewer` is used in all public examples and so do I.

burkhardr:
I agree that `viewer` is quite frontend-specific (e.g. relay), but `query` would be confusing (`query { query { ... } }`). Also, it kinda of is a "view" of your data structure *

\* lame excuse ;)

calebmer:
Ok, viewer is general enough. We could recommend `me` or `game` etc. for something more application specific. Also people instantly understand what `viewer` is.

I'll implement `Node` for viewer then merge this in so people can start using it. I'll implement @msimulcik's advised mutation fields in another PR.

calebmer:
Ok, implemented node for viewer field. If you guys think this is ready to merge in it’s most basic form let me know!

Look at [this](https://github.com/calebmer/postgraphql/blob/feat/relay-viewer-field/tests/integration/fixtures/viewer-query.graphql) and [this](https://github.com/calebmer/postgraphql/blob/feat/relay-viewer-field/tests/integration/fixtures/viewer-query.json) integration test files to see the viewer field used.

ferdinandsalis:
If I am not mistaken the viewer type is missing the `id` which is required to configure mutations (e.g. `RANGE_ADD` see https://facebook.github.io/relay/docs/guides-mutations.html#range-add. The commonly used in examples is just a plain string of `me`.

msimulcik:
Looks great! I'm gonna try it with Relay today and let you know.

calebmer:
Released in [`1.5.0`](https://github.com/calebmer/postgraphql/releases/tag/v1.5.0) 🙌

rojobuffalo:
> @calebmer 
> After some work with Relay I better understand the need for this. With that knowledge, do you think viewer is the right name? What do you think about making it called root or query?
>
> Then the name viewer can then be reserved for a procedure returning a User type or something.

It would be really useful if `viewer` resolved to a User type for the user who is logged in. 

What if this was moved to root and then viewer resolved a user node?
#145

calebmer:
@rojobuffalo:

In #145 I did end up renaming `viewer` to `query` as I think many users would like to define their own custom `viewer` field in the form of a procedure :blush:

the functionality is exactly the same and you can always alias `query` to `viewer` if you prefer that :+1:

valorize:
Aliasing is done by `query { viewer: query { ... } }` right?

calebmer:
@valorize correct :+1:

Out of curiosity why are you interested in aliasing the field to `viewer`?

valorize:
@calebmer Just saw that you mentioned it's possible. So I wanted to be sure how to do it. I guess it's best to use as less aliases as possible.

Issue #51, opened 2016-05-18
Disable auto updatable views


pvdvreede:
Hey there,

I am playing with postgraphql with one of our current schemas and noticed that if i create a view to a table in another schema it also generates the mutations for that view.

I would prefer it not to show those mutations, and have tried creating rules set to `DO NOTHING` for the view, but the graphql introspection is still showing the mutations. 

Is there a way to tell postgraphql to not generate mutations for particular views?

pvdvreede:
It also looks like the associations between two tables does not come through in the views. Is there a way to get postgraphql to see an association between two views?

I would like to be able to manually tweak how the tables are exposed through to graphql so that the underlying table structure is abstracted away. This way I can show domain objects rather than normalised db table objects.

calebmer:
As I mentioned in #53, I feel like the best solution is to detect granted access levels of the role used at schema creation time. If that role doesn’t have access to update an auto updatable view, we can hide that from our mutations list.

calebmer:
Released in `1.7.0` 🎉

PR #53, opened 2016-05-19
feat(postgres): add custom types to table filter.

pvdvreede:
This request expands the table definition to include custom types. I would like to use custom types in functions to represent data I want to return but that I do not want to persistently store, so it would be good for graphql to represent these types in its type system.

Currently if I add a custom type as a function return, graphql represents it as a String.

Not sure if you are happy to have as part of "tables" as I have done here, but let me know if this is alright or if you want to have a discreet type for it and have it inherit from the `Type` class as per enums, I was just following the KISS principle 😄 .

### DIFF
--- a/src/graphql/mutation/createMutationType.js
+++ b/src/graphql/mutation/createMutationType.js
@@ -22,6 +22,7 @@ const createMutationType = schema =>
       ...(
         schema
         .getTables()
+        .filter(table => table.kind === 'r')
         .map(table => createMutationFields(table))
         .reduce(ary(assign, 2), {})
       ),
--- a/src/graphql/mutation/createMutationType.js
+++ b/src/graphql/mutation/createMutationType.js
@@ -15,7 +15,6 @@ const createMutationType = schema =>
         schema
         .getProcedures()
         .filter(({ isMutation }) => isMutation)
-        .filter(procedure => !procedure.hasTableArg())
         .map(procedure => [procedure.getFieldName(), createProcedureMutationField(procedure)])
       ),
       // Add standard fields for tables.
--- a/src/postgres/catalog.js
+++ b/src/postgres/catalog.js
@@ -120,10 +120,11 @@ export class Schema {
  * @member {ForeignKey[]} reverseForeignKeys
  */
 export class Table {
-  constructor ({ schema, name, description }) {
+  constructor ({ schema, name, description, kind }) {
     this.schema = schema
     this.name = name
     this.description = description
+    this.kind = kind
   }
 
   getColumns = once(() => {
--- a/src/postgres/getCatalog.js
+++ b/src/postgres/getCatalog.js
@@ -72,14 +72,15 @@ const addTables = (client, catalog) =>
     select
       n.nspname as "schemaName",
       c.relname as "name",
-      d.description as "description"
+      d.description as "description",
+      c.relkind as "kind"
     from
       pg_catalog.pg_class as c
       left join pg_catalog.pg_namespace as n on n.oid = c.relnamespace
       left join pg_catalog.pg_description as d on d.objoid = c.oid and d.objsubid = 0
     where
       n.nspname not in ('pg_catalog', 'information_schema') and
-      c.relkind in ('r', 'v', 'm', 'f');
+      c.relkind in ('r', 'v', 'm', 'f', 'c');
   `)
   .then(({ rows }) => rows)
   .map(row => new Table({
@@ -111,7 +112,7 @@ const addColumns = (client, catalog) =>
         cp.conkey::int[] @> array[a.attnum::int]
     where
       n.nspname not in ('pg_catalog', 'information_schema') and
-      c.relkind in ('r', 'v', 'm', 'f') and
+      c.relkind in ('r', 'v', 'm', 'f', 'c') and
       a.attnum > 0 and
       not a.attisdropped
     order by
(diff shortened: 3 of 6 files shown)


pvdvreede:
Just realised that adding it to tables will probably mean that mutations will be created. I will have a look and see what needs to be tweaked.

pvdvreede:
I have now filtered the table generated mutations to only occur with ordinary tables (`r`). Obviously auto updatable views could be included but I am not sure how to introspect whether a view is auto updatable, and as per my issue #51 it would be good to have a way to control whether an auto updatable view appears as a mutation. Please let me know your thoughts/suggestions on this.

Until this is worked out, I have erred on the side of caution to only allow ordinary tables to generate mutations, and assumed that users can just create functions for inserting things if they want more flexibility.

pvdvreede:
Just added another commit to fix what I think is a bug? I noticed that if i set a proc as volatile where the arg is a table type, that it would show up as part of a query rather than a mutation.

I had hoped to see it as a mutation so that I can create functions to create objects with some extra logic or checks but not have to re write a tables columns as separate inputs.

So my commit ignores whether or not an arg is a table type and just relied on the volatility being true.

src/graphql/mutation/createMutationType.js
```diff
@@ -15,13 +15,13 @@ const createMutationType = schema =>
         schema
         .getProcedures()
         .filter(({ isMutation }) => isMutation)
-        .filter(procedure => !procedure.hasTableArg())
```
calebmer: The reason I filtered out procedures with table args is because I wasn’t sure how input for such procedures would work, could you maybe add a test case?

src/graphql/mutation/createMutationType.js
```diff
@@ -15,13 +15,13 @@ const createMutationType = schema =>
...
         .map(procedure => [procedure.getFieldName(), createProcedureMutationField(procedure)])
       ),
       // Add standard fields for tables.
       ...(
         schema
         .getTables()
+        .filter(table => table.kind === 'r')
```
calebmer: This would be a breaking change, so I’d rather not have it in the PR. Although I get your point about not allowing views that aren’t updatable to have mutations. I think the best solution is to use access levels to determine what relations (tables, views, etc) the user has been granted rights to mutate.

src/postgres/catalog.js
```diff
@@ -120,10 +120,11 @@ export class Schema {
  * @member {ForeignKey[]} reverseForeignKeys
  */
 export class Table {
-  constructor ({ schema, name, description }) {
+  constructor ({ schema, name, description, kind }) {
     this.schema = schema
     this.name = name
     this.description = description
+    this.kind = kind
```
calebmer: Could you create some constants for `kind`? So for example:

```js
export const KIND_ORDINARY_TABLE = 'r'
export const KIND_VIEW = 'v'
…
```

This could help contributors in the future. If you could also add `kind` to the JSDoc above that would be great 😊👍

src/graphql/mutation/createMutationType.js
calebmer: The best solution for this might be to allow custom table types and hide the rest.

calebmer:
Thanks so much for the contributions 😊 🎉

When I thought about implementing this, I didn’t even consider it being part of the tables class. I like it, it makes a lot of sense. The only thing I might warn about is make sure that query/mutation fields aren’t created for the custom types. Extending the table type does add complexity, but I think it’s manageable especially if we add a note in the table JSDoc comment so that contributors know the table class may be a custom type.

If you could add a couple integration tests (see [the kitchen sink schema](https://github.com/calebmer/postgraphql/blob/master/examples/kitchen-sink/schema.sql) and [these](https://github.com/calebmer/postgraphql/tree/master/tests/integration/fixtures) files) that would be awesome 👍. Also an integration test for the volatile procs with a table type arg would be nice. I’m not sure what the input for that would be, I initially hid those procedures so I wouldn’t have to think about that 😊

Excited to see this get merged, it’s definitely been a feature on my personal roadmap. Once we get this merged I’ll add you as a contributor as this project is being run [open open source](http://openopensource.org/).

calebmer:
Sorry about the lint errors. You don’t have to worry about those, I’ll clean them up.

pvdvreede:
Great, thanks for the feedback! I will work on these changes over the next week or so.

I agree that it is better to work out what mutations should be displayed based on whether those views/tables are writable. Do you know the best way to surface that information on the table objects? 

I was pondering this and am not sure what the best way to go about it is. Do I need to consider the authentication side of things and have to recreate the schema for every request as each request might be run in the context of a different postgres role that will have different permissions? It would be optimum to only display a schema that the user has access to read/write to rather than see a table in the schema but get a permission denied error when they try and write to it. However, this might have performance issues to recreate the schema every request and perhaps is not a good starting point given the extra complexity.

So are you happy to have the permissions be determined by the postgraphql user on boot up even if they may differ from a role that executes a mutation?

pvdvreede:
Just saw the integration tests which I completely missed! Nice! 👍 

I will certainly add in some integration tests this round.

calebmer:
If you look at [`pg_class`](http://www.postgresql.org/docs/current/static/catalog-pg-class.html) it has a `relacl` column. I'm not sure the specifics but that should be a good start.

I always imagined the schema would only ever be created on start, throwing errors after that for permission errors. How PostgreSQL roles work, the initial authenticated user should have access to everything child roles do.

I also believe GraphQL was designed with a single schema approach for all users in mind.

pvdvreede:
Sorry I havent had a chance to look at this yet, Im hoping to get some time today and tomorrow to get going.

calebmer:
It's all good, I'm pretty excited to land this PR 😊

pvdvreede:
I have gone about this different way given your feedback so have started a new PR at #73. So I will close this one, and if you can give me feedback on the new one that would be great.

PR #57, opened 2016-05-29
fix(graphql): explicitly check input values for undefined

ferdinandsalis:
Otherwise boolean and null values get coerced which will render
undesired result.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "postgraphql",
-  "version": "1.3.0",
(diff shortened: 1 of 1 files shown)


calebmer:
Good catch, thanks so much for the contribution!

ferdinandsalis:
@calebmer anything else you want me to do here?

ferdinandsalis:
Ups that was unintended …

PR #58, opened 2016-05-30
fix(procedure): convert js arrays to postgres array type

ferdinandsalis:
Just my first humble try on solving this. Currently only handles one-dimensional arrays. Maybe to naive, etc. Feedback very welcome.

### DIFF
--- a/src/graphql/resolveProcedure.js
+++ b/src/graphql/resolveProcedure.js
@@ -1,6 +1,11 @@
-import { camelCase } from 'lodash'
+import { camelCase, isArray, mapValues } from 'lodash'
 import { $$rowTable } from '../symbols.js'
 import createProcedureCall from './createProcedureCall.js'
+
+function asPgValue (value) {
+  const newValue = isArray(value) ? `{${value.map(i => JSON.stringify(i)).join(',')}}` : value
+  return newValue
+}
 
 const resolveProcedure = (procedure, getProcedureArgs) => {
   // If this type is a table type, this variable will be a reference to that table.
@@ -34,12 +39,21 @@ const resolveProcedure = (procedure, getProcedureArgs) => {
   return async (source, args, { client }) => {
     const procedureArgs = getProcedureArgs(source, args)
 
-    // Actuall run the procedure using our arguments.
-    const result = await client.queryAsync({
-      name: query.name,
-      text: query.text,
-      values: argEntries.map(([name]) => procedureArgs[camelCase(name)]),
+    const values = argEntries.map(([name, type]) => {
+      const obj = procedureArgs[camelCase(name)]
+
+      // See https://github.com/calebmer/postgraphql/pull/58
+      if (type.isTableType) {
+        mapValues(obj, (val, key) => {
+          obj[key] = asPgValue(val)
+        })
+      }
+
+      return obj
     })
+
+    // Actuall run the procedure using our arguments.
+    const result = await client.queryAsync({ ...query, values })
 
     // If the procedure returns a set, return all of the rows.
     if (procedure.returnsSet)
--- a/examples/kitchen-sink/schema.sql
+++ b/examples/kitchen-sink/schema.sql
@@ -16,6 +16,13 @@ create table relation (
   primary key (a_thing_id, b_thing_id)
 );
 
+create table another_thing (
+  id               serial not null primary key,
+  note             text not null,
+  published        boolean not null,
+  tags             text[] not null
+);
+
 create function add(a int, b int) returns int as $$
   select a + b
 $$ language sql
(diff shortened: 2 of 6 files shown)


calebmer:
Thanks so much for your contribution! 🎉

Although I feel like this is a small band aid on a more systemic problem. Might you try running the objects through `JSON.stringify`? Also a test case would be nice. Look at the kitchen sink example and its corresponding integration tests.

ferdinandsalis:
Could you elaborate more on the systemic problem …

Will look at the kitchen sink example for tests. Also not sure what you mean by running the objects through `JSON.stringify` — the `values` object?

ferdinandsalis:
I have added test (still local). Of course all the procedure tests fail with this pull request. I have underestimated the complexity … ☺️  I assume this is because procedures can have tables as arguments or individual fields.

ferdinandsalis:
Ok, second try. Excuse the naming and such. Happy to restructure to your wishes.
Will fix the linting errors.

ferdinandsalis:
rebased to include the newest change on master …

calebmer:
What I mean by “systemic problem” is the very way we support table arguments in the first place. So let’s take a computed column for example:

```sql
create function full_name(person person) returns text as $$
  select person.given_name || ' ' || person.family_name
$$ language sql;
```

Somehow we have to call that procedure with an argument of type `person`. This is difficult because we don’t want to re-query the person object from the database (we just did after all), so we send the JSON object we just selected instead. Basically our two options are:

```sql
select full_name((select * from person where id = 1));
```

or:

```sql
select full_name('{"givenName":"Test","familyName":"User"}'::person)
```

We have chosen the second approach which can lead to edge-casey errors like this one. The optimal solution would be to look at the GraphQL query so we can “predict” if we will need a computed column like so:

```sql
select *, full_name(person) from person where id = 1
```

But that requires more effort and may be out of scope for now.

* * *

I’ll comment on what I mean about using `JSON.stringify` on the code 👍

src/graphql/resolveProcedure.js
```diff
@@ -34,13 +39,26 @@ const resolveProcedure = (procedure, getProcedureArgs) => {
...
-    const result = await client.queryAsync({
-      name: query.name,
-      text: query.text,
-      values: argEntries.map(([name]) => procedureArgs[camelCase(name)]),
+    const values = argEntries.map(([name, type]) => {
+      const obj = procedureArgs[camelCase(name)]
+      let value
+
+      if (type.isTableType) {
+        value = reduce(obj, (acc, val, key) => {
```
calebmer: Try instead of using `reduce` and converting keys to an array here, try just using `JSON.stringify`. I’m not sure if it will work, just curious 😊

Another approach may be to convert the object into a PostgreSQL tuple directly. So for example a table that looks like:

```sql
create table test (a int, b int);
```

And an object that looks like:

```json
{
  "a": 1,
  "b": 2
}
```

Would become a tuple that looks like:

```sql
(1,2)
```

The main difference being we remove the JSON middle man and convert directly to a PostgreSQL type. This might be more future proof?

This could also be a bug in [`pg-types`](https://www.npmjs.com/package/pg-types). Thus why I’m curious to see if `JSON.stringify` works.

src/graphql/resolveProcedure.js
```diff
@@ -34,13 +39,26 @@ const resolveProcedure = (procedure, getProcedureArgs) => {
...
+          acc[key] = convertToPostgresArrayIfArray(val)
+          return acc
+        }, {})
+      }
+      else {
+        value = convertToPostgresArrayIfArray(obj)
```
calebmer: Do things break if we remove this? If so is there a test case?

src/graphql/resolveProcedure.js
ferdinandsalis: This was actually one of the first things I tried. However this does not work and returns the same error. More information here — https://www.postgresql.org/docs/9.5/static/functions-json.html.

> Note: In json_populate_record, json_populate_recordset, json_to_record and json_to_recordset, type coercion from the JSON is "best effort" and may not result in desired values for some types.

It does not seem to be a bug with `pg-types` though. After investigating this I looks like it only convert pg types to javascript types but maybe I am wrong.

How should we go about this now? If we go the route of converting it to postgres tupe directly how would you approach this?

calebmer: It should be fine for now then. Just some small changes, instead of `reduce` could you use `mapValues`? Also could you rename `convertToPostgres`… to `asPgValue`. It's shorter and follows our `pg` convention. Also it can be extended to support more cases in the future 😊

ferdinandsalis: Will do. Thanks for the feedback.

src/graphql/resolveProcedure.js
ferdinandsalis: No it doesnt. Honestly I expected it to break. I am curious as why it doesn’t. For example shouldn’t it break when values has the following form `['c', true, ["a", "b", "c"]]`.

calebmer:
Looks good to me. What do you think?

ferdinandsalis:
Yes. I think we can merge it.

calebmer:
Merged, as I want to run this project as open-open source do you want to become a collaborator?

ferdinandsalis:
@calebmer of course. Thats a first for me. Thanks.

calebmer:
Done :+1:

I’m going to wait to release this with #64.

calebmer:
Released in `1.5.1`.

PR #60, opened 2016-05-31
fix(graphql): explicitly check input values for undefined

ferdinandsalis:
Otherwise boolean and null values get coerced which will render
undesired result.


ferdinandsalis:
Sorry again. Had the fix on master and wanted to change that. 
https://github.com/calebmer/postgraphql/pull/57

calebmer:
Looks good to me 👍

PR #61, opened 2016-05-31
style(graphql): fix linting errors

ferdinandsalis:
Disregard if you want this solved a different way or just keep as is. ☺️ 

🖖


calebmer:
Oh man, much thanks. These errors have been around a while 😊

PR #65, opened 2016-06-06
fix(resolve): check for empty string in query values

ferdinandsalis:
Foreign keys that are of type `int` are converted to strings [(diff)](https://github.com/calebmer/postgraphql/compare/master...ferdinandsalis:fix-resolve-single?expand=1#diff-0f601b95d640a7e7b90036b6dfb374fbL51). I know this is intended since `any`’s righthand side expects an array expression (if my research was correct). However when a foreign key is null it is interpreted as a string and result in the following error: `error: invalid input syntax for integer`.

I have yet to add a test. Though I am not sure where ☺️.

### DIFF
--- a/examples/kitchen-sink/schema.sql
+++ b/examples/kitchen-sink/schema.sql
@@ -20,7 +20,8 @@ create table another_thing (
   id               serial not null primary key,
   note             text not null,
   published        boolean not null,
-  tags             text[] not null
+  tags             text[] not null,
+  thing_id         int references thing(id) on delete cascade
 );
 
 create function add(a int, b int) returns int as $$
@@ -123,11 +124,6 @@ $$ language plpgsql
 strict
 set search_path from current;
 
-insert into another_thing (note, published, tags) values
-  ('hello', true, '{"a", "b"}'),
-  ('world', true, '{"c", "d"}'),
-  ('foo', false, '{"a"}');
-
 insert into thing (note) values
   ('hello'),
   ('world'),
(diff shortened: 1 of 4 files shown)


ferdinandsalis:
@calebmer anything I missed here or misunderstood?

calebmer:
A test case so a regression isn't made in the future 😊

ferdinandsalis:
Will do, but see
> I have yet to add a test. Though I am not sure where ☺️.

calebmer:
Gotcha, missed that 😊

I'm not necessarily sure what the issue was (thus why a test would help). Is adding a table or column to `kitchen-sink` not enough?

ferdinandsalis:
Will look into it. Thanks.

ferdinandsalis:
I have added a test. Let me know if you want me to change something, especially the additions in the kitchen sink ☺️ Wasn’t sure where and how to add it best. It looks like this will get a bit unwieldy if more tests need to be added. What do you think?

calebmer:
It might be easier to add a column like this:

```sql
create table another_thing (
  id               serial not null primary key,
  note             text not null,
  published        boolean not null,
  tags             text[] not null,
  thing_id          int references thing(id) on delete cascade
);
```

And then nullify `thing_id` in `another_thing` in one of the rows.

ferdinandsalis:
@calebmer thanks. done.

calebmer:
Cool, just rebase and I'll merge! 🎉

ferdinandsalis:
@calebmer done 😄

PR #68, opened 2016-06-13
feat(graphql): add relay mutation types support

ferdinandsalis:
Feedback very much welcome.

### DIFF
--- a/src/graphql/mutation/createDeleteMutationField.js
+++ b/src/graphql/mutation/createDeleteMutationField.js
@@ -76,7 +76,7 @@ const createPayloadType = table =>
 // Resolves the id from the primary keys of the deleted resource
 const resolveDeletedFieldId = table => ({ output }) => {
   if (!output)
-    return null 
+    return null
 
   const primaryKeys = table.getPrimaryKeys()
   const deletedIds = primaryKeys.map(pkey => output[pkey.name])
--- a/src/graphql/mutation/createDeleteMutationField.js
+++ b/src/graphql/mutation/createDeleteMutationField.js
@@ -1,6 +1,7 @@
 import {
   GraphQLNonNull,
   GraphQLObjectType,
+  GraphQLID,
   GraphQLInputObjectType,
 } from 'graphql'
 
@@ -8,6 +9,7 @@ import { fromPairs } from 'lodash'
 import { $$rowTable } from '../../symbols.js'
 import SQLBuilder from '../../SQLBuilder.js'
 import getType from '../getType.js'
+import { toID } from '../types.js'
 import createTableType from '../createTableType.js'
 import getPayloadInterface from './getPayloadInterface.js'
 import getPayloadFields from './getPayloadFields.js'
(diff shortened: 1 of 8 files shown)


src/graphql/mutation/createDeleteMutationField.js
```diff
@@ -57,15 +58,25 @@ const createPayloadType = table =>
     description: `Contains the ${table.getMarkdownTypeName()} node deleted by the mutation.`,
     interfaces: [getPayloadInterface(table.schema)],
     fields: {
-      [table.getFieldName()]: {
```
calebmer: I really don’t want to introduce breaking changes wherever possible, could we leave this field but also have the `deleted*id` field? This field could also be helpful for non-Relay users.

src/graphql/mutation/createInsertMutationField.js
```diff
@@ -60,12 +63,45 @@ const createPayloadType = table =>
...
+
+  // if the edge field was handed an orderBy argument we use
+  // this as the cursor
+  if (orderBy) {
+    cursor = output[orderBy]
+  }
+  // if orderBy is not present use the primary keys of the table
+  else {
+    const primaryKeys = table.getPrimaryKeys()
+    cursor = primaryKeys.map(column => output[column.name])
```
calebmer: Instead of doing this, could we use a default value for the argument same as what we do [here](https://github.com/calebmer/postgraphql/blob/master/src/graphql/createConnectionArgs.js#L16)? This means calling `fooEdge` without arguments would result in the exact same cursor you would get from the connection without an `orderBy` argument.

tests/integration/fixtures/mutation-delete.json
```diff
@@ -2,45 +2,31 @@
   "data": {
     "a": {
       "clientMutationId": null,
-      "thing": {
-        "rowId": 1,
```
calebmer: Remember, if you add this field back also add these tests back. I think adding this field back actually would be super helpful because it’s the only way to see now deleted values like `note`.

tests/integration/fixtures/mutation-insert.graphql
```diff
@@ -6,27 +6,33 @@ mutation Insert {
 
 fragment payloadThing on InsertThingPayload {
   clientMutationId
-  thing {
```
calebmer: Keep this test too in addition to the new one.

tests/integration/fixtures/mutation-insert.graphql
```diff
@@ -6,27 +6,33 @@ mutation Insert {
...
+      note
+    }
   }
 }
 
 fragment payloadRelation on InsertRelationPayload {
   clientMutationId
-  relation {
-    aThingId
-    bThingId
```
calebmer: Ditto, generally removing tests is a bad sign because it means we might break things for some users.

tests/integration/fixtures/mutation-insert.graphql
```diff
@@ -6,27 +6,33 @@ mutation Insert {
 
 fragment payloadThing on InsertThingPayload {
   clientMutationId
-  thing {
-    id
-    rowId
-    note
+  thingEdge(orderBy: ROW_ID) {
```
calebmer: Could we also test this without arguments? On the same payload?

ferdinandsalis:
@calebmer thank you for the extensive feedback 🙇🏽.

ferdinandsalis:
@calebmer I have addressed all of your issues to the best of my knowledge. ☺️

calebmer:
Looks good to me, update and I’ll merge 👍

calebmer:
Released in `1.6.0` 🎉

PR #71, opened 2016-06-14
fix(server): throw error for bad auth header

### LINKED ISSUE #62: How to force valid JWT tokens
I love this project, it is so concise and to the point, especially the scope of what it is trying to achieve!

So I was able to generate a token and have posgraphql evaluate the claim:

```
curl -XPOST -H "Authorization:Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhdWQiOiJwb3N0Z3JhcGhxbCIsInJvbGUiOiJncmFwaHFsIiwidXNlcl9pZCI6Mn0.oQVZfTK-LdC1btVvR8OrzRAhSPEKsFJHeqxhF9xk8rc" -H 'Content-Type:application/graphql' -d '{ searchPosts(search:"s",first:2){nodes{rowId}}}' 127.0.0.1:3000
```

So I was able to get it to return valid date for a valid token containing a valid role.

However, I also able to get the same data with an empty token:

```
curl -XPOST -H "Authorization:Bearer " -H 'Content-Type:application/graphql' -d '{ searchPosts(search:"s",first:2){nodes{rowId}}}' 127.0.0.1:3000
```

Is there a way to mandate a non-empty token if the secret flag has been set (other than say enforcing this with an upstream nginx proxy)?

Any other non-empty token string appears to evaluate correctly (it would be nice to get a 403 instead of a 500, but that's a separate issue).

calebmer:
Closes #62

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "express": "^4.13.4",
     "express-graphql": "^0.5.0",
     "graphql": "^0.5.0",
+    "http-errors": "^1.5.0",
     "jsonwebtoken": "^6.2.0",
     "lodash": "^4.8.2",
     "morgan": "^1.7.0",
--- a/src/createServer.js
+++ b/src/createServer.js
@@ -2,6 +2,7 @@ import path from 'path'
 import { forEach } from 'lodash'
 import Express from 'express'
 import onFinished from 'on-finished'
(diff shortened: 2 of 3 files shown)


ferdinandsalis:
Looks good to me ☺️

PR #70, opened 2016-06-14
feat(graphql): add fields based on unique constraints

calebmer:
Closes #67

Let the reviews, begin!

PR #73, opened 2016-06-15
feat(postgres): only show mutations for types that are updatable.

pvdvreede:
This is a rework of #53.

I will filter mutations to only show postgres objects that are updatable. This will hide views that are not auto updatable from the mutations graphql schema. 

It will also hide custom types, so the second part of this pr will add in custom types so that they can be used as input and output values of queries and mutations.

### DIFF
--- a/examples/kitchen-sink/schema.sql
+++ b/examples/kitchen-sink/schema.sql
@@ -24,6 +24,12 @@ create table another_thing (
   thing_id         int references thing(id) on delete cascade
 );
 
+create view non_mutation_view as
+  select 1;
+
+create view mutation_view_for_thing as
+  select * from thing;
+
 create function add(a int, b int) returns int as $$
   select a + b
 $$ language sql
--- a/src/postgres/getCatalog.js
+++ b/src/postgres/getCatalog.js
@@ -71,11 +71,33 @@ const addSchemas = (client, catalog) =>
   .each(schema => catalog.addSchema(schema))
 
 const addTables = (client, catalog) =>
+  /* eslint-disable */
+  // Determine any views or other objects that we cannot mutate with
+  // pg_relation_is_updatable, more info:
+  // https://www.postgresql.org/message-id/CAEZATCV2_qN9P3zbvADwME_TkYf2gR_X2cLQR4R+pqkwxGxqJg@mail.gmail.com
+  // https://github.com/postgres/postgres/blob/2410a2543e77983dab1f63f48b2adcd23dba994e/src/backend/utils/adt/misc.c#L684
+  // https://github.com/postgres/postgres/blob/3aff33aa687e47d52f453892498b30ac98a296af/src/backend/rewrite/rewriteHandler.c#L2351
+  /* eslint-enable */
   client.queryAsync(`
     select
       n.nspname as "schemaName",
       c.relname as "name",
-      d.description as "description"
+      d.description as "description",
+      case when
+           (pg_catalog.pg_relation_is_updatable(c.oid, true)::bit(8) & B'00010000') = B'00010000'
+           then true
+           else false
+      end as "isInsertable",
+      case when
+           (pg_catalog.pg_relation_is_updatable(c.oid, true)::bit(8) & B'00001000') = B'00001000'
+           then true
+           else false
+      end as "isUpdatable",
+      case when
+           (pg_catalog.pg_relation_is_updatable(c.oid, true)::bit(8) & B'00000100') = B'00000100'
+           then true
+           else false
+      end as "isDeletable"
     from
       pg_catalog.pg_class as c
       left join pg_catalog.pg_namespace as n on n.oid = c.relnamespace
--- a/tests/integration/fixtures/non-updatable-views-filtered.graphql
+++ b/tests/integration/fixtures/non-updatable-views-filtered.graphql
@@ -1,0 +1,27 @@
+query {
+  noInsertMutation: __type(name: "InsertNonMutationViewPayload") {
+    name
+  }
+  noUpdateMutation: __type(name: "UpdateNonMutationViewPayload") {
+    name
+  }
+  noDeleteMutation: __type(name: "DeleteNonMutationViewPayload") {
+    name
+  }
+
+  insertMutation: __type(name: "InsertMutationViewForThingPayload") {
+    name
+  }
+  updateMutation: __type(name: "UpdateMutationViewForThingPayload") {
+    name
+  }
+  deleteMutation: __type(name: "DeleteMutationViewForThingPayload") {
+    name
+  }
+
+  nonMutationViewNodes(orderBy: COLUMN) {
+    nodes {
+      column
+    }
+  }
+}
--- a/tests/postgres/getCatalog.test.js
+++ b/tests/postgres/getCatalog.test.js
@@ -55,6 +55,10 @@ create view b.yo as
   from
     a.hello;
 
+create view c.no_update as
+  select
+    (1 + 1) as col1;
+
 comment on view b.yo is 'YOYOYO!!';
 comment on column b.yo.constant is 'This is constantly 2';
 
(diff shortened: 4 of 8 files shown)


pvdvreede:
It took me a while to figure out how to ask postgres which objects are mutable. Bitmasking is not really my strong suit so it would be good to have another pair of eyes look into the `pg_catalog.pg_relation_is_updatable` function and let me know if the query needs to be altered.

pvdvreede:
Just added a second WIP commit that I am playing around with how best to implement composite types for any initial thoughts on whether this is the right path to go down.

ferdinandsalis:
@pvdvreede can you help with understanding this by illustrating a simple use case for custom types. I am ignorant ☺️

src/postgres/getCatalog.js
```diff
@@ -75,7 +75,22 @@ const addTables = (client, catalog) =>
     select
       n.nspname as "schemaName",
       c.relname as "name",
-      d.description as "description"
+      d.description as "description",
+      case when
+           (pg_catalog.pg_relation_is_updatable(c.oid, true)::bit(8) & B'00010000') = B'00010000'
```
calebmer: Oh wow, this looks like a nice function. Could I see its documentation? Also if you could include a link to the documentation in the code that would be amazing :blush:

tests/integration/fixtures/non-updatable-views-filtered.graphql
```diff
@@ -0,0 +1,17 @@
+query {
+  noInsertMutation: __type(name: "InsertNonMutationViewPayload") {
```
calebmer: Clever :blush:

In the future I’m probably going to use [this](http://graphql.org/docs/api-reference-utilities/#printschema) function so that we can better test the contents of our generated GraphQL schema.

tests/postgres/getCatalog.test.js
```diff
@@ -207,6 +211,11 @@ describe('getCatalog', function testGetCatalog () {
     expect(catalog.getColumn('a', 'types', 'domain2').type.baseType.id).toEqual(23)
   })
 
+  it('contains relation updatable information', () => {
+    expect(catalog.getTable('c', 'no_update').isInsertable).toBe(false)
```
calebmer: Could you also test `isUpdatable` and `isDeletable`?

src/postgres/getCatalog.js
calebmer: Another option is we could try to do a mock insert/update/delete and observe the error codes. So something like:

```sql
begin;
insert into non_updatable values (1);
rollback;
```

And if a `55000` is thrown ([error codes](https://www.postgresql.org/docs/9.1/static/errcodes-appendix.html)) we know we can’t insert. This is just an idea, until I see the `pg_relation_is_updatable` docs I can’t be certain which method is better so I defer this decision to you.

src/graphql/getType.js
```diff
@@ -102,6 +103,17 @@ const getType = memoize(type => {
   if (type.isDomain)
     return getType(type.baseType)
 
+  if (type.isComposite) {
+    return new GraphQLInputObjectType({
+      name: upperFirst(camelCase(type.name)),
```
calebmer: Normally, I like to put the name conversions in `catalog.js`. See [here](https://github.com/calebmer/postgraphql/blob/master/src/postgres/catalog.js#L167-L173). This also has the advantage of being able to leverage [these two](https://github.com/calebmer/postgraphql/blob/master/src/postgres/catalog.js#L8-L13) functions which allow us to preserve leading and trailing underscores. So `__foo_bar__` would become `__fooBar__` and not `fooBar`.

calebmer:
Just an FYI, all integration tests are failing. I think the reason for that is your WIP commit which should be ok. Are you fine with taking your WIP commit and putting it in another PR? If you think the permissions bit and custom types bit are linked in an important way we should keep them in one PR. Otherwise it may be easier (and we can get the permissions PR merged faster) with two PRs :blush:

tests/integration/fixtures/non-updatable-views-filtered.graphql
pvdvreede: Yeah that function is perfect for declaratively testing what I am doing, and it will make sure things are not missing as well.

src/postgres/getCatalog.js
pvdvreede: We could do inserts and updates with rollbacks, but we would need to keep in mind any static fields in views, cause you will get an error about that field but the actual view is updatable as long as you dont include that field. So would there be a lot of complexity in adding that functionality?

Another option is to outer join to `information_schema.views` but there is no distinction between update and delete there: https://www.postgresql.org/docs/9.1/static/infoschema-views.html. And you would the logic to check for instead of triggers as well.

the `pg_relation_is_updatable` function was a bit hard to find , there doesnt seem to be any official documentation that is jumping out of Google at me. I found it via this: https://www.postgresql.org/message-id/CAEZATCV2_qN9P3zbvADwME_TkYf2gR_X2cLQR4R+pqkwxGxqJg@mail.gmail.com

Then I looked at the code here: https://github.com/postgres/postgres/blob/2410a2543e77983dab1f63f48b2adcd23dba994e/src/backend/utils/adt/misc.c#L684 and https://github.com/postgres/postgres/blob/3aff33aa687e47d52f453892498b30ac98a296af/src/backend/rewrite/rewriteHandler.c#L2351

It doesnt really say whether it is a private or public api, so I just assume it is public because we can access it? In any event it does seem to be the most comprehensive way to check and takes into account instead of triggers as well as auto updatable requirements. Given this, it would be my preference to keep using that - however if there is a less obscure function you would prefer let me know.

pvdvreede:
I have made some changes as per your comments and taken out the WIP. I dont think it has a dependency on the custom type stuff so I can do that in another PR. See my comments re the postgres function and if you are happy to continue using that, this is all good to merge!

pvdvreede:
I have also rebased it against master v1.6.0, so it should be right to merge.

pvdvreede:
@ferdinandsalis Im sure there are people more skilled as postgres than me that can give a better answer  for when to use composite types 😄. For me personally, I would like to be able to create functions that might take a domain model which is then translated and inserted into (possibly) multiple (ie normalised) tables.

So the table types do not represent my domain models, and the domain models do not get stored in that state and dont need a table. Thus, a composite type can be used as a GraphQL input type or output type, without having any persistence behind it or having postgraphql auto generate queries or mutations.

This way, I can create my normalised data tables in one schema and expose a separate schema as graphql with views and functions that better represent the domain and abstract the underlying storage.

ferdinandsalis:
@pvdvreede thanks for the answer 🙇🏽. Sounds interesting.

src/postgres/getCatalog.js
calebmer: Yep, this seems to be the most comprehensive. Good job 😊

Would you mind throwing those links in a comment in the code?

calebmer:
@pvdvreede, @ferdinandsalis correct me if I’m wrong, but the advantage I see with composite types is that they can be typed objects in the GraphQL schema. So for example, a type of:

```sql
create type hello as (a int, b int);
```

Could be the following in the GraphQL schema:

```graphql
type Hello {
  a: Integer
  b: Integer
}
```

That’s the possibility that really excites me 😊

examples/kitchen-sink/schema.sql
```diff
@@ -24,6 +24,9 @@ create table another_thing (
   thing_id         int references thing(id) on delete cascade
 );
 
+create view non_mutation_view as
+  select 1;
```
calebmer: I really want to merge this, but first could you please add a test for auto updatable views that can be mutated? That we can be sure this code never breaks in the future, or with different PostgreSQL versions :blush:

tests/postgres/getCatalog.test.js
```diff
@@ -55,6 +55,10 @@ create view b.yo as
   from
     a.hello;
 
+create view c.no_update as
+  select
```
calebmer: A test for auto-updatable views here too would be nice :blush:

src/postgres/getCatalog.js
calebmer: Nvm, looks like you did put those links in the code :+1:

pvdvreede:
I added some more to the tests, hopefully this is what you meant. Let me know if there are any more tweaks needed.

calebmer:
Looks good 👍 

I’ll try to release this with #70 this weekend.

calebmer:
Just as a note, I’m very hesitant to publish breaking changes, but I don’t consider this a breaking change. Mainly because users shouldn’t be inserting into non-updatable views anyway 😊

calebmer:
Released in `1.7.0` 🎉

PR #81, opened 2016-07-09
Remove emoji which causes "invalid byte sequence for encoding"

tim-field:
Took me a while to track this down when running this schema.

image


calebmer:
That’s too funny ;)

Issue #78, opened 2016-07-06
Some example curl commands would be helpful


dbkaplun:
It would be nice if the README or [forum example](https://github.com/calebmer/postgraphql/tree/master/examples/forum) had some sample curl/wget commands to help one get started. As someone who is just experimenting with postgraphql and GraphQL in general, I would really appreciate it!

For instance, I am querying:
```graphql
{post {id}}
```

The error message `Field "post" argument "id" of type "ID!" is required but not provided.` is not so helpful!

ferdinandsalis:
Instead of using wget or curl I suggest using graphiql, its the graphical interface for your graphql endpoint. If you start up PostGraphQL with the `--development` flag you can access it at the entry point of your grahlql server.

The error message is actually very descriptive. It just says that its missing the id argument. So to make your query work you need to add the id attribute to the query, which could look like this:
```graphql
{
  post(id: "YXJ0aXN0OjE=") { 
    id 
  }
}
```
This would get the post's id field belonging to the post. This is using the opaque id generated by PostGraphQL. If you want to use the id's of your database you will need to do the following:
```graphql
{
  postByRowId(rowId: 1) {
    id
  }
}
```
I hope this helped.

dbkaplun:
Yes, I am familiar with the GraphiQL interface. I meant to add some curl commands specifically, because they are self-contained and serve as kind of canonical representations of HTTP requests. Any graphql query would do as the forum example has no sample queries. :/

Here is the solution I found to my problem:
```graphql
{
  postNodes {
    nodes {
    	id
    }
  }
}
```

ferdinandsalis:
:) Glad you solved the problem.

dbkaplun:
So are you planning on adding some sample queries to the forum example? Would be super helpful

ferdinandsalis:
Adding sample queries to the forum example and explaining them would definitely help people new to GraphQL so I guess it would make sense. What do you think @calebmer?

calebmer:
Yeah, adding sample queries is definetly something we should do.

dbkaplun:
Made a PR with some sample queries: #85

ferdinandsalis:
@dbkaplun thank you for the contribution!

PR #85, opened 2016-07-15
doc(example): add sample queries to forum example README

dbkaplun:
See also #78


calebmer:
Thanks so much 👍  🎉

calebmer:
I consider documentation very important to this project, and we could always use more.

PR #86, opened 2016-07-15
Add CORS headers

calebmer:
Closes https://github.com/calebmer/postgraphql/issues/84

Credit to @jmcelroy5 for the starting point.

Some recommended reading if you’re interested in CORS:

- http://www.html5rocks.com/en/tutorials/cors/
- http://enable-cors.org/

### DIFF
--- a/src/createServer.js
+++ b/src/createServer.js
@@ -31,8 +31,40 @@ const createServer = ({
 }) => {
   const server = new Express()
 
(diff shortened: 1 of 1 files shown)


estaub:
For anyone else who struggles to understand... the CORS headers are now disabled by default, and enabled on the command line by `--cors`.

PR #95, opened 2016-08-11
fix(graphql): non-unique cursors

calebmer:
Fixes non-unique cursors by adding values for the primary key to the cursor and using that to select.

Closes https://github.com/calebmer/postgraphql/issues/93

### DIFF
--- a/src/graphql/resolveConnection.js
+++ b/src/graphql/resolveConnection.js
@@ -8,7 +8,7 @@ const resolveConnection = (
   getFromClause = constant(table.getIdentifier()),
 ) => {
   const columns = table.getColumns()
-  const primaryKey = table.getPrimaryKeys()
+  const fullPrimaryKey = table.getPrimaryKeys()
 
   return (source, args, { client }) => {
     const { orderBy: orderByName, first, last, after, before, offset, descending, ...conditions } = args
@@ -28,10 +28,12 @@ const resolveConnection = (
     const orderBy = columns.find(({ name }) => orderByName === name)
     const fromClause = getFromClause(source, args)
 
+    const primaryKeyNoOrderBy = fullPrimaryKey.filter(({ name }) => name !== orderBy.name)
+
     // Get the cursor value for a row using the `orderBy` column.
     const getRowCursorValue = row => ({
-      value: row[orderBy.name] || '',
-      primaryKey: primaryKey.map(({ name }) => row[name]),
+      value: row[orderBy.name],
+      primaryKey: primaryKeyNoOrderBy.map(({ name }) => row[name]),
     })
 
     // Transforms object keys (which are field names) into column names.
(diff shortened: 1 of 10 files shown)


src/graphql/resolveConnection.js
```diff
@@ -8,6 +8,7 @@ const resolveConnection = (
   getFromClause = constant(table.getIdentifier()),
 ) => {
   const columns = table.getColumns()
+  const fullPrimaryKey = table.getPrimaryKeys()
```
ferdinandsalis: I am not sure what you mean with **full**PrimaryKey. Though I am guessing it just means it has all the primary keys including a pk that is used to order the collection — which you later filter out. Why not just call this `primaryKeys`?

src/graphql/resolveConnection.js
```diff
@@ -51,14 +57,33 @@ const resolveConnection = (
...
+      const sql = new SQLBuilder()
+
+      const cursorCompareLHS = `(${
+        [orderBy.name, ...primaryKeyNoOrderBy.map(({ name }) => name)]
+          .map(name => `"${name}"`)
+          .join(', ')
+      })`
+
+      const cursorCompareRHS = `(${
+        [null, ...primaryKeyNoOrderBy].map((x, i) => `$${i + 1}`).join(', ')
```
ferdinandsalis: I am correct in assuming that the null is a stand in for the orderBy.name? Could it not just be orderBy.name since we only use it to increment the `$1` value.

ferdinandsalis:
👍 — Looks good to me. Great work!

src/graphql/resolveConnection.js
calebmer: Here we really just care about the length so we can generate the right number of placeholders. So `Array(1 + primaryKeyNoOrderBy.length).fill(null)` is better. Thanks for the catch!

src/graphql/resolveConnection.js
calebmer: So I (or others) didn't accidentally use it instead the "no order by variant" 😊

Do you think that's an ok justification?

Also, for historical purposes (I should probably doc this inline) the reason I filtered out `orderBy` was because I was hitting really wired bugs where `(id, id) > (1, 1)` would *include* rows with an `id` of 1 instead of excluding them as the `>` operator assumes. This was happening locally in my PG 9.5 (beta?) instance, but not in the CI 9.4 instance. In any case though I think it's useful to be deduping that comparison of possible.

ferdinandsalis: I dont understand what you mean by —
> So I (or others) didn't accidentally use it instead the "no order by variant" 😊

calebmer: I changed it anyway, I just didn’t want it to get accidently used so I named it something more opaque.

calebmer:
Released in `1.8.2` :tada:

PR #100, opened 2016-08-29
feat(server): Allow for custom logging middleware to be passed to server

philostler:
I have custom log formats configured for all my servers but PostGraphQL was dumping out it's own format that I could not configure (or change to winston as I wanted).

I've added a `logger` option to the `createServer` interface so you are free to pass in your own logging middleware, falling back to the original `morgan` option.

### DIFF
--- a/src/createServer.js
+++ b/src/createServer.js
@@ -3,7 +3,7 @@ import { forEach } from 'lodash'
 import Express from 'express'
 import onFinished from 'on-finished'
 import { Forbidden, BadRequest } from 'http-errors'
-import logger from 'morgan'
+import morgan from 'morgan'
 import favicon from 'serve-favicon'
 import finalHandler from 'finalhandler'
 import jwt from 'jsonwebtoken'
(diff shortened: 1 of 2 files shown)


src/createServer.js
```diff
@@ -29,12 +31,13 @@ const createServer = ({
   secret,
   development = true,
   log = true,
+  logger = morgan(development ? 'dev' : 'common'),
```
calebmer: Could we make this just the `log` option? So `false` disables logging, `true` uses the default `morgan` logger, and anything else will be passed directly into `server.use`?

calebmer:
Looks good, but out of curiosity, is there a reason you couldn’t just set `log` to `false` and then use your own middleware outside of PostGraphQL? Something like:

```js
const app = express()

app.use(myLogger)
app.use(postgraphql({ log: false, … }))
```

In the future I’d like the option to remove `express` as a dependency entirely and just return an HTTP middleware function to reduce footprint and potentially improve performance. So allowing users to hook into the server internals directly with middleware would mean there would be potentially breaking changes when removing `express`.

philostler:
Doh! I was so focused on the logger object internal to postgraphql that I'd completely forgotten I could just use my logger in the way you described. I don't see a need for this PR then. Sorry to take your time

calebmer:
No problem @philostler, it’s always great to hear new ideas :blush:

PR #104, opened 2016-09-04
Add David. Watching your node.js dependencies.

langpavel:
https://david-dm.org/calebmer/postgraphql


langpavel:
@calebmer You can also integrate https://greenkeeper.io/ to automatically test new if tests pass with new dependencies :-)

calebmer:
Yeah, I'd rather do Greenkeeper because it will send a notification in the form of a PR 😊

As for tests, it should just run the Travis setup which will work.

calebmer:
Enabled Greenkeeper, see #105. Thanks :+1:

langpavel:
You can have both, greenkeper and David for badges, it is good combination.. ;-)

calebmer:
You'll note that I also don't have the Travis build status button, that's because I really don't want someone's first impression to be "build failing" and/or "dependencies out of date" 😊

I definitely want to know when these things are breaking/outdated, but I consider part of the `README`s job is to be marketing material (especially because we don't have a site yet…).

I may reconsider once we put up a docs site.

PR #121, opened 2016-09-21
Fix JSON type serialization

calebmer:
JSON objects have been serializing incorrectly. This has been mentioned a couple of times like in https://github.com/calebmer/postgraphql/issues/116 or https://github.com/calebmer/postgraphql/pull/102.

This fixes the problem, but still doesn’t add arbitrary JSON to the tree as is advocated in https://github.com/calebmer/postgraphql/pull/102 :+1:

### DIFF
--- a/examples/kitchen-sink/schema.sql
+++ b/examples/kitchen-sink/schema.sql
@@ -29,7 +29,9 @@ create table another_thing (
 create table anything_goes (
   foo              int,
   bar              int,
-  interval         interval
+  interval         interval,
+  json             json,
+  jsonb            jsonb
 );
 
 create view non_mutation_view as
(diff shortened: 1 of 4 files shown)


valorize:
I am still encountering the "False boolean values always null" / unknown. But I'm already on newest master (2.3.0 I guess). 😢  Anything I could check? `function public.upsert_vote(integer, integer, integer, unknown) does not exist"` whereas the unknown is the false value.

benjie:
@valorize: are you using the `dynamicJson` setting? Can you come up with a minimal test case (i.e. schema, query and variables) that reproduces the issue?

benjie:
Or, even better would be if you could reproduce it using one of the postgraphql tests - that way we can prevent it from happening again in future 👍

valorize:
@benjie Good idea.  Which parameters do I have to supply to `$npm_bin/jest $@`?

benjie:
@valorize Check out the [`CONTRIBUTING.md`](https://github.com/calebmer/postgraphql/blob/master/CONTRIBUTING.md) file for instructions - let me know if that doesn't cover your questions as one of my missions is to improve developer onboarding to PostGraphQL 😃

Issue #126, opened 2016-09-22
Support mutations by global ID


ivosabev:
I think **postgrahql** should support the global `id` in mutations, since in Relay this is the primary source of identifying an object and here `rowId` is only provided as a convenience.

```
mutation updateTeam($input: UpdateTeamInput!) {
  updateTeam(input: $input) {
    team {
      id,
      name
    }
  }
}

{
  "input": {
    "id":"dGVhbTox",
    "newName": "My Team"
  }
}
```

calebmer:
Yep, this is definitely something that will be in PostGraphQL 2 :blush:. Also being able to insert/update/delete by any unique constraint as well similar to how you can select by any unique constraint.

The one difficulty around this is how do we support procedures? For mutations I’m fine with an `updateTeam`, `updateTeamByRowId`, `updateTeamByUniqueColumn` etc. However, for procedures that take a row as an argument do we make extra copies of that as well? To illustrate the problem, say we have a procedure and it’s a mutation:

```sql
create function do_stuff(person person, i integer) returns person …
```

What would our GraphQL mutations look like?

```graphql
type Mutation {
  doStuff(personId: ID, i: Int): DoStuffPayload
  doStuffByRowId(personRowId: ID, i: Int): DoStuffPayload
  doStuffByEmail(personEmail: Email, i: Int): DoStuffPayload
}
```

…or do we only have a mutation using the global id?

calebmer:
I also want to say thanks, keep opening issues questioning assumptions made in PostGraphQL 1. We’ve learned a lot and while I don’t want to make too many breaking changes in 2, I think it gives us a great chance to question our assumptions and make a better API :+1:

I’ll be opening an issue soon with some of the breaking changes I want to make and we can discuss them individually. There won’t be many and there will be backwards compatibility options for most of them :+1:

ivosabev:
I think mutations should be only by global `id` and the `rowId` should not be given any primary key status, but be treated as a simple database column (like email).

The only update mutation will be:
```
mutation updateStuff($input: UpdateStuffPayload!) {
  updateStuff(input: $input) {
    id,
    rowId,
    email 
  }
}
{
  "input": {
     "id": "Xja1j=",
     "email": "new@email.com"
  }
}
```

I am not sure I understand your concern with the procedures?

calebmer:
I’m going to close this as its in the PostGraphQL 2.0 beta 🎉 

[...]

Issue #125, opened 2016-09-22
Rationale of using "new" prefix in mutation input


ivosabev:
I do not understand the reasoning behind prefixing the fields with "new" inside the `UpdateInput` object. I haven't seen such behavior in any Relay app and I think it makes unnecessary assumptions about the input data.

I think most of the people will have their objects both during insert and update in a object with keys based on the table columns, rather then prefixing them with "new" in during an update.

Of course it might be a matter of style, but I think that the simplest thing will be not to make any assumptions if the user will prefix his own object with "new" and just go with the plain column names as in the `InsertInput`.

calebmer:
Yep, the rationale was because we also need keyspace for the row identifier. So for example if you had a primary key on `person_id` and `post_id` for example, you may have an input type of:

```graphql
type UpdateInput {
  personId: String
  postId: String
  newPersonId: String
  newPostId: String
  newTimestamp: Number
}
```

What other options do you see? Perhaps a:

```graphql
type UpdateInput {
  personId: String
  postId: String
  patch: UpdateInputPatch
}

type UpdateInputPatch {
  personId: String
  postId: String
  timestamp: Number
}
```

Any other ideas are welcome :+1:

ivosabev:
If we use the global ID as a primary source of truth this would be unnecessary as the global ID is recalculated on the server only after the `personId` and `postId` are updated.

So you can safely do this and pass the new values for the `personId` and `postId`, and the old value for `id` as this will be the only one available until the server returns the new calculation.

```
type UpdateInput {
  id: ID
  personId: String
  postId: String  
  timestamp: Number
}
```

calebmer:
I’m going to close this as its in the PostGraphQL 2.0 beta (#145) 🎉 

We are going with the patch approach like I mentioned, but we are also allowing you to perform mutations using the global id like you mention. Because I agree it should be the one source of truth :blush:

Start playing with the pre-release, the final release should be out really soon. Tell me what you think! :+1:

```
npm install -g postgraphql@next
postgraphql
```

Issue #114, opened 2016-09-15
Mutation Payloads and Relations


tim-field:
Lets say I have a comment table related to my posts table.

```
Post {
  commentNodesByPostId
}
```

And my post page uses `commentNodesByPostId` to show a list of comments.
I want to add a comment, so in another component I call the generated;

`insertComment` mutation

The problem I'm having is that payload of this mutation doesn't include the post that I'm related to, and because of this I haven't been able to work out how to get Relay to correctly update it's local store.  

Relay's `RANGE_ADD` seems to need `post` in the top level of the mutation payload in order to work. If it did I could follow along with this example;

https://facebook.github.io/relay/docs/guides-mutations.html#range-add-example

And have a fatQuery that looked like;

```
fragment on InsertCommentPayload {
        post { commentNodesByPostId },
        commentEdge,
}
```

Without `post` in the payload I can't get Relay to update correctly ( I can't find an incarnation of `FIELDS_CHANGE` to force the correct update either ) 

I imagine this is really common use case, how are people getting around this?

calebmer:
Yeah, a similar issue is this https://github.com/calebmer/postgraphql/issues/110 and this keeps popping up from time to time in the chatroom. I’m not sure of a good way to fix this (short of praying Relay 2 gets released quickly :wink:), I’d like to hear if other people have found workarounds.

I think what we’ll need to do is just flatten out some of the fields of the return type unfortunately. I don’t like having to do that, but I agree that might be the best/only way to make Relay happy.

If anyone is interested in writing a proposal, I’ll see what I can do to implement it :+1:

tim-field:
> I think what we’ll need to do is just flatten out some of the fields of the return type unfortunately

Yes this seems to be what's needed. Could we do this for all foreign key fields in the table being mutated. How difficult would this be? If you don't have time could you point me in the general direction in the code base.

calebmer:
The fields would be defined in the [`createInsertMutationField`’s](https://github.com/calebmer/postgraphql/blob/b443473f15b4c36a6f378b204172ed5676703d7f/src/graphql/mutation/createInsertMutationField.js#L62) payload. It should probably be as easy as taking a GraphQL field off of the type returned by [`createTableType(table)`](https://github.com/calebmer/postgraphql/blob/b443473f15b4c36a6f378b204172ed5676703d7f/src/graphql/mutation/createInsertMutationField.js#L64) and adding them to the payload. So something like:

```js
fields: {
[...]

Issue #110, opened 2016-09-07
Stored Procedure is not Relay compliant


ssomnoremac:
Because the mutation payload of a stored procedure is nested in the output field, the Relay fatQuery must look something like this:

    getFatQuery() {
        return Relay.QL`fragment on myStoredProcedurePayload {
          output {
              applicationByApplicationId
          }
        }`
    }

`output` is the only acceptable root field for the fatQuery fragment besides `viewer`. However, the fatQuery must include a field that you intend to modify in the `getConfigs()`:

    getConfigs() {
        return [{
            type: 'FIELDS_CHANGE',
            fieldIDs: { application : this.props.application.id }
        }];
    }

if the response from the stored procedure wasn't nested in the `output` field and the response fields matched the field types, all would be fine. But that isn't the case and I'm not sure it's possible to fix this using Relay.

calebmer:
What is the type signature for your procedure?

(As a side note, I can't wait for Relay 2 where mutations won't be so declarative…)

tombollich:
Hello Caleb, I'm working with @ssomnoremac on the problem. The proc returns type punchcard which is a table in the db. The issue we are having is that you are wrapping the return with output{}, so it works fine with the graphiql interface but fails with relay. Relay cannot access a nested object, output.punchcard is not allowed.

calebmer:
Why not do a `FIELDS_CHANGE` on the `output` field? This should correctly propagate changes.

As a side note, I see how the solution may seem simple (move `application` to the root mutation payload type), but it's actually much more complex and nuanced. If we have the mutation directly return the table, we wouldn't be compliant with the Relay spec, we'd also be severely limiting future design space (as we couldn't put fields `viewer` and `clientMutationId` parallel to `output`). So we need to keep `output` nested, so the next question is how and what do we need to put in the payload type alongside `output`? All the fields of `output` that may return single nodes? If we do that, then we've got other things to worry about like the query optimizations I'm working on. It's a tough problem.

If this is critical for you and running `FIELD_CHANGE` on `output` won't work let's see if we can come to an ideal solution knowing this problem may be short term, but the fix would be in for a while 👍

tombollich:
Yes I agree the problem is very complicated and nuanced, especially with all the different parts that need to work together in a sometimes hidden way (relay mostly). I believe the heart of the problem could be that most of these technologies are designed around a document database and not a relational. Anyway, I have a couple of more things I want to try so I can give you concrete examples of what is not working.

[...]

PR #145, opened 2016-10-10
PostGraphQL 2 Beta

calebmer:
# PostGraphQL 2 Beta

With this PR, I’m very excited to announce PostGraphQL 2 is now in beta 🎉

Get it now and start playing around by running (make sure you have a database running on `postgres://localhost:5432`):

```bash
postgraphql
```

PostGraphQL 2 fixes a number of long standing bugs, makes some breaking changes to reflect emerging GraphQL best practices, and sets a foundation which can be used to do truly amazing things in the future (most popularly MySQL support). To understand more of the theory behind PostGraphQL 2, read “[#87 My vision for the future of PostGraphQL](https://github.com/calebmer/postgraphql/issues/87).”

A number of breaking changes were made in PostGraphQL 2. I hate making breaking changes, especially given that GraphQL itself is designed to be change-tolerant, but I felt these were (in most cases) necessary small fixes that would be best for PostGraphQL’s feature. While this PostGraphQL 2 release does add a lot of things, remember that its real purpose is to create a strong base on which to build a data interfacing platform. Therefore, it felt like breaking changes made sense. If you disagree with any breaking changes, let me know! I would be more than happy to add a flag to provide backwards compatibility where possible. I’m hoping most breaking changes are small enough to not be a big deal, and perhaps even be desired changes. Most are inspired by dialog with the community.

## New Stuff
- **Friendly console UI.** Now when you run PostGraphQL it has colors and a much more useful request logger. See a preview image in [this tweet](https://twitter.com/calebmer/status/784898536903176192).
- **Easy to use CLI.** PostGraphQL can now be run without a single option! Just run `postgraphql` and we will automatically connect to your local database running at `postgres://localhost:5432` and the `public` schema. Before the release a `--demo` flag will also be added which will connect you to a database running in the cloud. So easy you have to try 😉
- **Authentication.** For a while PostGraphQL supported authorization so users could only see their data, but not authentication which meant you would need to build a small service to issue JSON Web Tokens. In PostGraphQL 2.0.0, this is no longer required. I’ll be writing up an authentication tutorial before the final release.
- **Better Postgres types support,** including compound types. PostGraphQL now uses ISO 8601 for dates and times (#107), custom handling of interval types, better support for Postgres ranges (#90), correct handling for all array types, and support for dynamic JSON with the `--dynamic-json` flag (#102). Compound type support means now you can create nested objects in your schemas, input or output.
- **No `rowId` fields by default.** By default, PostGraphQL no longer renames `id` columns to `rowId`. This was a controversial decision made in PostGraphQL 1 to support Relay 1, as Relay 1 requires the `id` field to be a special global id. However, it would appear that the [GraphQL team is moving towards a standard global field name in `__id`](https://github.com/facebook/graphql/issues/188). This new field name makes a lot of sense for PostGraphQL so now all global ids are named `__id` by default. However, this does not mean PostGraphQL no longer supports Relay 1! You could use the [Relay 1 fork that supports `__id`](https://github.com/facebook/relay/issues/1061#issuecomment-230721693), or you could pass the option `--classic-ids` to PostGraphQL which will use the original behavior. If you prefer the classic behavior, all power to you!
- **Related tables included in mutations.** If you are using Relay 1, it may be helpful to have access to related rows from mutations at the payload level (#110, #114, #115). In PostGraphQL 2, such rows are included in the mutation payload for create, update, delete, and procedure mutations.
- **Mutations using any unique identifier.** With PostGraphQL 2 you can execute update and delete mutations using the global `Node` identifier. The row’s primary key, or a unique constraint (#126).
- **Cursor pagination by offset.** The “correct” way to implement cursor pagination in GraphQL does not involve using SQL limits and offsets. Rather PostGraphQL must control the columns and the ordering. This makes using custom orderings incredibly difficult. Whether you had a view that was already ordered or a connection procedure that defined an order. Now that offset cursor pagination is implemented, we can support natural ordering as well as the “correct” method. Offset ordering has a few gotchas, but they aren’t bad enough to warrant not having this feature.
- **Better field names.** It has been annoying to many that the top level method of selecting rows has been named `{tableName}Nodes`, as in `personNodes` or `postNodes`. This was originally because I was against bringing in a pluralization package, mainly because it may have been awkward for non-English speakers. However, as PostGraphQL aims to be more and more extensible, it isn’t hard to change/add pluralization rules. So in PostGraphQL 2, names are not pluralized and read like `allPeople` or `allPosts`. This feels like much more natural and may always be configured.
- **Introspection of multiple schemas.** In PostGraphQL 1, you could only introspect a single schema. In 2 you can introspect as many as you would like.

## Other Breaking Changes
- No more `--development` flag. Instead all of the features in `--development` have been broken out into their own flags. The reason this was done is because sometimes a GraphiQL interface in production makes since. GraphiQL is now also enabled by default, use `--disable-graphiql` to disable it.
- GraphQL and GraphiQL live on different routes. In PostGraphQL 1 if you went to the `/graphql` route with a browser, you would see GraphiQL. This turns out to not be a best practice as it tightly couples GraphiQL to the GraphQL endpoint. If we wanted to add features to GraphiQL like a headers editor, or a SQL terminal it would be much more difficult if the routes were not separated. Whenever you start PostGraphQL you will be told the exact URL GraphiQL is served on so you won’t be confused.
- “Insert” mutations renamed to “Create.” This is to better reflect the industry standard CRUD acronym which stands for: create, read, update delete. “Insert” was initially chosen to feel more like SQL.
- No more “new” prefix in update mutation (#125). The “new” prefix in PostGraphQL 1 has gotten some understandable complaints as it has made updating rows without first renaming the fields impossible. The types for update mutations have been refactored so that now there is one `patch` object whose fields correspond to the object’s fields.
- Removed the `descending` argument in connection fields, instead descending information is contained in the `orderBy` enum. So instead of configuring `orderBy: NAME, descending: true`, you would instead just write: `orderBy: NAME_DESC` or `orderBy: NAME_ASC`.
- The root level `viewer` field has been renamed to `query` in PostGraphQL 2. The root level `viewer` field is a hack for Relay 1 to workaround some of Relay 1’s limitations when it comes to top level fields (such limitations should be gone in Relay 2). However, the `viewer` field could also have some semantic meaning such as the literal “viewer,“ or “current user” of the schema. By naming this root field `viewer`, PostGraphQL 1 disallowed users to add their own implementation for `viewer`. In PostGraphQL 2, the field still exists, but it has been renamed to `query`. It can be used in the exact same way, just now with a better name :wink:

## Steps To Release
The core codebase for PostGraphQL 2.0.0 is done and ready for production use. The reason it isn’t going straight to a release is that there are a few extra things that need to be done.

Also, I wanted to add a period of time to allow for feedback. Once we hit 2.0.0, I don’t want to make any breaking changes for as long as possible, so now would be the time to make those breaking changes!

The list of things that need to get done before we can release are as follows.

- [x] Improve forum example schema. I want to add the concept of likes to the forum example schema and a full authorization/authentication example.
- [x] Write documentation article explaining how authentication and authorization works.
- [ ] Deploy a demo database and demo PostGraphQL instance to enable starting PostGraphQL with `postgraphql --demo`.
- [x] Upgrade some existing apps using PostGraphQL 1. Including the Relay example (#66).

## What Comes Next?
So what’s the plan after PostGraphQL 2.0.0 is released? We’ll continue with my rough timeline as proposed in #87, but here it is again with a few extra things.

- Documentation/marketing website. We really need this…
- GitHub GraphQL API equivalence. GitHub recently announced their GraphQL API. How awesome would it be if we could make a PostGraphQL implementation of that API 😉
- Extensibility. This is a big one. Being able to extend PostGraphQL in Node.js is important to the future of the project.
- Performance. Experiment with different ways of executing queries to make PostGraphQL as fast as possible.
- CMS?

* * *

Please try out PostGraphQL 2 and tell me what you think! It’s as easy as (if you have a Postgres database running on `postgres://localhost:5432`):

```
postgraphql
```


ferdinandsalis:
Congratulations 🎉

src/graphql/utils/buildObject.ts
```diff
@@ -0,0 +1,50 @@
...
+  const object = {}
+
+  entriess.forEach(entries => entries.forEach(entry => {
+    if (!entry)
+      return
+
+    const [key, value] = entry
+
+    if (object.hasOwnProperty(key))
+      throw new Error(`Naming conflict when building object. Cannot have two definitions for property '${key}'.`)
```
bendiy: This error should include some more details. It would be helpful to get the object name so I know where to look when I get this error.
```

Error: Naming conflict when building object. Cannot have two definitions for property 'offset'.
    at entries.forEach.entry (/usr/lib/node_modules/postgraphql/build/graphql/utils/buildObject.js:40:19)
    at Array.forEach (native)
    at entriess.forEach.entries (/usr/lib/node_modules/postgraphql/build/graphql/utils/buildObject.js:35:41)
    at Array.forEach (native)
```

I have a large schema and I don't know what/where `offset` is at first glance.

examples/forum/schema.sql
```diff
@@ -2,111 +2,106 @@
...
+comment on column forum_example.post.updated_at is 'The latest time this post was updated.';
 
 -------------------------------------------------------------------------------
 -- Query Procedures
 
 -- Computes the full name for a person using the person’s `given_name` and a
 -- `family_name`.
-create function person_full_name(person) returns varchar as $$
-  select $1.given_name || ' ' || $1.family_name
+create function forum_example.person_full_name(person) returns text as $$
```
bendiy: `ERROR:  type person does not exist`
Need to set the schema in a few places in this file for `person` and `post`. e.g. `forum_example.person`.

src/graphql/utils/buildObject.ts
calebmer: Makes sense, how would you feel if the conflicting column names were turned into `before_`, `after_`, etc.? Another option is `before2` or `after2`. In the future I’d like to add another argument, `condition`, which allows for richer filtering and would move all of these field names out of the top level connection argument space.

examples/forum/schema.sql
calebmer: Thanks for the catch :+1:

src/graphql/utils/buildObject.ts
benjie: Another idea: you could prefix it with the table name, e.g. table `foo` column `before` -> `fooBefore` (obviously with a work around in case `foo_before` already exists as a column)

bendiy: Longer term adding `condition` to move these off the top level would be best. You'll still have the possible naming conflict, but at least it's only one. Maybe `queryCondition`?

For the short term, I like the table prefix idea `fooBefore`. But, you're looking to have a long term stable API in v2.0. Is the `condition` change going to mean v3.0?

calebmer: So we'd also reserve the name `condition` with this change with the same treatment as the others. Why do you like the idea of `fooBefore` more than `before_`? An underscore before (say `_condition`) I see as private, but an underscore afterwards I think of as a collision fix. Also, the solution must apply to Postgres function arguments as well. What if you had a Postgres function with an argument named `after`? Would it become `myFunctionNameAfter`?

bendiy: `fooBefore` and `myFunctionNameAfter` more closely follows the SQL pattern of `mytablename_mycolumnname` for naming columns. And if `foo_before` and `before` columns both exists on a table, then `before` can before `fooBefore2`.

I'm not sure what happens if `before` is processed in the for loop before `foo_before`. So maybe that won't work. Something is going to have to keep the map order consistent.

All of these tell me that the column is on the table `foo`. When I see `before_`, I'm confused. Is that something private? Is it a concatenation bug? If we're camel casing properties, what is `before_` going to be?

calebmer: So I’m less inclined to find a solution to tables with `before` and `foo_before` until it becomes a problem (assuming the table name is `foo`). `before_` would have documentation, of course so it really should not be hard to tell the difference (just open up the documentation browser in GraphiQL). If we were going to do `fooBefore`, I’d want to prefix all the names. I feel like by prefixing one and not others you give it some level of importance that just wasn’t there before, it stands out. An underscore suffix makes it different, but not that different. Make sense?

The more we talk about this, the more I think just moving it into `condition` now is the right move. It cleans up the argument list for connections (major win in my opinion and something that’s a breaking change), and allows us to punt this decision for later if it turns out we do want these conditions at the top level.

For procedures this only breaks for query Postgres functions that return a set and have a conflicting name. I can live with that erroring for now.

bendiy: I think the best solution would be to split the two into `condition` and `properties` and _never the twain shall meet_. I think you'll find most API make this split. I don't know the code or GraphQL well, so I'm not sure if that's possible.

calebmer: Ok, I moved all the equality arguments into `condition`. Haven’t published it yet, it’s just on this branch. Why don’t you check it out and tell me what you think?

calebmer: The change is now published in `2.0.0-beta.5`

calebmer:
I’ve just published a [documentation article](https://github.com/calebmer/postgraphql/blob/787bbe4ee842e0d931989df88acbdb3a88e1c380/examples/forum/tutorial.md) detailing the entire process of creating a PostGraphQL schema. It also in detail describes [authentication and authorization](https://github.com/calebmer/postgraphql/blob/787bbe4ee842e0d931989df88acbdb3a88e1c380/examples/forum/tutorial.md#authentication-and-authorization), some of which is new in PostGraphQL 2.

The last thing I need to do is setup a demo database to enable people running `postgraphql --demo` to get started then PostGraphQL 2 will be out of beta :tada:

Any last requests for breaking changes as I hope to not break anything for a while after this.

benjie:
Excellent work, @calebmer! I was reading through the first half of the tutorial and have a few tweaks to suggest:

> Postgres permissions work as a whitelist and not a blacklist, so therefore no one has more access then you explicitly give them

then -> than

> The word strict just means that this function will not accept null input. That is password cannot be null or first_name cannot be null. 

The way this reads is a little misleading, I think - strict means if you give it null input it'll give null output. I'm not sure how postgraphql 2 handles this, but v1 would just return null (and no error).

> Make sure that when you create a function with security definer there are no ‘holes’ a user could use to see or mutate more data than they are allowed to.

Missing "not"

Other than these minor niggles, it seems fine so far!

calebmer:
Ok, looks like we’re good, seems like all that’s left is bug fixes. I feel we got most of the major regressions so I’ll merge today and do the final `2.0.0` release tomorrow 🎈

benjie:
🎉  Congrats!

PR #174, opened 2016-10-20
Fix error messages

calebmer:
Found some errors that could use better error messages to avoid confusion in the future:

- Passing non-objects into the HTTP GraphQL endpoint. For example arrays, this may happen if your GraphQL client assumes query batching is supported.
- Using a relation on columns that don’t have a uniqueness constraint.

/cc @rojobuffalo, @benjie

### DIFF
--- a/src/postgres/inventory/addPgCatalogToInventory.ts
+++ b/src/postgres/inventory/addPgCatalogToInventory.ts
@@ -37,14 +37,14 @@ export default function addPgCatalogToInventory (
 
   // Add all of the relations that exist in our database to the inventory. We
   // discover relations by looking at foreign key constraints in Postgres.
+  // TODO: This implementation of relations could be better…
   for (const pgConstraint of pgCatalog.getConstraints()) {
     if (pgConstraint.type === 'f') {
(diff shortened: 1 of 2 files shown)


src/postgres/inventory/addPgCatalogToInventory.ts
```diff
@@ -56,10 +56,22 @@ export default function addPgCatalogToInventory (
...
+          `${pgCatalog.getClassAttributes(pgConstraint.foreignClassId, pgConstraint.foreignKeyAttributeNums).map(({ name }) => `'${name}'`).join(', ')} ` +
+          'of table ' +
+          `'${pgCatalog.assertGetClass(pgConstraint.foreignClassId).name}'. ` +
+          'Cannot create a relation without such a constraint. Without this ' +
+          'constraint referenced values are not ensured to be unique and ' +
+          'lookups may not be performant.'
+        )
+      }
+
+      inventory.addRelation(new PgRelation(tailCollection, headCollectionKey, pgConstraint))
```
benjie: 👍  That would have been super helpful 😃

PR #178, opened 2016-10-21
Enhance graphiql with superpowers

ferdinandsalis:
This adds an enhanced version of GraphiQL that listens to schema changes and updates its schema accordingly. Also it allows you to set a JWT token. I am sure there is plenty to improve. Grateful for input! 😀

## Todos
- [x] Add custom ui
- [x] Add tests


src/postgraphql/graphiql/graphiql.js
```diff
@@ -0,0 +1,99 @@
...
+      onChange: this.onChange,
+      onOpen: this.onOpen,
+      onError: this.onError,
+    })
+  }
+
+  render() {
+    return (
+      <GraphiQL schema={this.state.schema} fetcher={this.fetcher}>
+        <GraphiQL.Logo>PostGraphQL GraphiQL</GraphiQL.Logo>
```
calebmer: Could we just call this GraphiQL still? I’d like to keep PostGraphQL branding to a minimum for now. If we wanted, we could add a `--graphiql-logo` option to customize it and allow users to whitelabel.

src/postgraphql/graphiql/graphiql.js
```diff
@@ -0,0 +1,99 @@
...
+      </GraphiQL>
+    )
+  }
+
+  onTokenSet = event => {
+    this.setState({ token: event.target.value })
+  }
+
+  onChange = event => {
+    this.fetcher({ query: introspectionQuery })
```
calebmer: Should we `this.setState({ schema: null })` here while we are fetching the new schema? This way GraphiQL won’t display the old one.

src/postgraphql/graphiql/graphiql.js
```diff
@@ -0,0 +1,99 @@
...
+}) => {
+  // Starts listening to the event stream
+  const source = new EventSource('/_postgraphql/stream')
+  // Setup listeners for specific events
+  source.addEventListener('changed', event => onChange(event), false)
+  source.addEventListener('open', event => onOpen(event), false)
+  source.addEventListener('error', event => onError(event), false)
+}
+
+(function renderGraphiQL() {
```
calebmer: Should this be in an iifee? If so we might need a semicolon here:

```js
}

;(function renderGraphiQL () {
```

As some JS parsers may interpret it as:

```js
}(function renderGraphiQL () {
```

src/postgraphql/graphiql/index.html
```diff
@@ -0,0 +1,21 @@
...
+        overflow: hidden;
+        width: 100%;
+      }
+    </style>
+  </head>
+  <body>
+    <div id="root">
+    </div>
+    <script>window.GRAPHQL_PATH = '{GRAPHQL_PATH}'</script>
+    <script src="/_postgraphql/graphiql.js"></script>
```
calebmer: You might want to throw this in:

```html
<script src="https://cdn.polyfill.io/v2/polyfill.min.js"></script>
```

Just in case the browser doesn’t support `fetch` :+1:

src/postgraphql/http/createPostGraphQLHttpRequestHandler.d.ts
```diff
@@ -22,6 +22,8 @@ export default function createPostGraphQLHttpRequestHandler (config: {
   // The actual GraphQL schema we will use.
   getGqlSchema: () => Promise<GraphQLSchema>,
 
+  emitter: any,
```
calebmer: I also feel like there is a better solution, for now could you rename to `_emitter` or throw a `TODO` comment above it. This is definitely the best solution for now, but in the future when we add subscriptions we may have a better story for events.

src/postgraphql/graphiql/graphiql.js
```diff
@@ -0,0 +1,99 @@
+import React from 'react'
+import { render } from 'react-dom'
+import GraphiQL from 'graphiql'
+import { buildClientSchema, introspectionQuery } from 'graphql'
+import 'graphiql/graphiql.css'
```
calebmer: Could we move the CSS to just be a `<link>` tag in the HTML? This removes the need for `style-loader` and `css-loader`.

calebmer:
The build is failing for two reasons:

1. TypeScript is picking up `graphiql.dist.js` and trying to compile it.
2. `EventEmitter` should be imported as a named import, like so:

```ts
import { EventEmitter } from 'events'
```

Instead of:

```ts
import EventEmitter = require('events')
```

[Source](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/522fd939421c00d414573a427cb7b30af02886f0/node/node-4.d.ts#L512-L529)

src/postgraphql/graphiql/graphiql.js
ferdinandsalis: I am using css (modules) for the components. Are you still opposed? I might have made things a bit too complicated. I will push the new changes and you can evaluate if this is ok.

src/postgraphql/graphiql/graphiql.js
ferdinandsalis: It really doesnt need to be. Made it normal function call.

src/postgraphql/graphiql/index.html
ferdinandsalis: Will do!

src/postgraphql/graphiql/graphiql.js
ferdinandsalis: Makes sense! Done.

src/postgraphql/graphiql/graphiql.js
ferdinandsalis: I understand. Its bit of a mouth full too 😄

ferdinandsalis: What about 
screen shot 2016-10-23 at 17 33 41 😄

src/postgraphql/http/createPostGraphQLHttpRequestHandler.js
```diff
@@ -112,6 +119,58 @@ export default function createPostGraphQLHttpRequestHandler (options) {
...
+
+      // Set headers for Server-Side Events
+      res.writeHead(200, {
+        'Content-Type': 'text/event-stream',
+        'Cache-Control': 'no-cache',
+        'Connection': 'keep-alive'
+      })
+
+      // Expose function on response to write to the stream
+      res.sse = string => {
```
ferdinandsalis: Yes. That will work. My thoughts were that you could also send sse further down the line if one wanted to. But will leave that to later when we actually need it.

src/postgraphql/http/renderGraphiQL.js
```diff
@@ -1,64 +1,8 @@
-/**
- * Heavily derived from [`express-graphql`][1]. In the future we want to create
- * our own custom GraphiQL package with PostGraphQL specific features.
- *
- * [1]: https://github.com/graphql/express-graphql/blob/681b627c1a4bf01ea193ee9fe0077dd8d731bd76/src/renderGraphiQL.js
- */
+import { readFileSync } from 'fs'
```
ferdinandsalis: Its gotten a bit bigger again …

src/postgraphql/http/createPostGraphQLHttpRequestHandler.js
```diff
@@ -112,6 +120,24 @@ export default function createPostGraphQLHttpRequestHandler (options) {
...
+      return
+    }
+
+    // Setup an event stream so we can broadcast events to graphiql, etc.
+    if (parseUrl(req).pathname === '/_postgraphql/stream') {
+      if (req.headers.accept !== 'text/event-stream') {
+        res.end()
+        return
+      }
+      setupServerSentEvents(req, res, options)
```
ferdinandsalis: @calebmer I wasn’t entirely sure how to best extract that out because of the control flow stuff that is going on in the request handler. Is there a better way?

ferdinandsalis: This makes testing a bit awkward since the sse stuff is not entirely encapsulated.

src/postgraphql/http/renderGraphiQL.js
calebmer: Gotcha :+1:

src/postgraphql/graphiql/graphiql.js
calebmer: That makes sense, keep it this way then :+1:

src/postgraphql/graphiql/graphiql.js
calebmer: That’s really clever I like that 😄

However, I’m still opposed for more subjective reasons. As a user I’d probably just be annoyed. I think I’d be fine with that if PostGraphQL is being run as a CLI, but plain “GraphiQL” if it’s being run as a middleware.

src/postgraphql/http/createPostGraphQLHttpRequestHandler.js
calebmer: I think it looks fine. One way to make it more encapsulated is to perhaps make the function be a 1-to-1 representation of the event emitter. So maybe rename the function to `emitAsServerSentEvents(req, res, emitter)`? The either listen to all events from the event emitter or pass a map, something like: `new Map(['schemas:changed', { event: 'change', data: 'schema' }])` which maps event names to payloads?

ferdinandsalis: I havent come around to doing it this way. Maybe you want to take a stab if you have time.

ferdinandsalis:
Continued in https://github.com/calebmer/postgraphql/tree/feat/extend-graphiql-2

PR #192, opened 2016-10-28
fix(graphql): fix ambiguous values from node field

### LINKED ISSUE #189: Looking up node by __id returns different node?!
I created a minimal test case to reproduce the issue on my schema:

```graphql
query {
  node(__id: "WyJ0ZWFtcyIsMTBd"){
    __typename
    __id
  }
  teamById(id: 10) {
    __id
    __typename
  }
}
```

(Note base64 decode of `WyJ0ZWFtcyIsMTBd` is `["teams",10]`)

This returns

```json
{
  "data": {
    "node": {
      "__typename": "User",
      "__id": "WyJ1c2VycyIsMTBd"
    },
    "teamById": {
      "__id": "WyJ0ZWFtcyIsMTBd",
      "__typename": "Team"
    }
  }
}
```

(base64 `WyJ1c2VycyIsMTBd` = `["users",10]`)

Which makes no sense - how has a `Team` turned into a `User`? They both have the same row id: 10; so could this be the reason somehow?

calebmer:
Fixes #189

What was happening is that the `isTypeOf` check in our type system was returning true for different tables. So even though we return a value from `similar_table_2` (see updated kitchen sink) it could be interpreted as a value from `similar_table_1`.

The fix was to have the `node` field return extra type information which the `GraphQLObjectType` may then use to pick the correct type in ambiguous cases.

There is a better fix for this which involves including explicit type information with *all* values. This is a solution I hope to integrate in the [types refactor](https://github.com/calebmer/postgraphql/tree/refactor/types).

/cc @benjie


benjie:
I've just tested this and it fixes the issue I had - thanks!

calebmer:
Fixed in `2.2.1` 💫

PR #193, opened 2016-10-28
fix(graphql): fix mutation query field returning null

calebmer:
Fixes https://github.com/calebmer/postgraphql/issues/188

In order to support selecting `Query` from `node` for Relay 1, the value of `Query` was required to be a symbol named `$$isQuery`. This wasn’t always the case, originally any value for the query was acceptable so mutations would just return `null` instead of the query value. This fixes that.

/cc @benjie


valorize:
Perfekt timing. I was just experiencing this behaviour / error.

I tested it in my app and the query is populated with my fields. 🎉

benjie:
Can confirm this fixes my issues - thanks!

calebmer:
Fixed in `2.2.1` 💫

PR #186, opened 2016-10-26
Help contributors get started

benjie:
Hi, I was trying to get started contributing. The first thing I tried to do was get the tests running, but sadly some are failing for me (probably because I'm using Postgres 9.5 - not on 9.4?)

```
    -           "smallint": 50,
    -           "textArray": Array [
    -             "hey",
    -             "i",
    -             "just",
    -             "met",
    -             "you",
    -           ],
    -         },
    -       ],
    +       "nodes": null,
          },
        },
    +   "errors": Array [
    +     [GraphQLError: Cannot return null for non-nullable field Type.money.],
    +   ],
      }

      at Object.<anonymous> (src/postgraphql/__tests__/postgraphqlIntegrationOperations-test.js:44:24)
      at next (native)
      at fulfilled (src/postgraphql/__tests__/postgraphqlIntegrationOperations-test.js:4:58)
      at process._tickCallback (internal/process/next_tick.js:103:7)


Snapshot Summary
 › 1 snapshot written in 1 test suite.
 › 3 snapshot tests failed in 1 test suite. Inspect your code changes or run with `npm test -- -u` to update them.
 › 1 obsolete snapshot found, run with `npm test -- -u` to remove them.

Test Suites: 1 failed, 39 passed, 40 total
Tests:       4 failed, 495 passed, 499 total
Snapshots:   3 failed, 1 added, 17 passed, 21 total
Time:        25.045s, estimated 30s
Ran all test suites.
```

I thought I'd start by helping others get started - at first I saw no `npm test` command, but noticed `jest` was listed so just ran that directly. Turns out I ought to have built first! So I've added the familiar `npm test` command to run `./scripts/test`, and aliased a couple of the other scripts under the `npm run` namespace; and I've added a CONTRIBUTING.md file to help people get started quicker. Obviously it's severely lacking right now, but I thought it might be a good start?

PR #197, opened 2016-10-30
Help contributors get started

benjie:
It seems that CI isn't the only place that has troubles running the tests without timeouts, my ancient iMac can't complete them as fast as necessary either. To that end I've removed the CI clauses and extended the timeouts to 20 seconds which now means that none of the tests timeout on my machine any more.

I've derived from the commit messages that you're using the same format as AngularJS/Karma; I found it more pleasant to link to the Karma page rather than google docs so I've linked there. I used the following command to extract the accepted types/scopes, and then filtered them down for consistency

```
git log --pretty='%s' | grep ':' | sed 's/:.*//' | sort -u
```

Since we use both doc and docs, test and tests, I have picked the one that I feel is more consistent with the upstream commit message styling (e.g. 'docs' is preferred in types, and 'tests' is preferred in scopes).

This PR replaces #186

Feedback welcome 👍

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,3 +15,62 @@ PostGraphQL uses Travis CI to test builds and enforce lint rules:
 The codebase is documented via READMEs throughout the src folder heirarchy,
 starting with [src/README.md](src/README.md). Contributions are also encouraged
 where these files are missing or inadequate.
+
+Commit messages
+---------------
+
+PostGraphQL team use [karma-style][] commit messages: a type
+(feat/fix/chore/docs/etc), a scope (graphql/postgraphql/examples/tests) and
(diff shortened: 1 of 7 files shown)


calebmer:
Thanks so much, I’m going to add some commits to this because adding a `CONTRIBUTING.md` file is something I’ve meant to do for a while now :blush:

benjie:
@calebmer Sure thing! Would you mind fast-tracking a086f6f into master? It's making life quite complex to keep having to cherry-pick it and then rebase it back out again.

calebmer:
Yeah sure, I’ll merge this once tests pass, but if you could take a look at the commit I made to `CONTRIBUTING.md` so we can retroactively review it.

PR #198, opened 2016-10-30
Add reverse relation conditions (fixes v2 regression)

benjie:
Fixes https://github.com/calebmer/postgraphql/issues/187 (following instructions specified by @calebmer as best I can).

This is my first TypeScript ever, and I still don't have a great handle on the PostGraphQL codebase, but this was an interesting introduction because it's not much more than a refactor. It was trial-and-error on various types though, so you should check the type declarations carefully.

⚠️  I do not yet know how best to add tests for this (which you may like to have before merging - it's completely untested currently!)


benjie:
I've got one type issue I can't figure out:

```

9 const getTailConditionFromHeadValue = (relation: Nullable<Relation<THeadKey>>, headValue: any) => {
                                                                     ~~~~~~~~

src/graphql/schema/collection/createConnectionFieldFromPaginator.ts(9,68): error TS2304: Cannot find name 'THeadKey'.


 58     args.condition ? (
        ~~~~~~~~~~~~~~~~~~
 59       conditionHelpers.and(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~
...
 73     )
    ~~~~~
 74     ) : true
    ~~~~~~~~~~~~

src/graphql/schema/collection/createConnectionFieldFromPaginator.ts(58,5): error TS2322: Type 'Condition' is not assignable to type 'TInput'.
  Type 'true' is not assignable to type 'TInput'.

```

benjie:
Okay I've rewritten this a number of times and force pushed a bunch so I'm afraid it's not as tidily split into two commits as it once was; however... IT WORKS! And it passes the existing tests 😀

src/graphql/schema/collection/createConnectionFieldFromPaginator.ts
```diff
@@ -0,0 +1,90 @@
...
+    description: `A condition to be used against \`${formatName.type(type.name)}\` object types. All fields are tested for equality and combined with a logical ‘and.’`,
+    fields: buildObject<GraphQLInputFieldConfig<mixed>>(gqlConditionFieldEntries),
+  })
+
+  return {gqlConditionType, gqlConditionFieldEntries}
+})
+
+type Nullable<T> = T | null
+
+export default function createConnectionFieldFromPaginator <TSource, TItemValue, THeadKey>(
```
calebmer: This function mostly seems to be branching logic between the two different cases (with and without relation). It seems like most of the shared logic is in `getPaginatorInput`, so can we return a function that computes the majority of that from `getConditionGqlTypeAndConditionFieldEntries` and duplicate the code here in its respective places?

src/graphql/schema/collection/createConnectionFieldFromPaginator.ts
```diff
@@ -0,0 +1,90 @@
...
+    )
+
+  // Creates our GraphQL condition type.
+  const gqlConditionType = new GraphQLInputObjectType({
+    name: formatName.type(`${type.name}-condition`),
+    description: `A condition to be used against \`${formatName.type(type.name)}\` object types. All fields are tested for equality and combined with a logical ‘and.’`,
+    fields: buildObject<GraphQLInputFieldConfig<mixed>>(gqlConditionFieldEntries),
+  })
+
+  return {gqlConditionType, gqlConditionFieldEntries}
```
calebmer: From here perhaps we should return (the names can be aliased in consuming functions):

```ts
{
  gqlType: GraphQLInputObjectType,
  fromGqlInput: (input: any) => Condition,
}
```

This way we don’t have to directly return the field entries (as the field entries will only be used in `fromGqlInput`). The implementation of the `fromGqlInput` function will mostly be the current implementation you have for `getPaginatorInput` (minus the relation bit).

src/graphql/schema/collection/createConnectionFieldFromPaginator.ts
```diff
@@ -0,0 +1,90 @@
+import { GraphQLFieldConfig, GraphQLInputObjectType, GraphQLInputFieldConfig } from 'graphql'
+import { Collection, Condition, conditionHelpers, NullableType, ObjectType, Paginator, Relation } from '../../../interface'
+import { formatName, buildObject, memoize2 } from '../../utils'
+import BuildToken from '../BuildToken'
+import getGqlType from '../getGqlType'
+import createConnectionGqlField from '../connection/createConnectionGqlField'
+import transformGqlInputValue from '../transformGqlInputValue'
+
+const getConditionGqlTypeAndConditionFieldEntries = memoize2((buildToken: BuildToken, type: ObjectType) => {
```
calebmer: Yeah, let’s remove `createConnectionFieldFromPaginator` and make this the single export (it’s okay if we aren’t 100% DRY, duplicated code is better than the wrong abstraction :wink:).

Also, could we rename this to `getConditionGqlType` (and consequently the file as well)?

src/graphql/schema/collection/createConnectionFieldFromPaginator.ts
```diff
@@ -0,0 +1,90 @@
...
+  return {gqlConditionType, gqlConditionFieldEntries}
+})
+
+type Nullable<T> = T | null
+
+export default function createConnectionFieldFromPaginator <TSource, TItemValue, THeadKey>(
+  buildToken: BuildToken,
+  paginator: Paginator<Condition, TItemValue>,
+  collection: Collection,
+  relation: Nullable<Relation<THeadKey>>,
```
calebmer: As a side note, even though I’d like to remove this function for future reference to make arguments optional use a `?` next to the argument name like so :wink:

```ts
export default function createConnectionFieldFromPaginator <TSource, TItemValue, THeadKey>(
  buildToken: BuildToken,
  paginator: Paginator<Condition, TItemValue>,
  collection: Collection,
  relation?: Relation<THeadKey>,
)
```

benjie:
@calebmer I believe I've implemented those changes, ready for re-review. I've also tested and it still works with my project 😀

calebmer:
Released in `2.3.0` :tada:

PR #204, opened 2016-11-02
GraphQL-js next (will be 0.8)

langpavel:
## Whats new?

Most important is `null` value in language and in variables

[Show all commits since 0.7.1](https://github.com/graphql/graphql-js/compare/v0.7.1...master)

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "dataloader": "^1.2.0",
     "debug": "^2.2.0",
     "finalhandler": "^0.5.0",
-    "graphql": ">=0.6.0 <1.0.0",
+    "graphql": ">=0.8.0-beta1",
     "http-errors": "^1.5.0",
     "jsonwebtoken": "^7.1.9",
     "parseurl": "^1.3.1",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -39,7 +39,7 @@
     "dataloader": "^1.2.0",
     "debug": "^2.2.0",
     "finalhandler": "^0.5.0",
-    "graphql": ">=0.6.0 <1.0.0",
+    "graphql": ">=0.8.0-beta1",
```
calebmer: Is there a reason we can’t keep the dependency the way it was? The reason we defined such a wide range is so that multiple GraphQL JS instances would not conflict.

calebmer: If it’s so that we can test on `0.8.0`, could we perhaps make `0.8.0` a `devDependency` and keep the wide range in the production dependencies?

calebmer:
I’m going to merge this because it has fixes for master we could use. I’ll add the tests for `mutation-create.graphql` and `mutation-update.graphql` myself. Again, thanks so much for all your great work pushing this through the standards process :blush: :+1: :tada:

PR #209, opened 2016-11-05
Server side network layer

benjie:
To enable server side rendering (aka SSR, isomorphic, universal, etc) for a relay app you'll probably want to use something like [isomorphic-relay-router](https://github.com/denvned/isomorphic-relay-router). However, if you're using postgraphql as a middleware it seems wasteful to try and do a full HTTP request to your server to achieve this, so this PR is a proof-of-concept (and abysmal code!) to create a ServerSideNetworkLayer that enables isomorphic-relay-router to run with minimal overhead.

I'm using it like this:

```js
expressApp.use((req, res, next) => {
  req.getServerSideNetworkLayer = (done) => {
    return networkLayerFactory(getJWTToken(req, res), done);
  };
  next();
});

[...]

        req.getServerSideNetworkLayer((err, networkLayer) => {
          if (err) {
            res.status(500).send(err.message);
            return;
          }
          IsomorphicRouter.prepareData(renderProps, networkLayer).then(render).catch(next);
        });
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,8 +1,13 @@
-import { postgraphql, createPostGraphQLSchema } from './postgraphql'
+import {
+  postgraphql,
+  postgraphqlServerSideNetworkLayerFactory,
+  createPostGraphQLSchema
+} from './postgraphql'
 
 export default postgraphql
 
 export {
   postgraphql,
+  postgraphqlServerSideNetworkLayerFactory,
   createPostGraphQLSchema,
 }
--- a/src/postgraphql/http/ServerSideNetworkLayer.js
+++ b/src/postgraphql/http/ServerSideNetworkLayer.js
@@ -1,3 +1,13 @@
+import executeQuery from './executeQuery'
+import {
+  Source,
+  parse as parseGraphql,
+  validate as validateGraphql,
+  execute as executeGraphql,
+  getOperationAST,
+  formatError as defaultFormatError,
+  print as printGraphql,
+} from 'graphql'
 // See: https://facebook.github.io/relay/docs/interfaces-relay-network-layer.html
 //
 // This class implements a "network layer" for relay that can be used on the
@@ -16,7 +26,6 @@ export default class ServerSideNetworkLayer {
   }
 
   sendQueries(queryRequests) {
-    console.dir(queryRequests)
     return Promise.all(queryRequests.map(
       queryRequest => this._fetch(queryRequest).then(result => {
         if (result.errors) {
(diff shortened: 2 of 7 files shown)


calebmer:
Is there anyway we could modify this so that the network layer lives in another package or in your own app? In other words, what could we do to enable this functionality to be implemented outside of core PostGraphQL?

Currently, the execution logic for PostGraphQL can be found in [`createPostGraphQLHttpRequestHandler`](https://github.com/calebmer/postgraphql/blob/8213195efdb914b586231ebcc57bae1ba5b13869/src/postgraphql/http/createPostGraphQLHttpRequestHandler.js#L278-L308). To support this feature, we need to make sure that we can execute queries outside of an HTTP context. It looks like what you have done is a good step towards that with the `executeQuery` function, but I think the best solution is one that goes entirely through GraphQL so that when we allow for schema extension people still don’t have to run `executeQuery`.

The ideal interface (in my opinion) is that the user executes a GraphQL query as normal using the `execute` or `graphql` functions from `graphql-js`, but they can pass in a Postgres client and a JWT token through the context. Then instead PostGraphQL (instead of inside the HTTP middleware) we can setup the transaction in addition to deserializing the JWT token into the transaction. Something like:

```js
const result = await graphql(
  postgraphqlSchema,
  query,
  null,
  {
    pgClient: await pgPool.connect(),
    jwtToken: '...',
  },
  variables,
)
```

`graphql-tools` has a great [`addSchemaLevelResolve`](https://github.com/apollostack/graphql-tools/blob/master/src/schemaGenerator.ts#L238-L256) function which we should take inspiration from, our difficulty is that we also need a cleanup function. So perhaps the first thing to do is contribute `onStart`/`onEnd` hooks to `graphql-js` itself.

benjie:
Yeah, I think that's a good thing to aim for. How about in the short term we add an experimental API that takes the regular postgraphql configuration options and returns an async function which takes a `pgPool`, `jwtToken` and a graphql query (`request`) as arguments and returns the graphQL result? Basically it'd be very similar to the `_fetch` method in my PR.

Also, with the no-semicolons rule, how in typescript do you do destructuring of objects when you can't store directly into a `const`? I was getting syntax errors on build when I just did `{a, b} = foo()` so I had to wrap the statement in parenthesis, which meant that I had to start the line with a semicolon to avoid it being interpreted as a function call.

calebmer:
Another option which could be good is to develop a “wrapper” function which returns an context object for PostGraphQL. So for instance:

```js
const result = await withPostGraphQLContext({ pgPool, jwtToken }, async context => (
  await graphql(
    schema,
    query,
    null,
    {  ...context },
  )
))
```

Because we wrap the GraphQL execution we can cleanup once it is done. This allows queries to be executed outside of PostGraphQL HTTP middleware while still preserving the traditional call semantics.

The `({ a, b } = { a: 1, b: 2 })` thing is not a use case I normally have, and to be honest if you had asked me earlier I would have told you that it isn’t possible in the spec :wink:. Throwing the semicolon out front is fine though, that shouldn’t break the `no-semicolons` rule.

benjie:
That looks great, except it loses the hot-reloading of the schema. If context and schema were both passed into the inner function we could work around that, but it's less clean.

For now I think this is the right direction.

calebmer:
I think it’s fine if we keep hot-reloading as an internal concern only. We don’t need to expose that to users. Perhaps if a user did want that functionality though, we could expose it perhaps some other way :wink:

benjie:
(I've not forgotten about this.)

benjie:
Closed in favour of #269

PR #217, opened 2016-11-08
Fix daylight savings time failures

calebmer:
As @benjie mentioned in the Gitter chat, tests in this repo have probably been failing thanks to daylight savings time lol :blush:

This should fix the problem by setting our tests to run in Eastern Daylight Time instead of oscillating between Eastern Daylight Time and Eastern Standard Time, I think. Honestly, I just looked up the [Postgres timezone](https://www.postgresql.org/docs/7.2/static/timezones.html) docs and made the change through GitHub, hope this works :wink:


calebmer:
I do also want to note that my first attempt to fix this I accidently commit to master instead of creating a PR. Reverted that change, but you’ll see two unnecessary commits in the log :+1:

PR #244, opened 2016-11-21
Fixed command examples

dustinmoorenet:
`--connection` parameter needs to be present for connection string otherwise default connection is made

### DIFF
--- a/docs/default-role.md
+++ b/docs/default-role.md
@@ -31,7 +31,7 @@ create role editor nologin;
 So how does this apply to PostGraphQL? PostGraphQL requires you to have at least one user (role that can log in) when connecting to the server. That role will be specified in your connection string and will from here on out be referred to as the `auth_user`. You’d connect with your `auth_user` as follows:
 
 ```bash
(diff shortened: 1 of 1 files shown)


benjie:
(Just waiting for the build to pass so I can merge it)

benjie:
It seems Travis had a bit of a backlog issue earlier, there's no indication it received the webhook for this push.

https://www.traviscistatus.com/

I'm going to edit a character to trigger a rebuild.

benjie:
Still no sign of it https://travis-ci.org/calebmer/postgraphql/pull_requests

benjie:
Ohhhhhhhhhhhhhh

My bad. I broke the config.

benjie:
In conclusion, for tests to run against master for greenkeeper we need "build pushes" turned on; for tests to run on remote pull requests we need "build pull requests" turned on. For local pull requests (e.g. from contributors) every commit in a PR will thus be built twice and there doesn't seem to be a way to prevent that.

PR #252, opened 2016-11-25
Add live reloading to GraphiQL

calebmer:
This PR adds live-reloading to GraphiQL and turns GraphiQL into a `create-react-app`. Lots of the live reload stuff was inspired by the work of @ferdinandsalis.

I implemented this in a plane/AirBnB room while falling asleep before GraphQL summit. That’s to say it needs significant review and testing.

I’m opening this now because it would have the side effect of fixing https://github.com/calebmer/postgraphql/issues/249


ferdinandsalis:
I’ll take a look. I am happy so see some of my work was ok ☺️

scripts/build
```diff
@@ -1,5 +1,18 @@
 #!/bin/bash
 
+set -e
+
 npm_bin=$(npm bin)
 
+echo "Cleaning build"
+rm -rf build/*
```
benjie: I'd rather

```bash
rm -rf build
mkdir build
```

Reasons being:

1. any dotfiles won't be removed with `build/*` (which could include caches)
2. I don't like `*` or `/` in `rm -rf` commands that other people are expected to run

---

Also this command doesn't set it's cwd, so if you were to call it from somewhere else (e.g. in a crontab file configured to run a build every day at 4am) then it would be deleting a different folder. I'd rather see this paired with something like this to change into the project directory first:

```bash
cd "$(dirname "$(dirname "$(realpath "$0")")")"
```

scripts/dev
```diff
@@ -2,11 +2,20 @@
...
 
+# Run Nodemon and watch the PostGraphQL source code for changes.
 $npm_bin/nodemon \
   --watch src \
   --ignore __tests__ \
   --ignore __mocks__ \
+  --ignore src/postgraphql/graphiql \
   --ext js,ts \
-  --exec "$npm_bin/ts-node --ignore node_modules --disableWarnings src/postgraphql/cli.ts --schema a,b,c --show-error-stack json --watch $@"
+  --exec "$npm_bin/ts-node --ignore node_modules --disableWarnings src/postgraphql/cli.ts --schema a,b,c --show-error-stack json --watch $@" & \
```
benjie: Trailing `\` isn't required?

scripts/dev
```diff
@@ -2,11 +2,20 @@
...
+
+# Run `react-scripts` in the GraphiQL directory as well parallel, but pipe the
+# output to `/dev/null`.
+PORT=5001 npm --prefix src/postgraphql/graphiql run start > /dev/null
+
+wait
```
benjie: This may not be sufficient depending on how the script exist; might be worth adding a trap straight after the nodemon fork too:

```bash
trap 'kill $(jobs -p)' EXIT
```

src/postgraphql/graphiql/public/index.html
```diff
@@ -0,0 +1,18 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>GraphiQL</title>
+    <meta name="robots" content="noindex"/>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script>
+      window.POSTGRAPHQL_CONFIG = {
+        graphqlUrl: 'http://localhost:5000/graphql',
+        streamUrl: 'http://localhost:5000/_postgraphql/stream',
```
benjie: These should be customisable before we merge it.

src/postgraphql/graphiql/src/components/PostGraphiQL.js
```diff
@@ -0,0 +1,195 @@
...
+        return nextField
+      }
+    })
+
+    // This is very hacky but works. React is cool.
+    this.graphiql.docExplorerComponent.setState({
+      // If we are not ok, just reset the `navStack` with an empty array.
+      // Otherwise use our new stack.
+      navStack: allOk ? nextNavStack : [],
+    })
```
benjie: This is likely quite brittle; we should add a version check or something so that on an update to graphiql we check this still continues to work.

benjie: An alternative is to throw away the GraphiQL instance and build a new one which can be accomplished using `render() {... <GraphiQL key={this.state.n} ... /> ...}` and `_updateGraphiQLDocExplorerNavStack(...) {... this.setState({n: n+1})}` but that will of course lose the other things in the GraphiQL state which will likely be both annoying and ugly.

benjie: In the long term it would be better if `<GraphiQL />` took care to reset its own navStack (and related dependencies) when the `schema` prop changes; but I doubt we have time to raise a PR to do that right now!

src/postgraphql/http/createPostGraphQLHttpRequestHandler.js
```diff
@@ -21,14 +23,26 @@ const httpError = require('http-errors')
...
 const debugRequest = new Debugger('postgraphql:request')
 
+const graphiqlDirectory = resolvePath(__dirname, '../graphiql/public')
+
 const favicon = new Promise((resolve, reject) => {
   readFile(resolvePath(__dirname, '../../../resources/favicon.ico'), (error, data) => {
     if (error) reject(error)
-    else resolve(data)
+    else resolve(data.toString())
+  })
```
benjie: Why are you converting your favicon to a `utf8` string here? I think you actually want a `Buffer` here; using `fs-promise` it'd just be:

```
const favicon = fspromise.readFile(resolvePath(...))
```

src/postgraphql/http/createPostGraphQLHttpRequestHandler.js
```diff
@@ -21,14 +23,26 @@ const httpError = require('http-errors')
...
+})
+
+const origGraphiqlHtml = new Promise((resolve, reject) => {
+  readFile(resolvePath(__dirname, '../graphiql/public/index.html'), (error, data) => {
+    if (error) reject(error)
+    else resolve(data.toString())
```
benjie: It's more efficient to pass `{encoding: 'utf8'}` to `readFile` (because then it doesn't have to allocate a buffer and then immediately turn it into a string); also it'd be more convenient to use a promise-backed `readFile`; e.g. 

```
import {readFile} from 'fs-promise';
...
const origGraphiqlHtml = readFile(resolvePath(...), {encoding: 'utf8'})
```

src/postgraphql/graphiql/public/index.html
ferdinandsalis: This seems to be taken care of here https://github.com/calebmer/postgraphql/blob/8f93257769258d39dc294072d4d1d84a9cb02094/src/postgraphql/http/createPostGraphQLHttpRequestHandler.js#L95-L98

benjie: Good catch 👍

scripts/dev
calebmer: I believe it is because we are running in parallel with the `PORT=5001 npm --prefix src/postgraphql/graphiql run start > /dev/null` script below (note the `&`). I may be wrong though.

scripts/dev
calebmer: Could you submit a commit with this change? I’m not entirely sure what you mean. Here’s the edit link: https://github.com/calebmer/postgraphql/edit/feat/extend-graphiql-2/scripts/dev?pr=/calebmer/postgraphql/pull/252

src/postgraphql/graphiql/public/index.html
calebmer: I’m not a huge fan of this approach, but we need something that will both work in the `create-react-app` development environment and be replaced on serve vs. on build (thus the regex approach). Any other ideas are welcome.

src/postgraphql/graphiql/src/components/PostGraphiQL.js
calebmer: The versions of both [`graphql` and `graphiql` are locked](https://github.com/calebmer/postgraphql/blob/8f93257769258d39dc294072d4d1d84a9cb02094/src/postgraphql/graphiql/package.json#L10-L11).

The reason we create a new `navStack` is so that the documentation viewer doesn’t reset or have stale data. When we update our schema, vanilla GraphiQL does not update the `navStack` to reflect the changes as the `navStack` is an array of `GraphQLFieldConfig | GraphQLObjectType` from `graphql-js`. The old field configs and object type’s are not replaced with the new field configs and object type’s when the schema prop changes. An example of this is you are on the `Person` type page in the GraphiQL documentation viewer and the schema changes so `Person` has a new description. Without this hack the description in GraphiQL would not change. Same for adding/removing fields.

This is why we manually replace the `navStack`, we can maintain the user’s navigation state while still showing the new data from the schema. Throwing away the GraphiQL instance would lose user navigation state and potentially even their query/variables/response.

I definitely agree though that we should submit a PR to GraphiQL incorporating these changes. I was on the clock to get this done, thus why it is in PostGraphQL :wink:. I’m fine with merging it in as-is with the recognition that we will need to PR GraphiQL next time we touch this logic.

src/postgraphql/http/createPostGraphQLHttpRequestHandler.js
calebmer: Good catch. I generally try to avoid bringing in dependencies we don’t really need, so I’ll switch `favicon` to be a `Promise<Buffer>`, but let’s keep the promise conversion for now :+1:

src/postgraphql/http/createPostGraphQLHttpRequestHandler.js
calebmer: This is a good point that I did not think about. Same comment on the `fs-promise` thing though.

scripts/dev
benjie: I don't think that's how bash does things; it's simply running as a background job.

src/postgraphql/graphiql/src/components/PostGraphiQL.js
benjie: Yep, works fine for now 👍

scripts/build
calebmer: Changed :+1:

calebmer:
@ferdinandsalis, @benjie: Alright, made the suggested changes and I also rebased against master if you want to take another look. I also made some changes to the `CONTRIBUTING.md` document, because the workflow with the `create-react-app` may be non-intuitive.

Thanks again @ferdinandsalis for the event sourcing stuff and @benjie for the review :+1:

calebmer:
Also, does `postinstall` run only when installing for development? Asking because if we could automate the installation of the `create-react-app` dependencies, that would be great (currently we need to run `npm --prefix src/postgraphql/graphiql install` to get them).

calebmer:
Alternatively, would it make sense to have the GraphiQL `create-react-app` dependencies just be `devDependencies` of PostGraphQL?

scripts/dev
benjie: Done 👍  Did it through the web interface so it's untested.

benjie:
Have raised a few tweaks to this in https://github.com/calebmer/postgraphql/pull/270; feel free to cherry-pick any of them that you want on a commit-by-commit basis @calebmer

benjie:
@calebmer Now this runs on a completely different port, it separates us further from the JWT issues. Could we instead mount it as an express endpoint thus the headers would be passed through, so if there's already an Authorization header that will be passed through too (e.g. once you've logged in).

calebmer:
> @calebmer Now this runs on a completely different port, it separates us further from the JWT issues. Could we instead mount it as an express endpoint thus the headers would be passed through, so if there's already an Authorization header that will be passed through too (e.g. once you've logged in).

I’m not sure I entirely understand what you mean by this? As in create a new package `postgraphql-graphiql` that can be added as a middleware?

calebmer:
Ok, just added some commits which move GraphiQL dependencies to become `devDependencies` of the PostGraphQL package. Now they get installed with everything when running `npm install`. Also fixed the failing tests and added some new ones.

calebmer:
I also just rebased against master and force-pushed. So be careful pulling.

benjie:
I mean in my app I mount PostGraphQL as a middleware, so I can generate the required JWT token and pass it through to the postgraphql middleware as if the client had done so itself. This JWT token should flow down to graphiql but it can't if graphiql runs on a different port. By making it a middleware instead it can be mounted in a similar way (either under postgraphql or directly under my application) with any tokens passed through and without the need for another port being opened.

calebmer:
GraphiQL only runs on a different port in *internal PostGraphQL* development 😊 (I see how that can be confusing).

When built and used as a package it's served from `/_postgraphql/graphiql`. Perhaps we should clean this up so that it is always served from here? Running on a port in PostGraphQL development is nice for `create-react-app` live reload features.

benjie:
Oh I seeee; I'd not actually mounted it in my app so wasn't aware of that subtlety.

calebmer:
Does everyone feel comfortable with this, or do you think we need more changes? I added a note to `CONTRIBUTING.md` about GraphiQL.

calebmer:
Ok, going to wait until both @ferdinandsalis and @benjie give the ok. Then I’ll be comfortable merging 😊

benjie:
Good with me! 👍

martinheidegger:
I just noted that both gave their OK, maybe this pr is overlooked?

benjie:
Conflicts resolved; I think it's good to merge 👍

src/postgraphql/http/__tests__/setupServerSentEvents-test.js
```diff
@@ -0,0 +1,54 @@
...
+  const options = { _emitter, watchPg: true }
+  connection = req
+  setupServerSentEvents(req, res, options)
+}
+
+const createServer = () => http.createServer(requestHandler)
+
+test('will set the appropriate headers', async () => {
+  const server = createServer()
+  setTimeout(() => connection.emit('close'), 25)
```
benjie: https://travis-ci.org/calebmer/postgraphql/jobs/194803070

Seems this 25 millisecond timeout is not long enough for the tests:

```
TypeError: Cannot read property 'emit' of undefined
    at Timeout._onTimeout (/home/travis/build/calebmer/postgraphql/src/postgraphql/http/__tests__/setupServerSentEvents-test.js:26:59)
    at ontimeout (timers.js:365:14)
    at tryOnTimeout (timers.js:237:5)
    at Timer.listOnTimeout (timers.js:207:5)
 FAIL  src/postgraphql/http/__tests__/setupServerSentEvents-test.js
  ● Test suite failed to run
    ProcessTerminatedError: cancel after 2 retries!
      
      at Farm.<anonymous> (node_modules/worker-farm/lib/farm.js:81:25)
      at Array.forEach (native)
      at Farm.<anonymous> (node_modules/worker-farm/lib/farm.js:75:36)
      at ontimeout (timers.js:365:14)
      at tryOnTimeout (timers.js:237:5)
      at Timer.listOnTimeout (timers.js:207:5)
```

I think what's happening is the `requestHandler` is not called by the time the timeout fires, meaning `connection` is not defined and hence `connection.emit('close')` throws.

calebmer:
Yeah. I’ve been dragging my feet on this one because it’s a big change 😖

I’ll get this merged and released over the weekend 👍

PR #268, opened 2016-12-06
Adding link to forum sample tutorial from top level readme

kenota:
I think this readme gives perfect introduction to postgraphql with JWT so it worth having direct link from  project top-level readme.


benjie:
Great, thanks @kenota 👍

PR #269, opened 2016-12-07
feat(postgraphql): enable running PostGraphQL queries outside HTTP

benjie:
This replaces #209. It enables you to create a server side NetworkLayer for doing isomorphic rendering in Relay with PostGraphQL.

I'm using it like this:

```js
import {withPostGraphQLContext} from 'postgraphql'
import {
  Source,
  execute as executeGraphql,
  parse as parseGraphql,
  validate as validateGraphql,
} from 'graphql';

export default class ServerSideNetworkLayer {
  constructor(pgPool, gqlSchema, jwtToken, options = {}) {
    this.pgPool = pgPool;
    this.gqlSchema = gqlSchema;
    this.jwtToken = jwtToken;
    this.options = options;
  }

  sendMutation(mutationRequest) {
    return Promise.reject('Mutations not supported on the server yet');
  }

  sendQueries(queryRequests) {
    return Promise.all(queryRequests.map(
      async queryRequest => {
        try {
          const result = await this._fetch(queryRequest);
          if (result.errors) {
            queryRequest.reject(new Error("Query failed"));
          } else {
            queryRequest.resolve({response: result.data});
          }
        } catch(e) {
          queryRequest.reject(e);
        }
        return null;
      }
    ))
  }

  supports(...options) {
    return false;
  }

  // Private:

  async _fetch(request) {
    const query = request.getQueryString();
    const variables = request.getVariables();
    const source = new Source(query, 'GraphQL Http Request');
    const operationName = request.getDebugName();
    const queryDocumentAst = parseGraphql(source);
    const validationErrors = validateGraphql(this.gqlSchema, queryDocumentAst);
    if (validationErrors.length > 0) {
      return {errors: validationErrors};
    }
    return await withPostGraphQLContext(
      this.options,
      {
        pgPool: this.pgPool,
        jwtToken: this.jwtToken,
      },
      async context => (await executeGraphql(
        this.gqlSchema,
        queryDocumentAst,
        null,
        context,
        variables,
        operationName
      ))
    )
  }
};
```

```js
import IsomorphicRouter from 'isomorphic-relay-router';
import {Pool} from 'pg';
import ServerSideNetworkLayer from './ServerSideNetworkLayer';
import {parse as connectionStringParse} from 'pg-connection-string';
import {createPostGraphQLSchema} from 'postgraphql';

const options = {
  dynamicJson: true,
  pgDefaultRole: '...',
  jwtSecret: process.env.JWT_SECRET,
  disableQueryLog: process.env.NODE_ENV === 'production',
};


// in an `async` method:
    const pgPool = new Pool(connectionStringParse(process.env.POSTGRAPHQL_DATABASE_URL));
    const postgraphqlSchema = await createPostGraphQLSchema(process.env.POSTGRAPHQL_DATABASE_URL, '...', options);

    const networkLayerForJwtToken = (jwtToken) => new ServerSideNetworkLayer(
      pgPool,
      postgraphqlSchema,
      jwtToken,
      options
    );


// ... middleware (req, res, next) => {...
        match({routes, location: req.originalUrl}, (error, redirectLocation, renderProps) => {
          //... error handling
          const networkLayer = networkLayerForJwtToken(getJWTToken(req, res));
          IsomorphicRouter.prepareData(renderProps, networkLayer).then(render).catch(next);

          function render({data, props}) {
            const reactOutput = ReactDOMServer.renderToString(IsomorphicRouter.render(props));
            // ...
          }
```

A compiled version can be installed with `npm install` (or `yarn add`) `calebmer/postgraphql#ignore/bg`


src/postgraphql/withPostGraphQLContext.js
```diff
@@ -0,0 +1,45 @@
...
+export default async function withPostGraphQLContext(options, {pgPool, jwtToken}, functionToRun) {
+  // Connect a new Postgres client and start a transaction.
+  const pgClient = await pgPool.connect()
+
+  // Enhance our Postgres client with debugging stuffs.
+  debugPgClient(pgClient)
+
+  // Begin our transaction and set it up.
+  await pgClient.query('begin')
+  await setupPgClientTransaction(jwtToken, pgClient, options)
```
calebmer: Can we move `setupPgClientTransaction` into this function and then call `withPostGraphQLContext` from [here](https://github.com/calebmer/postgraphql/blob/c43ab8d852b4f83704571287c64662b1a3e6d915/src/postgraphql/http/createPostGraphQLHttpRequestHandler.js#L277-L308). DRY it up a little and allows us to internally dogfood the solution :wink:

src/postgraphql/withPostGraphQLContext.js
```diff
@@ -0,0 +1,45 @@
...
+        null,
+        {  ...context },
+        variables,
+        operationName,
+      )
+    ))
+
+*/
+
+export default async function withPostGraphQLContext(options, {pgPool, jwtToken}, functionToRun) {
```
calebmer: Could we add some types to this? Also, could we merge `options` and the `{ pgPool, jwtToken }` object into one? This is generally the pattern we’ve gone with for other functions.

calebmer: Also, what about renaming this to `runPostGraphQLInContext`?

benjie: @calebmer  I'd rather you added types, I'm not experienced enough with TypeScript to declare them idiomatically. I'll learn a lot by seeing how you do them, too 👍

benjie:
Yeah, I really want to write up about my isomorphic solution but, as with you, time is the limiting factor!

calebmer:
Ok, I’ll add the types tomorrow and see what I can do about drying up `withPostGraphQLContext` tomorrow 👍

src/postgraphql/withPostGraphQLContext.js
benjie: I think

```js
    withPostGraphQLContext({ ...options, pgPool, jwtToken }, async context => (
      await graphql(
        schema,
        query,
        null,
        {  ...context },
        variables,
        operationName,
      )
    ))
```

makes more sense than

```js
    runPostGraphQLInContext({ ...options, pgPool, jwtToken }, async context => (
      await graphql(
        schema,
        query,
        null,
        {  ...context },
        variables,
        operationName,
      )
    ))
```

I could see instead `runInPostGraphQLContext` since you're running a method in a postgraphql context; however I think `withPostGraphQLContext` is probably fine.

benjie: I've merged options into the object, but it feels weird putting dynamic content (jwtToken) into an options object.

src/postgraphql/withPostGraphQLContext.js
benjie: Done 👍  I had to disable a bunch of tests because they reference the now removed method, that'll likely have to have something done about but I don't really have time to re-implement all those tests.

benjie:
I'm not sure but I think the reason this is failing tests is because it *fixes* behaviour - was it not correctly setting the pgDefaultRole before?

```js
*     const pgDefaultRole = Symbol('pgDefaultRole')
      const server = createServer({ jwtSecret, pgDefaultRole })
      await (
        request(server)
        .post('/graphql')
        .send({ query: '{query}' })
        .expect(200)
        .expect('Content-Type', /json/)
        .expect({ data: { query: null } })
      )
      expect(pgPool.connect.mock.calls).toEqual([[]])
*     expect(pgClient.query.mock.calls).toEqual([['begin'], [], ['commit']])
```

^ This middle empty array in the bottom line should have the pgDefaultRole set, shouldn't it?

```
http › will setup a transaction and pass down options for requests that use the Postgres client
    expect(received).toEqual(expected)
    
    Expected value to equal:
      [["begin"], Array [], ["commit"]]
    Received:
      [["begin"], [{"text": "select set_config($1, $2, true)", "values": ["role", Symbol(pgDefaultRole)]}], Array [], ["commit"]]
    
    Difference:
    
    - Expected
    + Received
    
    @@ -1,9 +1,18 @@
     Array [
       Array [
         "begin",
       ],
    +  Array [
    +    Object {
    +      "text": "select set_config($1, $2, true)",
    +      "values": Array [
    +        "role",
    +        Symbol(pgDefaultRole),
    +      ],
    +    },
    +  ],
       Array [],
       Array [
         "commit",
       ],
     ]
```

src/postgraphql/http/__tests__/createPostGraphQLHttpRequestHandler-test.js
```diff
@@ -1,8 +1,5 @@
-jest.mock('../setupRequestPgClientTransaction')
```
calebmer: It looks like this is why the tests were failing 😊

Before we were mocking the actual setup behavior so the function would not actually run. I’m fine with this change 👍

src/postgraphql/withPostGraphQLContext.ts
```diff
@@ -0,0 +1,223 @@
+import createDebugger = require('debug')
```
calebmer: Refactored `withPostGraphQLContext` to add types, and I moved `debugPgClient` and `setupPgClientTransaction` to this file because they weren’t being used anywhere else 🤷‍♂️

benjie:
👍

calebmer:
Just added some documentation if you want to proof read.

docs/library.md
```diff
@@ -60,3 +68,80 @@ Arguments include:
 [graphql/express-graphql#82]: https://github.com/graphql/express-graphql/pull/82
 [`pg`]: https://www.npmjs.com/pg
 [morgan]: https://www.npmjs.com/morgan
+
+## Custom Execution
+
+The PostGraphQL middleware gives you a lot of excellent features for running your own GraphQL server. However, if you want to execute a PostGraphQL query in Node.js without having to go through HTTP you can use some other exported functions that PostGraphQL provides.
+
+The first function you will need is `createPostGraphQLSchema` who’s purpose is to create your PostGraphQL schema. This function is asynchronous as it will need to run the Postgres introspection query in your database.
```
benjie: I think it might be `whose` rather than `who's`.

docs/library.md
```diff
@@ -60,3 +68,80 @@ Arguments include:
...
+  - `dynamicJson`: Setting this to `true` enables dynamic JSON which will allow you to use any JSON as input and get any arbitrary JSON as output. By default JSON types are just a JSON string.
+  - `disableDefaultMutations`: Setting this to `true` will prevent the creation of the default mutation types & fields. Database mutation will only be possible through Postgres functions.
+  - `jwtPgTypeIdentifier`: The Postgres type identifier for the compound type which will be signed as a JWT token if ever found as the return type of a procedure. Can be of the form: `my_schema.my_type`. You may use quotes as needed: `"my-special-schema".my_type`.
+
+### `withPostGraphQLContext(options, callback): Promise<GraphQLExecutionResult>`
+
+With this function you are required to provide some options and a callback. The callback is expected to return a GraphQL execution result. Whatever value is returned by the callback will then be returned by `withPostGraphQLContext`. The context you get as an argument to `callback` will be invalid anywhere outside of the `callback` function.
```
benjie: This should talk about Promises. How about something like:

This function sets up a PostGraphQL context, calls (and resolves) the callback function within this context, and then tears the context back down again finally resolving to the result of your function. The callback is expected to return a promise which resolves to a GraphQL execution result. The context you get as an argument to `callback` will be invalid anywhere outside of the `callback` function.

docs/library.md
```diff
@@ -60,3 +68,80 @@ Arguments include:
...
+
+- **`options`**: An object of options that are used to create the context object that gets passed into `callback`.
+  - `pgPool`: A required instance of a Postgres pool from [`pg-pool`][]. A Postgres client will be connected from this pool.
+  - `jwtToken`: An optional JWT token string. This JWT token represents the viewer of your PostGraphQL schema.
+  - `jwtSecret`: The secret for your JSON web tokens. This will be used to verify the `jwtToken`.
+  - `pgDefaultRole`: The default Postgres role that will be used if no role was found in `jwtToken`. It is a best security practice to always have a value for this option even though it is optional.
+- **`callback`**: The function which is called with the `context` object which was created. Whatever the return value of this function is will be the return value of `withPostGraphQLContext`.
```
benjie: I'm concerned the term `callback` confuses that we've got promises going on. The term is technically correct, but I think it may be confusing to people new to promises (who often think you either use a promise or a callback).

I can't think of a better term though; so it's fine as-is 👍

benjie:
LGTM; comments added

docs/library.md
calebmer: I like it 👍

calebmer:
Released in `2.6.0` 🎉

PR #270, opened 2016-12-07
Tweaks to GraphiQL PR

benjie:
Note this merges into the `feat/extend-graphiql-2` branch (not master).

- switches from port 5001 (which conflicted for me so likely will for others) to 5783
- adds a postinstall script to build/dev that detects if the graphiql dependencies have ever been installed and if not installs them (does NOT handle later updates, but it's a good first stab I think)
- fixes a minor issue where the background process killer is too noisy

### DIFF
--- a/scripts/dev
+++ b/scripts/dev
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+set -e
+cd $(dirname $0)/..
+
+./scripts/postinstall
+
 npm_bin=$(npm bin)
 
 export POSTGRAPHQL_ENV=development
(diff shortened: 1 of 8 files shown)


benjie:
Oh; and I've fixed the lint issues. Personally I'd disable the `object-literal-key-quotes` rule, but I've fixed it instead. I think console statements in that file are legit, and since it's intended for developers I've just disabled them throughout the whole file.

src/postgraphql/graphiql/src/components/PostGraphiQL.js
```diff
@@ -56,10 +57,10 @@ class PostGraphiQL extends React.Component {
     const response = await fetch(POSTGRAPHQL_CONFIG.graphqlUrl, {
       method: 'POST',
       headers: Object.assign({
-        'Accept': 'application/json',
+        Accept: 'application/json',
```
calebmer: Yeah, I don’t like this quotes rule either…don’t know how it got set to enforce this. It’s fine for now though.

scripts/dev
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+set -e
+cd $(dirname $0)/..
+
+./scripts/postinstall
```
calebmer: There’s got to be a better way to do this `postinstall`. I’m going to try moving all of the GraphiQL dependencies into the project root and see how that goes. Makes it harder to extract `graphiql` into its own package eventually, but it would avoid some of these workarounds.

src/postgraphql/graphiql/src/components/PostGraphiQL.js
```diff
@@ -1,3 +1,4 @@
+// tslint:disable no-console
```
calebmer: I may switch this to disable on a line-by-line basis.

scripts/dev
benjie: The issue with that is you have to remember to synchronise the package.json files; but I suppose you could write a test that enforces that so 👍

PR #278, opened 2016-12-15
Remove isTypeOf check from collection types

calebmer:
A small refactor that removes the main reason we need an `isTypeOf` function in collection types. This gives us a super small performance win, but also makes it easier to implement computed column performance wins and selective column selections.


benjie:
I learned so much just reviewing this. It looks like it should work to me (and who doesn't like PRs that delete more lines than they add?!) but I don't feel confident enough in my understanding of GraphQL to specifically approve these changes.

_/me passes the buck_

PR #281, opened 2016-12-16
Refactor types

calebmer:
This is a PR I’ve been working on every now and then when I get the time. It’s a full refactor of the types for the abstract interface we’ve defined. Once complete, this refactor should enable us to do more optimizations on types and provide even better support for Postgres types.

There are three files you really need to see to understand the refactor:

- https://github.com/calebmer/postgraphql/blob/4b5aa1d9d0e0518499d185922bc3983be586b031/src/interface/type/switchType.ts
- https://github.com/calebmer/postgraphql/blob/4b5aa1d9d0e0518499d185922bc3983be586b031/src/graphql/schema/type/getGqlInputType.ts
- https://github.com/calebmer/postgraphql/blob/4b5aa1d9d0e0518499d185922bc3983be586b031/src/graphql/schema/type/getGqlOutputType.ts

There are two things we’re trying to clean up here:

1. The type system used abstract classes and not interfaces. This meant implementations had to extend from the abstract classes. However, when it comes to something like a Postgres type, it would be advantageous for Postgres types to all extend from a common `PgType` class instead of the abstract class their type was a kind of. Also, as classes, we also could only assume that we had a certain number of type kinds. So there’s a lot of places where we have defensive programming to pass type-checking. `switchType` solves that problem.
2. The type-system (TypeScript) types were sloppy especially around GraphQL input. Object values were always `Map<string, mixed>`s (or `ObjectType.Value`s as we referred to them in the codebase). This led to a lot of `mixed` types and unnecessary `isTypeOf` checks. With the new implementation of `getGqlInputType` this has been cleaned up.


calebmer:
This is not ready to be merged. I haven’t touched `src/postgres` and `src/postgraphql` yet :blush:

calebmer:
Oh, and we’re also moving off our custom GraphQL-JS types to the community maintained [`@types/graphql`](http://npmjs.com/@types/graphql).

calebmer:
I’m just going to merge this unless someone has significant comments 👍

PR #285, opened 2016-12-19
Increase the size limit of the request's body

haikyuu:
The default limit of a request's body is `100kb`, which is not enough if you want to save something big (an image for example) in the database.
[This is the link](https://github.com/expressjs/body-parser#limit) to the `limit` option in the docs.

### DIFF
--- a/src/postgraphql/http/createPostGraphQLHttpRequestHandler.js
+++ b/src/postgraphql/http/createPostGraphQLHttpRequestHandler.js
@@ -70,9 +70,14 @@ export default function createPostGraphQLHttpRequestHandler (options) {
   // want that.
   const bodyParserMiddlewares = [
     // Parse JSON bodies.
-    bodyParser.json(),
+    
+    // Limit: Controls the maximum request body size. If this is a number,
+    // then the value specifies the number of bytes; if it is a string,
+    // the value is passed to the bytes library for parsing.
+    // Defaults to '100kb'
+    bodyParser.json({ limit: '50mb' }),
     // Parse URL encoded bodies (forms).
-    bodyParser.urlencoded({ extended: false }),
+    bodyParser.urlencoded({ extended: false, limit: '50mb' }),
     // Parse `application/graphql` content type bodies as text.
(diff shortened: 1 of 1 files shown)


calebmer:
Relay solves this problem by allowing to attach files. See the mutation config [`getFiles`](http://facebook.github.io/relay/docs/api-reference-relay-mutation.html#getfiles) function. Could we perhaps support that behavior instead of extending the request body limit? My fear with extending the body limit is that it could allow attackers to abuse the system and insert insanely large things.

I'd also like to know the logic for why the default is what is. There must be some reason, right?

haikyuu:
Yes, [why is the parser limit defaulted to 100kb?](https://github.com/expressjs/body-parser/issues/129). There was a [security issue](https://nodejs.org/en/blog/vulnerability/v8-memory-corruption-stack-overflow/) and there is a hit on performance since [parsing the body blocks the event loop](https://github.com/expressjs/body-parser/issues/132).
I think it's a much better idea to implement `getFiles` like in relay.

I am not particularly familiar with Relay. Any thoughts on how we could do that?

calebmer:
It looks like when Relay detects the user has defined some files it switches from sending JSON to `FormData` with a `Content-Type` of `multipart/form-data`. The same format used by HTML5 forms. See: https://github.com/facebook/relay/blob/30162b25194705e770301922023f1d03bf1791e7/src/network-layer/default/RelayDefaultNetworkLayer.js#L93-L111

An example of that format can be found in the HTML4 specification: https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4

For parsing multipart form bodies, it looks like `body-parser` won't do it, but it does list alternatives: https://www.npmjs.com/package/body-parser

calebmer:
This MDN article is helpful for context: https://developer.mozilla.org/en-US/docs/Web/API/FormData/Using_FormData_Objects

calebmer:
Going to close this PR. While we should still consider our approach to uploading files that discussion probably belongs to another issue. We are also adding the size limit as an option in this PR: https://github.com/calebmer/postgraphql/pull/379

PR #286, opened 2016-12-19
fix(postgres): make pg range objects nullable

ferdinandsalis:
See failing test on travis https://travis-ci.org/calebmer/postgraphql/jobs/185117607


calebmer:
Yeah, [here](https://github.com/calebmer/postgraphql/blob/d1e1024a52db032461614147a6472ace4a10f61b/src/postgres/inventory/type/PgRangeObjectType.ts#L83) we aren't handling null liberals. The fix should be fairly simple though.

ferdinandsalis:
I would prefer if someone else would tackle this … I dont yet feel confident enough with Typescript.

ferdinandsalis:
@calebmer I have fixed part of the problem by making the PgRangeObject nullable.

ferdinandsalis:
I cant update the failing snapshot of the introspection since this test has been failing for me some reasons I don’t understand,
```
*  will get everything needed in an introspection
    No class was found with id 26773
```

ferdinandsalis:
I am confused since it seems to work without changing the suggested `transformPgValueIntoValue`. I made sure by adding a line to the query fixtures.

Also sorry for the noise ☺️ I will try to be more succinct.

ferdinandsalis:
re: updating the snapshot. see https://github.com/calebmer/postgraphql/pull/286#issuecomment-268311044. Could you go ahead and do it for me. Also by any chance do you know why this tests fails for me.

benjie:
@ferdinandsalis and @calebmer: I've updated the snapshot (had to exclude my failing money tests which are still throwing `Cannot return null for non-nullable field Type.money`) but I'm not sure if you want to review the diff again before merging?

calebmer:
Released in `2.4.1` :tada:

ferdinandsalis:
@benjie thanks for updating the snapshot ☺️ did not think about excluding the failing test …

Happy Holidays everyone!

benjie:
I'm addicted to `--patch` in `git`; I use it with everything: `git add`, `git checkout`, `git stash`, `git reset`; so omitting the failing tests was just a couple of `n`s 😀 

Happy Holidays to you too 🎉

Issue #292, opened 2016-12-26
Auth0 JWT gives a jwt audience invalid. expected: postgraphql


FGRibreau:
Hello!

I'm experimenting around PostgraphQL for a new startup and it seems at the time being it requires clams.sub to be "postgraphql" while Auth0 gives `"sub": "auth0|70522e86-1fcf-415f-aeda-9325f3413adf"` (auth0|user_id), @calebmer I can't find a CLI parameter to change this, any idea to make PostgraphQL work with Auth0? :)

```json
{
  "errors": [
    {
      "message": "jwt audience invalid. expected: postgraphql"
    }
  ]
}
```

calebmer:
We could just remove the `sub` requirement, or make it optional with a command line flag. I was thinking about this use case when I made that requirement in effect to defer the decision :blush:

In my experience of integrating Firebase auth with another authentication system, I would always have two tokens. One for Firebase, and one for the app’s auth system. However I’m sure we can come to a more efficient pattern here.

Do you know if Auth0 will work out of the box with PostGraphQL if this requirement is removed?

FGRibreau:
@calebmer it should work yes, because it's the last thing that is blocking me from using postgraphql on my project, everything else works well on postgrest (I will just have to change my stored function to handle both way to access to `.sub` (postgrest use `postgrest.claims.sub` while postgraphql is `jwt.claims.sub`) and thus give an REST api **and** a GraphQL api to my project which would be purely awesome)

ugiacoman:
Going to be implementing Auth0 + Postgraphql very soon. Would really appreciate the sub requirement to be optional 👍

FGRibreau:
I think we just need to make the `{ audience: 'postgraphql' }` optional in https://github.com/calebmer/postgraphql/blob/e72696de3d67f5d478c009f8be4c9b25fdb1e2ed/src/postgraphql/http/setupRequestPgClientTransaction.js#L39

ugiacoman:
As long the secret is verified, JWT should be pretty safe. May I ask, why the audience is being verified?

angelosarto:
I have been playing around with this as well and I have some observations/ideas.  I am not quite seeing the same issue with sub, rather I saw it with aud(ience).  I was able to work around this in auth0.

## Auth 0 steps:

1. enable the APIs beta feature of Auth0
2. "create API" 
     1.  the Identifier field must be set to 'postgraphql' (because of the audience check currently in postgraphql.)  
[...]

PR #297, opened 2016-12-29
Add Optional Param for JWT Audience + Auth0 roles Support

ugiacoman:
### Custom JWT Audience
Currently, Postgraphql requires that any JWT have an audience of `postgraphql`. With this PR, the JS API now allows for a custom audience to be set. If none is provided, it will default to `postgraphql`.

### Auth0 roles Support
Currently, Auth0 has custom rules to add properties to your JWT. The default implementation of the `role` rule, adds an array of `roles`. With this PR, Postgraphql will check for a `role` property, but if a `roles` array exists, it will override and choose the first role in the array.

Example JWT from Auth0 with default implemenation of the `roles` array.
```json
{
   ...
  "roles": [
    "user_role"
  ],
  "clientID": "exampleClientID",
  "created_at": "2016-12-27T17:50:30.013Z",
  "iss": "exampleISS",
  "sub": "exampleSUB",
  "aud": "exampleAUD",
  "exp": 1483016195,
  "iat": 1482980195
}
```

A discussion of this can be found in the following issue: #292

### DIFF
--- a/docs/library.md
+++ b/docs/library.md
@@ -50,6 +50,7 @@ Arguments include:
   - `graphiqlRoute`: The endpoint the GraphiQL query interface will listen on (**NOTE:** GraphiQL will not be enabled unless the `graphiql` option is set to `true`). Defaults to `/graphiql`.
   - `pgDefaultRole`: The default Postgres role to use. If no role was provided in a provided JWT token, this role will be used.
   - `jwtSecret`: The secret for your JSON web tokens. This will be used to verify tokens in the `Authorization` header, and signing JWT tokens you return in procedures.
+  - `jwtAudience`: The audience for your JSON web tokens. If not specified, it will default to `postgraphql`.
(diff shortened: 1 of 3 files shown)


calebmer:
I see you closed this, any reason?

ugiacoman:
@calebmer Wasn't passing the tests, should've ran them before making the PR. Once I get them passing, I'll reopen.

jinty:
I also managed to hit this issue when using a JWT from another service. I think it will be a common issue people hit and make adoption just a more difficult.

So if at all possible, please set the default to not require an "audience" parameter to prevent new users stumbling over this. I'll note that rfc7519 mentions that the audience claim is optional.

jamiemccrindle:
I too have just hit this issue. It would be great if audience could be configurable or optional

calebmer:
I think this was a good implementation. If anyone wants to open a PR with a similar implementation and get it to pass the tests I’d be happy to merge 😊

ugiacoman:
@jamiemccrindle @calebmer Created a new PR without the `roles` support. Let me know your thoughts, ran the tests before I pushed this time haha #365

PR #305, opened 2017-01-08
Refactor: build to es5 instead of es6

calebmer:
My attempt at getting PostGraphQL to target ES5 instead of ES6 :wink:

Inspired by https://github.com/calebmer/postgraphql/issues/301 and https://github.com/calebmer/postgraphql/pull/304

There isn’t really a good reason for PostGraphQL to target ES6 instead of ES5. Mostly I wanted to see if we could get away with building to ES6 to get some likely minor performance wins. For a long time as it turned out we could get away with building to ES6, but as more people want to deploy to AWS Lambda (which is super cool!) we need to support Node.js 4.

Basically I couldn’t find a good reason to not target ES6. Deployment to AWS Lambda is a very good reason :wink:.

The migration should be as simple as changing the build target. In TypeScript 2.0 it appears that async functions in ES5 are not supported so we also have to upgrade to TypeScript 2.1. With the 2.1 upgrade it appears like the symbol primitive type was renamed to `symbol` from the name we were using, `Symbol`. In ES5 it also looks like for-of loops don’t support iterators so we need to wrap our `Map`s that were once in for-of loops with `Array.from`.


calebmer:
It’s also worth noting we depend on a lot of things from the ES6 standard library like `Map`, `Set`, and `Promise`. If these primitives are not in your environment the a polyfill may be installed.

rentrop:
@calebmer awesome. Thank you very much.
I would suggest using `node v4.3.2` in travis.
In the travis-config

```
node_js:
    - "4"
```

leads to `node v4.7.2` (see travis-log) whereas AWS Lambda is using [v4.3.2](http://docs.aws.amazon.com/lambda/latest/dg/current-supported-versions.html).

calebmer:
@rentrop done :+1:

resources/jest-preprocessor.js
```diff
@@ -1,12 +1,17 @@
...
 
 module.exports = {
   process (src, fileName) {
     // Make sure that source maps are always enabled and inline so that they
     // can be picked up by the `source-map-support` package.
-    return tsc.transpileModule(src, { compilerOptions, fileName }).outputText
+    return tsc.transpileModule(src, {
+      compilerOptions: compilerOptions,
+      fileName: fileName,
+    }).outputText
```
benjie: I'm assuming these changes are just necessary to run this file direct on Node 4 since it doesn't support destructuring and object literal shorthand syntax.

src/postgres/inventory/conditionToSql.ts
```diff
@@ -25,7 +25,12 @@ export default function conditionToSql (condition: Condition, path: Array<string
     case 'GREATER_THAN':
       return sql.query`(${sql.identifier(...path)} > ${sql.value(condition.value)})`
     case 'REGEXP':
-      return sql.query`regexp_matches(${sql.identifier(...path)}, ${sql.value(condition.regexp.source)}, ${sql.value(condition.regexp.flags)})`
+      // Parse out the regular expression. In Node.js v4 we can’t get the
+      // `flags` property so we need to do it this way.
+      const match = condition.regexp.toString().match(/^\/(.*?)\/(g?i?m?u?y?)$/i)
```
benjie: This seems to be dependent on the order of the flags, `/^\/(.*)\/([gimuy]*)$/i` may be better:

- no need for non-greedy matching
- allows the flags in any order

resources/jest-preprocessor.js
calebmer: Yep. That got me at first, but this file does not run through a build step so we can’t use new syntax if we want to run CI for Node.js v4.

src/postgres/inventory/conditionToSql.ts
calebmer: True. This was a really unfortunate incompatibility to discover. What I’d really like to see is an ES5 polyfill that implements `flags` that we can check our implementation against.

Ironically though, I don’t think this even matters because this part of the code never gets run :blush:. It’s for the advanced conditions feature that we don’t yet have.

benjie:
Approved 👍

PR #306, opened 2017-01-09
Feat: Build schema from prebuild PgCatalog

rentrop:
This enables the usage of a prebuild PgCatalog e.g. on AWS Lambda (https://github.com/calebmer/postgraphql/issues/301)


rentrop:
Sounds good to me.

Issue #287, opened 2016-12-20
Token handling on expired token


valorize:
When I use tokens the will expire at some time. In relay I'm fetching the expired error and unset the token, so I will get a response to my query. When 10 tokens expired I will get no response from postgraphql anymore. I have a guess what the problem could be but my PhpStorm ;) is not debugging quite well with all these async functions xD

First the output of my requests:
```
TokenExpiredError: jwt expired
    at Object.module.exports [as verify] (postgraphql/node_modules/jsonwebtoken/verify.js:121:19)
    at Object.<anonymous> (postgraphql/build/postgraphql/http/setupRequestPgClientTransaction.js:42:33)
    at next (native)
    at __awaiter (postgraphql/build/postgraphql/http/setupRequestPgClientTransaction.js:7:65)
    at __awaiter (postgraphql/build/postgraphql/http/setupRequestPgClientTransaction.js:3:12)
    at Object.setupRequestPgClientTransaction [as default] (postgraphql/build/postgraphql/http/setupRequestPgClientTransaction.js:26:12)
    at Object.<anonymous> (postgraphql/build/postgraphql/http/createPostGraphQLHttpRequestHandler.js:248:69)
    at next (native)
    at fulfilled (postgraphql/build/postgraphql/http/createPostGraphQLHttpRequestHandler.js:4:58)
    at process._tickCallback (internal/process/next_tick.js:103:7)
1 error(s) in 28370.96ms :: query App { query { id ...F0 } } fragment F0 on Query { currentUser { rowId username id } id }
```

Ok now my guess:
Something with the pgPool? That the pool is not freed when an exception happened?

pencilcheck:
I just create a new endpoint in expressjs to extend expired token.

The general idea is to use https://github.com/auth0/node-jsonwebtoken to verify first, then sign and create a new one with extended expiration date

benjie:
[semi-automated message] Hi, there has been no activity in this issue for over 8 weeks so I'm closing it to keep the issues/pull request manageable. If this is still an issue, please re-open with a comment explaining why.

![](https://media.giphy.com/media/RpNA43Cnd38Bi/giphy.gif?response_id=5925dc27fd5ebd80cc843ee9)

benjie:
(Thanks for the notes @pencilcheck 👍 )

PR #315, opened 2017-01-21
fix(postgraphql): try getting error code from error.statusCode before falling back to 500

colophonemes:
The HTTP error handler was looking in `error.status` for the status code to return to the client, but some error codes were being set in `error.statusCode`.

Fixes half the error reported in #287 where the JWT error throws a 500 rather than a 403.


calebmer:
Looks good to me. Although I agree that we should try to be more consistent about this…

PR #317, opened 2017-01-24
fix(postgres): prevent processing same constraint twice

### LINKED ISSUE #316: Redefinition of same Relation by introspection data
When trying to run postgraphql against my db it exits with: 

```
Error: Relation of name 'x' already exists between head collection 'y' and tail collection 'z' in the inventory.
    at Inventory.addRelation (/home/work/postgraphql/build/interface/Inventory.js:67:19)
    at _loop_1 (/home/work/postgraphql/build/postgres/inventory/addPgCatalogToInventory.js:59:23)
    at Object.addPgCatalogToInventory (/home/work/postgraphql/build/postgres/inventory/addPgCatalogToInventory.js:67:9)
    at Object.<anonymous> (/home/work/postgraphql/build/postgraphql/schema/createPostGraphQLSchema.js:57:32)
    at step (/home/work/tslib/tslib.js:114:27)
    at Object.next (/home/work/tslib/tslib.js:95:57)
    at fulfilled (/home/work/tslib/tslib.js:85:62)
    at process._tickCallback (internal/process/next_tick.js:103:7)z
```

Not sure what exactly causes it but I am pretty sure that this is not intended, is following line really a good idea? Wouldn't it be better to check for non-equality of the relation: I.e. if the same relation already exists, no problem but if someone tries to add a relation that is syntactically different: problem?!
https://github.com/calebmer/postgraphql/blob/3e24703584bb4baaa460da589b072408020a8e1c/src/interface/Inventory.ts#L79

martinheidegger:
Right now it is possible that the introspection sql script (and by extension postgres) returns the same constraint (with different names) twice. In this case you will get the error of #316. This PR closes #316 by storing a key/hash for a constraint and checking for its existence before adding it to the constraint set _(making it a true constraint set and not a constraint list)_.

### DIFF
--- a/src/postgres/introspection/PgCatalog.ts
+++ b/src/postgres/introspection/PgCatalog.ts
@@ -20,7 +20,7 @@ class PgCatalog {
   private _procedures: Set<PgCatalogProcedure> = new Set()
 
   constructor (objects: Array<PgCatalogObject>) {
-    const foundConstraints:Set<String> = new Set()
+    const foundConstraints: Set<String> = new Set()
     // Build an in-memory index of all our objects for ease of use:
     for (const object of objects) {
       switch (object.kind) {
@@ -37,13 +37,13 @@ class PgCatalog {
           this._types.set(object.id, object)
           break
         case 'constraint':
-          var constraint: PgCatalogForeignKeyConstraint = object as PgCatalogForeignKeyConstraint
-          var constraintKey = JSON.stringify({
+          const constraint: PgCatalogForeignKeyConstraint = object as PgCatalogForeignKeyConstraint
+          const constraintKey = JSON.stringify({
             type: constraint.type,
             classId: constraint.classId,
             foreignClassId: constraint.foreignClassId,
             keyAttributeNums: constraint.keyAttributeNums,
-            foreignKeyAttributeNums: constraint.foreignKeyAttributeNums
+            foreignKeyAttributeNums: constraint.foreignKeyAttributeNums,
           })
           if (!foundConstraints.has(constraintKey)) {
             foundConstraints.add(constraintKey)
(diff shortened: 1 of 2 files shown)


calebmer:
I would prefer if we could get the fix in the introspection SQL query itself: https://github.com/calebmer/postgraphql/blob/a9b810a77565b77fb2d3a29569b8a9619e4a669a/resources/introspection-query.sql. Would this be possible? Perhaps using [`SELECT DISTINCT`](https://www.postgresql.org/docs/9.6/static/sql-select.html) like `select distinct on (con.conrelid, con.confrelid, con.conkey, con.confkey)`.

martinheidegger:
@calebmer Sadly I don't know postgres well enough about this.

calebmer:
Could you give me a simple schema where this error may occur? Both so that we can add it to tests and I can give it a shot this weekend.

benjie:
If you can give a schema that reproduces the issue I'd be happy to attempt the SQL modification so Caleb can focus his efforts elsewhere.

martinheidegger:
Let me see if I can whip up one in a few hours.z

martinheidegger:
Sorry, wasn't able to get it done before now: the following schema should result in the error, i think.

```sql
CREATE TABLE b_content_type (
    id integer NOT NULL,
    a_type_id integer,
    name character varying(100) NOT NULL,
    type character varying(100) NOT NULL
);


CREATE TABLE a_content_type (
    id integer NOT NULL,
    name character varying(100) NOT NULL,
    model character varying(100) NOT NULL
);

ALTER TABLE ONLY a_content_type
    ADD CONSTRAINT a_content_key PRIMARY KEY (id);

ALTER TABLE ONLY b_content_type
    ADD CONSTRAINT b_content_key PRIMARY KEY (id);

ALTER TABLE ONLY b_content_type
    ADD CONSTRAINT constraint_A FOREIGN KEY (a_type_id) REFERENCES b_content_type(id) DEFERRABLE INITIALLY DEFERRED;


ALTER TABLE ONLY b_content_type
    ADD CONSTRAINT constraint_B FOREIGN KEY (a_type_id) REFERENCES b_content_type(id) DEFERRABLE INITIALLY DEFERRED;
```

calebmer:
Thanks so much for your contribution and your test case @martinheidegger 👍

I’m going to close this in favor of a SQL based solution in https://github.com/calebmer/postgraphql/pull/326

martinheidegger:
Awesome!

Issue #321, opened 2017-01-26
Error: Name "__id" must not begin with "__", which is reserved by GraphQL introspection.


skhro87:
I have tried a lot of things but always get this error message when trying to connect with postgraphql to a postgres instance.

$postgraphql -c postgres://postgres:postgres@localhost:5432/postgres -s kitchen_sink

PostGraphQL server listening on port 5000 🚀

  ‣ Connected to Postgres instance postgres://localhost:5432/postgres
  ‣ Introspected Postgres schema(s) kitchen_sink
  ‣ GraphQL endpoint served at http://localhost:5000/graphql
  ‣ GraphiQL endpoint served at http://localhost:5000/graphiql

* * *

Error: Name __id must not begin with __, which is reserved by GraphQL introspection.
    at assertValidName (/usr/lib/node_modules/postgraphql/node_modules/graphql/utilities/assertValidName.js:27:11)
    at /usr/lib/node_modules/postgraphql/node_modules/graphql/type/definition.js:328:42
    at Array.forEach (native)
    at defineFieldMap (/usr/lib/node_modules/postgraphql/node_modules/graphql/type/definition.js:327:14)
    at GraphQLInterfaceType.getFields (/usr/lib/node_modules/postgraphql/node_modules/graphql/type/definition.js:401:44)
    at /usr/lib/node_modules/postgraphql/node_modules/graphql/type/schema.js:207:27
    at typeMapReducer (/usr/lib/node_modules/postgraphql/node_modules/graphql/type/schema.js:219:7)
    at Array.reduce (native)
    at typeMapReducer (/usr/lib/node_modules/postgraphql/node_modules/graphql/type/schema.js:202:39)
    at Array.reduce (native)

The installation of postgres is fresh.
I can connect using the same connection string via psql and also via other tools.
I tried with and without specifying the schema. I run the kitchen sink sql from here http://calebmer.com/postgraphql-docs/examples/kitchen-sink/

$uname -r
4.8.13-1-ARCH

$psql --version
psql (PostgreSQL) 9.6.1

$postgraphql --version
2.5.0

benjie:
This relates to the newly released graphql 0.9.0 which broke our dependencies - see https://github.com/calebmer/postgraphql/pull/319

Potential solutions: downgrade to graphql < 0.9 or try running postgraphql with `--classic-ids`

skhro87:
ahhh! thanks a lot, it works with `--classic-ids` !

benjie:
[...]

PR #322, opened 2017-01-26
fix(deps): restrict graphql version range

benjie:
See https://github.com/calebmer/postgraphql/pull/319 and #321; this is clearly breaking things for people who want to get started with postgraphql, so we should get a simple fix out ASAP before fixing it more permanently by renaming `__id` later. 

@calebmer can you release this one commit as a bugfix release?

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "dataloader": "^1.2.0",
     "debug": "^2.3.3",
     "finalhandler": "^0.5.1",
-    "graphql": ">=0.6.0 <1.0.0",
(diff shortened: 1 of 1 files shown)


calebmer:
Good call. I’m going to put this and your PR (https://github.com/calebmer/postgraphql/pull/269) into two more `2.x` releases and then do a `3.x` release with the id breaking change and a breaking change to big integer types so that they return strings and not numbers.

benjie:
Sweet 👍

Issue #224, opened 2016-11-10
Only NATURAL order when creating a procedure


charlyz:
Hey !

I have a question about the `orderBy` argument. I created a procedure to search for a value that looks like the one in your examples:

```
create function forum_example.search_posts(search text) returns setof forum_example.post as $$
  select post.*
  from forum_example.post as post
  where post.headline ilike ('%' || search || '%') or post.body ilike ('%' || search || '%')
$$ language sql stable;
```

When I use `searchPost`, the only order I have for the argument `orderBy` is `NATURAL`. How can I get all the other types of ordering like `NAME_ASC`, `NAME_DESC`? I tried to find an answer by searching the issues but I had no luck. Sorry if you already answered that question.

**EDIT:** I'm adding a second question. How can I use a condition to return all the results with a value lower or greater than a threshold? For example `condition: {clicksLast30Days < 60}`.

Cheers

calebmer:
Hey @charlyz 😊 

Adding the other orderings would be fairly easy to do, just know that with natural ordering you define whatever ordering you want inside of the procedure. Take for instance:

```sql
create function forum_example.search_posts(search text) returns setof forum_example.post as $$
  select post.*
  from forum_example.post as post
  where post.headline ilike ('%' || search || '%') or post.body ilike ('%' || search || '%')
  order by post.author_id, post.headline
$$ language sql stable;
```

Here we order by the `author_id` *and* `headline`, or any other variant. How important is ordering in GraphQL to you? Can ordering in the procedure be enough for now?

Relevant links for anyone interested in implementing this:

- Here is the code responsible for only `NATURAL` ordering in procedures: https://github.com/calebmer/postgraphql/blob/032bee46b22560e83dc5f76bd7bc5174be4d2e7a/src/postgraphql/schema/procedures/PgProcedurePaginator.ts#L30-L40
- Here is the code responsible for creating all of the column orderings for tables: https://github.com/calebmer/postgraphql/blob/032bee46b22560e83dc5f76bd7bc5174be4d2e7a/src/postgres/inventory/paginator/PgCollectionPaginator.ts#L49-L117

The implementation may be as simple as an if-branch and a copy/paste.

* * *

For your second question, there is no current way to do this in the GraphQL API 😣. Mostlty because we can’t decide on a good API design for this. See one of our oldest issues, https://github.com/calebmer/postgraphql/issues/47, for more information on the topic. If you have a good design, please share :+1:

[...]

Issue #318, opened 2017-01-24
Limit rows returned for specific users?


Kurren123:
Hi there

I have two database roles, Users and Admins. How can I limit the size of query results returned by graphQL by (e.g.) 100 rows for Users, but not limit the results for Admins?

Thank you for any suggestions,

Kurren

benjie:
Hi @Kurren123, postgraphql doesn't implement any features like this in itself; but here's a couple of ways you might achieve it:

1. put some kind of GraphQL query validation service in front of postgraphql so that it can validate which queries area allowed and which are not before sending them on to postgraphql and postgres
2. add custom functions in Postgres that look at the current user's access levels and decide on the row limit accordingly
3. specify the row limit in a JWT token and use custom functions again

This may be of interest: https://github.com/calebmer/postgraphql/blob/master/docs/advanced-queries.md

Kurren123:
Hi @benjie . Thank you for your reply. If I go with 2 or 3, how can I prevent a malicious client from selecting millions of rows?

benjie:
One way is to not allow access to query the tables directly, instead requiring the user to go through your custom mutations. Another is to use built in postgres features such as statement timeout, to bolster your defence. 

However I'd strongly encourage you to go with option 1  though there's not any good ready-to-use solutions out there for this yet, as far as I know. I recommend this post on StackOverflow as a good starting point:

http://stackoverflow.com/a/37338465/141284

It sounds like your issue could be solved by writing a validation rule (like the  [built in ones here](https://github.com/graphql/graphql-js/tree/master/src/validation/rules)) that a) requires limits on certain endpoints, and b) ensures that those limits are sensible.

Kurren123:
Thank you for your answer, much appreciated. I think we can close this issue 😄

benjie:
@Kurren123 Do let us know if you build your own or find a solution, I think this area of GraphQL is still pretty cutting edge.

PR #324, opened 2017-01-27
Custom row matchers in connections

guigrpa:
**Motivation**: It would be nice to add flexibility to connection filtering (multi-table, multi-field, non-equality comparisons, etc.) without losing the ability to paginate with cursors. Currently, if the user implements a custom row matcher in the database (e.g. `search_posts()` in the forum example), he loses the great cursor-based pagination supported on collections.

**Implementation**: When the user specifies a custom row matcher (via the new `--matcher` CLI flag, or `pgRowMatcher` library option), the `condition` field of all GraphQL connections is re-defined as having type `Json`, and all connection rows are validated (in the `where` clause) through the specified row matcher. This new *condition clause* is used in both offset- and attribute-based orderings, and for all pagination fields: `hasNextPage`, `hasPreviousPage`, `count`, etc.

An example of a polymorphic row matcher with support for nested conditions, partial matches (only equality for the time being) and diacritic removal, can be found under the `examples` folder. I believe this solution gives a great degree of freedom to the user to implement whichever filtering strategies he needs.

Regarding the multiple-field, multiple-table sorting feature I mentioned in #224, that's a quite complex problem that requires further thought. This PR, while still limited to the single-attribute orderings available in the library, is just a stepping stone in that direction.

### DIFF
--- a/src/postgres/inventory/conditionToSql.ts
+++ b/src/postgres/inventory/conditionToSql.ts
@@ -31,6 +31,10 @@ export default function conditionToSql (condition: Condition, path: Array<string
       if (!match) throw new Error('Invalid regular expression.')
       const [, pattern, flags] = match
       return sql.query`regexp_matches(${sql.identifier(...path)}, ${sql.value(pattern)}, ${sql.value(flags)})`
+    case 'CUSTOM_MATCHER':
+      // We use `sql.raw` (beware of its dangers!) since the matcher is provided by the (server-side) user
+      // Since it does not come from the GraphQL interface, it can be trusted
+      return sql.query`${sql.raw(condition.matcher)}(to_json(${sql.identifier(Symbol.for('mainSelect'))}), ${sql.value(condition.value)})`
     default:
       throw new Error(`Condition of type '${condition['type']}' is not recognized.`)
   }
--- a/resources/example-row-matcher.sql
+++ b/resources/example-row-matcher.sql
@@ -1,0 +1,93 @@
+-- An example of polymorphic row matcher. Accepts matching criteria as a json object.
+-- Example of matching criteria, taking `forum_example.person` as the starting table:
+--
+-- {
+--   "first_name": "íllérmø",  -- will also find "Guillermo"
+--   "post:author_id:rev": {    -- <related table>:<fk name>[:rev] (rev, for reverse, is included when the key is in the related table)
+--     "headline": "first post",
+--     "comment:post_id:rev": {
+--       "body": "awesome"
+--     }
+--   }
+-- }
+--
+-- When a relation is traversed, the original record is included in the foundset if ANY of the N records
+-- matches the criterion.
+--
+-- Another example, with `forum_example.post` as the starting table:
+--
+-- {
+--   "headline": "first post",
+--   "post:author_id": {
+--     "first_name": "Guillermo"
+--   }
+-- }
+
+CREATE OR REPLACE FUNCTION forum_example.row_match(in_row json, criteria jsonb) RETURNS boolean AS $$
+DECLARE
+  found int = 0;
+  attr text;
+  is_match boolean;
+  colon_pos int;
+  colon_pos2 int;
+  related_table text;
+  related_fk text;
+  related_criteria jsonb;
+  related_query text;
+  related_row record;
+  related_count int;
+BEGIN
+  RAISE NOTICE '*** %: %', in_row->>'id', in_row;
+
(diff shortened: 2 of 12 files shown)


guigrpa:
Hmmm, I see TypeScript is complaining in CI. I'm no TypeScript expert, so some help would be great…

src/postgres/inventory/conditionToSql.ts
```diff
@@ -31,6 +31,10 @@ export default function conditionToSql (condition: Condition, path: Array<string
       if (!match) throw new Error('Invalid regular expression.')
       const [, pattern, flags] = match
       return sql.query`regexp_matches(${sql.identifier(...path)}, ${sql.value(pattern)}, ${sql.value(flags)})`
+    case 'CUSTOM_MATCHER':
+      // We use `sql.raw` (beware of its dangers!) since the matcher is provided by the (server-side) user
+      // Since it does not come from the GraphQL interface, it can be trusted
+      return sql.query`${sql.raw(condition.matcher)}(to_json(${sql.identifier(Symbol.for('mainSelect'))}), ${sql.value(condition.value)})`
```
benjie: ~~I think since the matcher is a function name you can use `sql.identifier` instead of `sql.raw`?~~

I suppose the function could be scoped to a schema, e.g. `foo.filter`, in which case `sql.identifier` would not work. 👍

benjie: I don't think you should perform `to_json` here; just passing the `record` through would be more performant and easier for PG to optimise, I think? (The matcher itself can then cast it to JSON if required.)

benjie:
I have performance concerns about this; e.g. if it's the only condition applied to a table with say 100 million rows, that's a hell of a lot of plpgsql function invocations (especially with the json overhead).

I'm just spitballing here, but instead could we implement the row filtering using a javascript function that generates an SQL fragment - this can be better optimised by postgres? This would make specifying the matcher on the command line more challenging (perhaps give it a filename to import the matcher from?), but it could be passed in as a library option quite easily?

src/postgres/inventory/conditionToSql.ts
guigrpa: OK with the second remark. I will update the PR.

Regarding the first one, I was in fact testing with a schema-scoped function

guigrpa:
You're totally right with your performance concerns. Even if the user paginates, with a 100-million-row table and a too specific search condition, quite a few function calls may be needed to fill in the desired page.

So I like your idea of custom JS functions to process `condition` values in the GraphQL query. However, take into account that multi-table search criteria would require not only additional `WHERE` conditions but also `JOIN`s, etc. The implementation may start to become quite complex, don't you think? 

If @calebmer also agrees, I'd like to try out that idea, before proposing an API for such filter functions.

calebmer:
Great job on the PR, I’m glad you were able to find your ways around the PostGraphQL internals 😊

I’m not convinced this is the right solution, though when there are alternatives. Why not just generate attribute orderings for connection procedures like we do with tables? So in addition to `NATURAL` the user would get `ID_ASC`, `FIRST_NAME_DESC`, etc. When you need a custom ordering beyond those attributes, `NATURAL` should be fine. I’ve also heard a request for this feature through other channels. Would this solve your problem as well?

benjie:
@guigrpa Those joins could potentially be solved with subqueries which postgresql can often re-write during the query planner stage into joins, see https://www.postgresql.org/docs/9.4/static/runtime-config-query.html#GUC-FROM-COLLAPSE-LIMIT

With subqueries it's be possible to express quite complex relations (even with conditional joins) using this format.

guigrpa:
@benjie Thanks, I'll investigate the sub-query option. That would certainly be a better solution.

@calebmer The codebase is really awesome: well-organised, full of comments and types (which are also documentation). You've really done a good job! Regarding orderings, I'm not addressing them here; I'm only tackling multi-table filtering as a first step. I'll check out @benjie's proposal and get back to you.

benjie:
Subqueries often have less-good performance than joins (depends how postgres plans it); sometimes they're similar, and rarely they're faster. But for a user-provided function it does provide more flexibility.

Perhaps the function could return more than just a where clause? I'd love to be able to modify things about the query to be executed (e.g. Adding/modifying limits like in #318, adding joins, etc); maybe it could be fed the fragments of the query that is about to be ran and it can manipulate them as desired and then return them to be joined and queried...

benjie:
Closing this PR in the hopes that someone might implement it with a v4 plugin

https://github.com/postgraphql/postgraphql/wiki/v4-plugins

PR #329, opened 2017-01-30
Adds optional properties to update/set created and modified timestamps.

martinheidegger:
Some databases use created/modified fields to mark changes in the data rows. Usually those fields are not meant to be set during api queries but automatically be filled. This PR adds a customizable options to postgraphql that allow setting the fields following a naming pattern.

### DIFF
--- a/docs/library.md
+++ b/docs/library.md
@@ -62,6 +62,7 @@ Arguments include:
   - `watchPg`: When true, PostGraphQL will watch your database schemas and re-create the GraphQL API whenever your schema changes, notifying you as it does. This feature requires an event trigger to be added to the database by a superuser. When enabled PostGraphQL will try to add this trigger, if you did not connect as a superuser you will get a warning and the trigger won’t be added.
   - `disableQueryLog`: Turns off GraphQL query logging. By default PostGraphQL will log every GraphQL query it processes along with some other information. Set this to `true` to disable that feature.
   - `enableCors`: Enables some generous CORS settings for the GraphQL endpoint. There are some costs associated when enabling this, if at all possible try to put your API behind a reverse proxy.
+  - `timestamps`: Enables automatic setting of timestamps to fields when editing/creating new rows. `null`, the default, indicates that no fields should be set. `{modified: 'lastMod'}` would indicate that the `lastMod` field is set on modification. `{created: 'init', modified: 'lastMod'}` would additionally sets `created` to the creation timestamp once a new row is created.  
 
 [connect]: https://www.npmjs.com/connect
 [express]: https://www.npmjs.com/express
--- a/src/graphql/schema/BuildToken.ts
+++ b/src/graphql/schema/BuildToken.ts
@@ -1,6 +1,13 @@
 import { GraphQLOutputType, GraphQLFieldConfig } from 'graphql'
 import { Inventory, Type, ObjectType } from '../../interface'
 
+interface BuildTokenTimestamps {
+  // Field name to be set on modification with the current timestamp
+  modified?: string,
+  // Field name to be set on creation with the current timestamp
+  created?: string,
+}
+
 /**
  * A `BuildToken` is a plain object that gets passed around to all of the
  * functions used to create a GraphQL schema. The `BuildToken` contains a copy
@@ -30,6 +37,9 @@ interface BuildToken {
     // If true then the default mutations for tables (e.g. createMyTable) will
     // not be created
     readonly disableDefaultMutations: boolean,
+    // If timestamps are given the fields are excluded from create/modify
+    // queries and automatically filled
+    readonly timestamps?: BuildTokenTimestamps,
   },
   // Hooks for adding custom fields/types into our schema.
   readonly _hooks: _BuildTokenHooks,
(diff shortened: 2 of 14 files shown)


benjie:
An excellent pull request with well formatted code, and even tests!

I'm just wondering about the motivation behind this - what advantages does doing it in postgraphql have compared to doing it in your database directly using triggers?  I use the following:

```sql
create function hookhaven_utils.set_created_at() returns trigger as $$
begin
  new.created_at := current_timestamp;
  new.updated_at := new.created_at;
  return new;
end;
$$ language plpgsql;

create function hookhaven_utils.set_updated_at() returns trigger as $$
begin
  new.updated_at := current_timestamp;
  if new.updated_at <= old.updated_at then
    -- We must guarantee that updated_at will always be large than it was before after an edit
    new.updated_at = old.updated_at + interval '1 millisecond';
  end if;
  return new;
end;
$$ language plpgsql;
```

then when I create each table I add:

```sql
create trigger created_at before insert on table_name for each row execute procedure set_created_at();
create trigger updated_at before update on table_name for each row execute procedure set_updated_at();
```

martinheidegger:
The solution with the triggers has two *difficult* problems for me:

1.) The database might already exist and adding the triggers is way more work than to just offer the fields
2.) I am not sure if the current postgraphql hides trigger fields in `Update<X>Type`'s in other words: The user should not be able to provide an own timestamp.

Introspecting the procedures and checking them for hiding the fields seems like an insanely complex task. And simply hiding the fields in the update/create fields might result in weird error messages that are unexplainable to users...

benjie:
1. fair enough (aside: I've had success in the past using introspection queries to generate SQL which I then pipe to a migration file to solve issues like this)
2. it does not; but you can defend against this in two ways: one would be to simply not allow the postgraphql user to update those fields (with column-based grants), and another is a slight modification to `set_updated_at` that would always reset created_at to the previous value (updated_at already ignores user input):

```sql
create function set_updated_at() returns trigger as $$
begin
  new.created_at := old.created_at;
  new.updated_at := current_timestamp;
  if new.updated_at <= old.updated_at then
    -- We must guarantee that updated_at will always be large than it was before after an edit
    new.updated_at := old.updated_at + interval '1 millisecond';
  end if;
  return new;
end;
$$ language plpgsql;
```

Removing fields from PostGraphQL's schema is something we should give an easy way to do 👍

martinheidegger:
Are column based grants supported to this degree?

benjie:
Sure, e.g.:

```sql
grant update (name, visibility) on team to postgraphql_user;
```
(i.e. disallows id, created_at, updated_at)

See [GRANT](https://www.postgresql.org/docs/9.6/static/sql-grant.html):
```
GRANT { { SELECT | INSERT | UPDATE | REFERENCES } ( column_name [, ...] )
    [, ...] | ALL [ PRIVILEGES ] ( column_name [, ...] ) }
    ON [ TABLE ] table_name [, ...]
    TO role_specification [, ...] [ WITH GRANT OPTION ]
```

calebmer:
Well done PR, but I agree with @benjie about alternatives.

There is a lot we could do in PostGraphQL to support common cases like this, but I don’t think we should. One of the main advantages of PostGraphQL is that you can completely move your business logic into the database and PostGraphQL acts merely as a view on top of that logic. This has the advantage from a user’s perspective that they are not locked into PostGraphQL. PostGraphQL is simply one, efficient, interaction point to their data of which there could be many.

To keep ourselves aligned with this goal I think there are two alternatives which may have a better payoff in the long term. Both alternatives align with the pain points you mentioned @martinheidegger 😊

1. Create a framework for building powerful Postgres databases quickly. This framework could sit on top of a database that had already been built, or allow a user to rapidly build out a brand new database. It would contain features like retroactively adding `created_at` and `updated_at` fields to all your tables (this would actually make for a pretty cool UI and SaaS product now that I think about it…).
2. Add better support to PostGraphQL for column based grants. Of course, if you try updating anything that you don’t have access to through a Postgres grant, you will get an error. But Postgres currently does not have any custom logic to *hide* columns you are not allowed to read or update. Adding this feature could provide the same benefits, but for a broader application.

@martinheidegger tell us what you think about these ideas 😊

martinheidegger:
> One of the main advantages of PostGraphQL is that you can completely move your business logic into the database and PostGraphQL acts merely as a view on top of that logic

Indeed it is! I have deep respect for your restraint for adding features to the core.
I have imagined postgraphql as a rapid graphql tool that allows to quickly setup a high-quality graphql adapter. And as such the `--created` and `--modified` fields seemed like a worthy addition because it would make getting started easier and reduce the general friction that possibly many people have when touching it first.

To answer to your plans:

1. Having this sort of framework / tool would certainly be useful. _Not sure how a SaaS / product would look like._ However it seems like it would take quite some time to get such a complex _?_ tool done, right? Also: Would this tool make the work easier? Setting a config seems like the easiest possible solution for this particular problem.
2. This would be highly useful but I am not quite sure if it would completely work. Meaning that at some point probably some sort of workaround or configuration might become necessary?!

This PR might not be the perfect solution but it works right now for my use case and helps me get closer to use postgraphql without a lot of changing of the sql schema. As such others might want to get their hands on this as well until another solution is available.

benjie:
Hi @martinheidegger, thanks for this high quality pull request, but sadly I don't think it is a good fit for PostGraphQL. Since the maintainers are very busy I'd like to maintain as small a core feature-set as we can get away with, and in my opinion this problem is better served by triggers or database logic outside of PostGraphQL itself. Sorry!

Kind regards,

Benjie.

PR #331, opened 2017-01-30
Better postgres Connection string support

martinheidegger:
This PR fixes two tiny issues I have had with the connection string option:

1) it gives a little demo in the help output
2) it shows the default port of postgres in case the connection string doesn't support one


martinheidegger:
🎉

PR #333, opened 2017-01-30
Added primaryKeyAPI option

martinheidegger:
Currently postgraphql will create two endpoints for e every collection. `collectionName(__id:ID!)` and `collectionNameById(id: int)`. This PR will hide the second endpoint in favor for the first endpoint unless the `primaryKey` option is set.

### DIFF
--- a/docs/library.md
+++ b/docs/library.md
@@ -62,7 +62,7 @@ Arguments include:
   - `watchPg`: When true, PostGraphQL will watch your database schemas and re-create the GraphQL API whenever your schema changes, notifying you as it does. This feature requires an event trigger to be added to the database by a superuser. When enabled PostGraphQL will try to add this trigger, if you did not connect as a superuser you will get a warning and the trigger won’t be added.
   - `disableQueryLog`: Turns off GraphQL query logging. By default PostGraphQL will log every GraphQL query it processes along with some other information. Set this to `true` to disable that feature.
   - `enableCors`: Enables some generous CORS settings for the GraphQL endpoint. There are some costs associated when enabling this, if at all possible try to put your API behind a reverse proxy.
-  - `primaryKeyAPI`: Unless multiple primary keys are defined PostGraphQL will only show the default endpoint for a type. By enableing `primaryKeyAPI` it will also add endpoints and fields for the default keys. 
+  - `primaryKeyAPI`: Unless multiple primary keys are defined PostGraphQL will only show the default endpoint for a type. By enabling `primaryKeyAPI` it will also add endpoints and fields for the default keys. 
 
 [connect]: https://www.npmjs.com/connect
 [express]: https://www.npmjs.com/express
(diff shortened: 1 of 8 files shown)


docs/library.md
```diff
@@ -62,6 +62,7 @@ Arguments include:
   - `watchPg`: When true, PostGraphQL will watch your database schemas and re-create the GraphQL API whenever your schema changes, notifying you as it does. This feature requires an event trigger to be added to the database by a superuser. When enabled PostGraphQL will try to add this trigger, if you did not connect as a superuser you will get a warning and the trigger won’t be added.
   - `disableQueryLog`: Turns off GraphQL query logging. By default PostGraphQL will log every GraphQL query it processes along with some other information. Set this to `true` to disable that feature.
   - `enableCors`: Enables some generous CORS settings for the GraphQL endpoint. There are some costs associated when enabling this, if at all possible try to put your API behind a reverse proxy.
+  - `primaryKeyAPI`: Unless multiple primary keys are defined PostGraphQL will only show the default endpoint for a type. By enableing `primaryKeyAPI` it will also add endpoints and fields for the default keys.
```
benjie: `enableing` typo

src/graphql/schema/collection/createCollectionGqlType.ts
```diff
@@ -20,6 +20,10 @@ export default function createCollectionGqlType<TValue> (
   const { options, inventory } = buildToken
   const { type, primaryKey } = collection
   const collectionTypeName = formatName.type(type.name)
+  let fields = Array.from(type.fields.entries())
+  if (primaryKey && !buildToken.options.primaryKeyAPI) {
+    fields = fields.filter(([ fieldName ]) => fieldName !== primaryKey.name)
+  }
```
benjie: For backwards compatibility, `primaryKeyAPI` should default true.

src/postgraphql/cli.ts
```diff
@@ -37,6 +37,7 @@ program
   .option('-a, --classic-ids', 'use classic global id field name. required to support Relay 1')
   .option('-j, --dynamic-json', 'enable dynamic JSON in GraphQL inputs and outputs. uses stringified JSON by default')
   .option('-M, --disable-default-mutations', 'disable default mutations, mutation will only be possible through Postgres functions')
+  .option('-y, --primary-key-api', 'enable additional endpoints to primary-key apis')
```
benjie: Since this should probably default true, adding a command line option to disable it instead would be wise.

benjie:
Would this cause issues with familiar operations? E.g. if you have SQL functions they cannot receive and understand a postgraphql ID (well, they can - it's just base64 encoded stuff - but it's a faff to extract that in each function and validate the collection matches what's expected) yet they need to know the primary key of the thing they're manipulating. Does it handle mutations so that you can still, for example, post a new comment for a blog post by passing in the `__id`/`nodeId` rather than the current ID which accepts primary key.

Removing `id` just obfuscates the primary key at best, base64 decoding `__id`/`nodeId` would return the primary key (among other things). Is the intention of this PR to simply tidy up your schema a little?

martinheidegger:
Thinking about this PR a second time: it might indeed be in an early stage. The problem is that the existence of both the primary key and the ID is confusing - at best - and because of that I started working on this PR. It is also the reason why I set this new "hiding" as  "default"  _(and by extension: a breaking change)_ rather than optional since first-time users would see this as less weird. I guess.

calebmer:
We should definitely discuss a strategy for disabling some of the extraneous fields in APIs, but I’d like to see a cohesive plan instead of adopting one-off command line flags for everything the user wants to disable. For example, I’d also like to consider adding the inverse where a user could disable the `Node` interface and globally unique ids entirely. We could end up with a lot of flags 😣

Also, is this an issue we could fix with better documentation? I definitely think you could be onto something when you say that having two ways to select a row can be confusing to new users. How could we provide documentation to alleviate this?

EyMaddis:
@calebmer https://github.com/calebmer/postgraphql/pull/403#issuecomment-288678726 this might be a way to solve the flag problem, by directly mapping the cli parameters to the JS options passed in the library mode

benjie:

PR #334, opened 2017-01-30
Allowing graphiql to work both during development and when published

### LINKED ISSUE #330: graphiQL broken?
The html output of `/graphiql` Does not output any JavaScript code and shows an empty page:

https://github.com/calebmer/postgraphql/blob/939a91ed20fc183bcd0fc8367a7b6af891e6a2de/src/postgraphql/graphiql/public/index.html

martinheidegger:
It seems like the recent graphiql improvements don't work well when used in development mode since [the build script moves the build folder](https://github.com/calebmer/postgraphql/blob/939a91ed20fc183bcd0fc8367a7b6af891e6a2de/scripts/build#L20) out of the range of [`createPostGraphQLHttpRequestHandler`](https://github.com/calebmer/postgraphql/blob/939a91ed20fc183bcd0fc8367a7b6af891e6a2de/src/postgraphql/http/createPostGraphQLHttpRequestHandler.d.ts) which is no problem when the TypeScript code is built but when [running with scripts/dev](https://github.com/calebmer/postgraphql/blob/939a91ed20fc183bcd0fc8367a7b6af891e6a2de/scripts/dev#L18) it will render [`src/postgraphql/graphiql/public/index.html`](https://github.com/calebmer/postgraphql/blob/939a91ed20fc183bcd0fc8367a7b6af891e6a2de/src/postgraphql/graphiql/public/index.html) which is rather empty. This PR changes the `mv` to a `cp` and changes the reference of the root folder to work in development mode as well.
(Closes #330)

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -17,4 +17,4 @@ echo "Building GraphiQL"
 
 echo "Moving GraphiQL build output"
 mkdir -p build/postgraphql/graphiql/public
-mv src/postgraphql/graphiql/build/* build/postgraphql/graphiql/public
+cp -Rv src/postgraphql/graphiql/build/* build/postgraphql/graphiql/public
--- a/scripts/lint
+++ b/scripts/lint
@@ -4,4 +4,4 @@ npm_bin=$(npm bin)
 
 set -e
 
-$npm_bin/tslint $(find src -name "*.[tj]s" -a \! -wholename '*/node_modules/*') $@
+$npm_bin/tslint **/*.ts **/*.js $@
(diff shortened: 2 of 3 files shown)


calebmer:
Because PostGraphQL runs GraphiQL in development with `create-react-app` I don’t think this is necessary? How are you trying to run GraphiQL in development? The correct way to run it is with `cd src/postgraphql/graphiql && PORT=5783 $npm_bin/react-scripts start` or `./scripts/dev` which will open GraphiQL from `create-react-app` automatically.

martinheidegger:
`nodemon` didn't work on my computer and additionally I am developing against my target database (not the sample db). Which is why I `./scripts/dev` doesn't work for me. I just use the `ts-node src/postgraphql.ts` straight.

benjie:
You can run `./scripts/dev -c postgres://... --schema my_schema`; it will override the default options.

Could you expand on why `nodemon` didn't work? An error stack would be helpful.

martinheidegger:
It "didn't" work, an update and it works now ?! Seems like I can live without this PR. Sorry for taking your time.

benjie:
No problem at all, I recognise that developer onboarding is not great and have been taking some steps to improve it when I've had time. We should add more detail about `./scripts/dev` to the developer documentation

PR #337, opened 2017-01-31
Custom connection filters

guigrpa:
This PR develops [this idea](https://github.com/calebmer/postgraphql/pull/324#issuecomment-275845059) and allows the user to provide a custom filtering function for a connection. If such a function is specified, the GraphQL connection's argument `condition` becomes a string and is passed directly to the callback; in return, the callback is allowed not only to add conditions to the `where` clause of the SQL query, but also to completely replace the `from` clause and even to add a `group by` clause, handy in some scenarios. This gives the user great flexibility and allows custom filters, for example, to:

* Filter connection rows by criteria encompassing multiple tables (starting at the connection's base table) — see complete example at resources/exampleConnectionFilter.js
* Implement inequality comparisons (the user is free to choose whatever DSL he prefers for the `condition` argument), and any other kind of criteria (in the example: partial, case-insensitive, diacritic-less comparisons)
* Integrate seamlessly with `postgraphql`'s nice cursor- and offset-based pagination schemes, as well as its predefined single-table orderings
* Use the most suitable types of `join`s (inner, outer left, outer right, outer full, etc.), and hence have some degree of freedom wrt. performance optimisation (benefit wrt. #324)

The user can specify such a function:

* enabling the new CLI `--connection-filter` flag (`-C`), indicating the path of the module to be loaded; or
* passing it directly in the new library option `gqlConnectionFilter`

The function should have this **signature** (see `interface/collection/ConnectionFilter.ts`):

```ts
type ConnectionFilter = (
  rawCondition: string,  // raw GraphQL connection argument `condition`
  sql: mixed,  // a set of SQL generation helper functions
  context: {
    initialTable: string,  // the name of the connection's base table
    initialSchema: string,  // the schema for the connection's base table
    convertRowIdToId: boolean,
  }
) => {
  conditionSql: sql.Sql,  // part of the `where` clause
  fromSql: sql.Sql,  // `from` clause; leave `undefined` to use the default
  groupBySql: sql.Sql,  // `group by` clause; leave `undefined` to use the default (empty)
}
```

The PR currently includes no additional tests, because I wanted to confirm that this approach seems reasonable before pursuing it further.

Finally, some **examples of GraphQL queries** that are possible **with the example filtering function** (but again, the idea is that the user can implement whatever he wants):

```
// Find people whose first name includes "guille" and with at least
// one post whose title includes the word "first"; for those people, 
// get their full name, as well as all of their posts
{
  allPeople(condition: "{\"first_name\": \"guille\", \"post:author_id:rev\": {\"headline\": \"first\"}}") {
    edges { node {
      fullName
      postsByAuthorId { edges { node { headline } } }
    } }
  }
}

// The filter is also applied to nested connections:
{
  allPeople {
    edges { node {
      fullName
      postsByAuthorId(condition: "{\"comment:post_id:rev\": {\"body\": \"gibberish\"}}") {
        edges { node { headline } }
      }
    }}
  }
}
```

### DIFF
--- a/resources/exampleConnectionFilter.js
+++ b/resources/exampleConnectionFilter.js
@@ -13,7 +13,7 @@ module.exports = (rawCondition, sql, context) => {
   const condition = JSON.parse(rawCondition)  // may throw
   const where = []
   const from = [{ table: initialTable, schema: initialSchema }]
-  if (!condition) return { from, where }
+  if (!condition) return { conditionSql: sql.query`true` }
   Object.keys(condition).forEach(attr => {
     const wherePartial = []
     processCondition(sql, context, initialTable, attr, condition[attr], from, wherePartial)
--- a/resources/exampleConnectionFilter.js
+++ b/resources/exampleConnectionFilter.js
@@ -42,6 +42,8 @@ const processCondition = (sql, context, curTable, attr, val, from, where) => {
 
   // Process attribute in a related table, with the FK in the current table (N-->1)
   // e.g. `person:author_id` (from post)
+  // ... or in the related table (1-->N)
+  // e.g. `post:author_id:rev` (from post)
   const [relatedTable, fk] = tokens
   if (from.find(({ table }) => table === relatedTable) == null) {
     from.push({ table: relatedTable, schema: context.initialSchema })
--- a/resources/exampleConnectionFilter.js
+++ b/resources/exampleConnectionFilter.js
@@ -1,0 +1,74 @@
+// Inputs:
+// - rawCondition: the string provided in the GraphQL `condition` field
+// - sql: a set of SQL utilities to be used for output generation
+// - context:
+//   - initialTable: the name of the table where the filter should start
+//   - initialSchema: schema for the `initialTable`
+// Output:
+// - conditionSql: SQL to be added to the `where` clause
+// - fromSql (optional): SQL to replace the default `from` clause
+// - groupBySql (optional): SQL to replace the default `group by` clause
+module.exports = (rawCondition, sql, context) => {
+  const { initialTable, initialSchema } = context
+  const condition = JSON.parse(rawCondition)  // may throw
+  const where = []
+  const from = [{ table: initialTable, schema: initialSchema }]
+  if (!condition) return { from, where }
+  Object.keys(condition).forEach(attr => {
+    const wherePartial = []
+    processCondition(sql, context, initialTable, attr, condition[attr], from, wherePartial)
+    where.push(sql.join(wherePartial, ' and '))
+  })
+  const conditionSql = where && where.length ? sql.query`(${sql.join(where, ' and ')})` : sql.query`true`
+  const out = { conditionSql }
+  if (from.length > 1) {
+    out.fromSql = sql.join(from.map(({ table, schema }) =>
+      sql.query`${sql.identifier(schema, table)} as ${sql.identifier(Symbol.for(table))}`
+    ), ', ')
+    out.groupBySql = sql.query`group by ${sql.identifier(Symbol.for(initialTable), 'id')}`
+  }
+  return out
+}
+
+// Recursive
+const processCondition = (sql, context, curTable, attr, val, from, where) => {
+  const tokens = attr.split(':')
+
+  // Process attribute in the current table, e.g. `first_name`
+  if (tokens.length === 1) {
+    where.push(sqlAttrComparison(sql, curTable, attr, val))
+    return
+  }
+
+  // Process attribute in a related table, with the FK in the current table (N-->1)
+  // e.g. `person:author_id` (from post)
+  const [relatedTable, fk] = tokens
+  if (from.find(({ table }) => table === relatedTable) == null) {
+    from.push({ table: relatedTable, schema: context.initialSchema })
+  }
+  if (tokens.length === 2) {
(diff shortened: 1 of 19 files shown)


calebmer:
I’m still not sure why this is better than using a Postgres procedure except perhaps keeping the attribute orderings.

You can do complex join based procedures with something like:

```sql
create function my_complex_function(my_int int) returns my_table as $$
  -- Insert complicated SQL query.
$$ language sql;
```

If at all possible, I would like to enable this to be done through SQL. If it turns out that we can’t I’d like to explore allowing a user to define custom procedures in JavaScript vs. intercepting generated procedures and inserting new logic.

Let me know what you think 😊

guigrpa:
I explored user PG procedures at the beginning (before writing #324 and this PR), but they have several drawbacks:

* (Currently) no support for attribute orderings and cursor-based pagination. This wouldn't be a big deal, since (as you suggested [here](https://github.com/guigrpa/test-postgraphql/blob/d90aa278e9cf818a4b132957cd84f24cbfeb1880/schema.sql#L69-L151)) attribute orderings could be added to procedures.

* They would only be available at the GraphQL query level, but they can't be used to filter a connection from a nested node (a benefit from both #324 and this PR).

* They may bring performance problems on large tables: the Postgres optimizer may find it harder to optimize the subquery returned by the user's procedure jointly with postgraphql's pagination constraints. Alternatively (as hinted [here](https://github.com/calebmer/postgraphql/issues/224#issuecomment-275143576)), the user's procedure may receive pagination (and sorting) attributes and do the pagination itself (even returning cursors), but that would make them much more complex and replicate functionality already present in the library.

That's why I think @bernie's [suggestion](https://github.com/calebmer/postgraphql/issues/224#issuecomment-275143576) to allow custom connection filters (developed here) seemed a good idea.

calebmer:
> They would only be available at the *root* GraphQL query level, but they can't be used to filter a connection from a nested node (a benefit from both #324 and this PR).

You can have connection procedures at the object level if you prefix the name with the table’s name and use that type as the first argument. Such as:

```sql
create function person_latest_posts(person person) returns setof post as $$
  ...
$$ language sql;
```

* * *

I can see how your performance concern may be an issue, though. However, I don’t think that the right solution is to provide hooks into core functionality. Instead, I’d like to explore how we can expose PostGraphQL internals in such a way that users may write their own procedure functions in JavaScript. This was always a direction I wanted to take PostGraphQL, and this seems like a good place to start moving to that point.

Ideally, I’d like a user to be able to extend `PgPaginator` or `PgPaginatorOrderingAttributes` in your own code and add whatever extra logic that you please. This would provide users with the most power and it would mean that we don’t need to continually add one-off hooks.

guigrpa:
**Regarding procedures in nested connections:** yes, but if you want to use a procedure for the `people` entity, no matter which *tail* collection you come from, you will need one procedure per base collection. If you generalize to all *head* collections (which is what I'm trying to do), you will need N^2 procedures.

<hr>

**Regarding extending `PgPaginator` or `PgPaginatorOrderingAttributes`**, do you mean that the user would be able to pass in a subclass of these superclasses as an option in the library API? If so:

* What would be the approximate interface? (`getFromEntrySql`, `getConditionSql` might not be enough; I proposed a more generic `getQuerySqlFragments` in my PR)

* How would you allow the GraphQL schema to include additional arguments for the user's paginator? I proposed modifying the `condition` field to make it an arbitrary `string` instead of an entity-dependent object, but maybe you have something different in mind.

* In the end, isn't passing in a subclass another form of hook? Or did you mean something else entirely?

I look forward to more details on how you would envisage such an extension point. Let me know if I can contribute to push it forward.

guigrpa:
Sorry, I meant N*(N-1) procedures :)

calebmer:
What I meant was passing an entirely new `Paginator` instance to PostGraphQL. For example, an option that would say: ”add a GraphQL field with this paginator I just provided to the `Person`, `Post`, and `Comment`” type. So I’m not really thinking about a hook, but more like an extension mechanism 😊

Making `condition` an arbitrary string would be a breaking change and we would be losing the type-safety that GraphQL provides for clients in statically typed programming languages.

guigrpa:
What would be your envisaged interface for this `Paginator`? If we clarify this point, maybe I can put together a proposal (please let me know if you're interested). I would not touch the `condition` field, but rather allow custom Paginators to specify additional arguments to be included in the GraphQL schema if they want…

calebmer:
Here is what [`Paginator`](https://github.com/calebmer/postgraphql/blob/54879f694b3918d256fedf84071703ea4dd5ec10/src/interface/Paginator.ts) looks like now, but there are a lot of changes that we could probably make. We would also need to introduce a way to add arguments. I had some designs for this at one point when adding procedures, but I don’t think they are saved anywhere.

Here is an interface that I started working on which is probably what we should work too: [`Paginator` in the `__ideas__` folder](https://github.com/calebmer/postgraphql/blob/54879f694b3918d256fedf84071703ea4dd5ec10/src/interface/__ideas__/00-new-paginator.md).

guigrpa:
I'll have a look at it as soon as I can; right now it's getting a bit hectic at this side!

benjie:
I think this thread needs to be moved to an issue really so we can close the PR.

benjie:

PR #338, opened 2017-02-01
Custom Naming Support

martinheidegger:
_Disclaimer: I have been working on this but I am not sure if it is ready for action yet, sharing in advance to test the waters_

The names for fields and types are different in various graphql databases like reindex or scaphold. This PR extends `formatNames` to have fields for all the usages in order to be able to replace them to be compatible in order for using postgraphql when moving from/to one of those databases.

### DIFF
--- a/src/graphql/utils/formats/default.ts
+++ b/src/graphql/utils/formats/default.ts
@@ -1,0 +1,48 @@
+import { camelCase, pascalCase, constantCase } from 'change-case'
+import FormatName from '../../schema/FormatName'
+
+const formatInsideUnderscores = (formatter: (name: string) => string) => (fullName: string) => {
+  const [, start, name, finish] = /^(_*)(.*?)(_*)$/.exec(fullName)!
+  return `${start}${formatter(name)}${finish}`
+}
+
+const camelCaseInsideUnderscores = formatInsideUnderscores(name => camelCase(name, undefined, true))
+const pascalCaseInsideUnderscores = formatInsideUnderscores(name => pascalCase(name, undefined, true))
+const constantCaseInsideUnderscores = formatInsideUnderscores(constantCase)
+
+const formatName: FormatName = {
+  type: pascalCaseInsideUnderscores,
+  field: camelCaseInsideUnderscores,
+  arg: camelCaseInsideUnderscores,
+  enumValue: constantCaseInsideUnderscores,
+
+  queryAllMethod: (collectionType: string): string => camelCaseInsideUnderscores(`all_${collectionType}`),
+  queryAllOrderByType: (collectionType: string): string =>  pascalCaseInsideUnderscores(`${collectionType}_order_by`),
+  queryAllEdgeType: (collectionType: string): string =>  pascalCaseInsideUnderscores(`${collectionType}_edge`),
+  queryAllRelationType: (collectionType: string): string =>  pascalCaseInsideUnderscores(`${collectionType}_connection`),
+  queryAllConditionType: (collectionType: string): string =>  pascalCaseInsideUnderscores(`${collectionType}_condition`),
+  queryAllEdgeFieldName: (collectionType: string): string =>  camelCaseInsideUnderscores(`${collectionType}_edge`),
+
+  queryMethod: camelCaseInsideUnderscores,
+  queryByKeyMethod: (collectionType: string, key: string): string => camelCaseInsideUnderscores(`${collectionType}_by_${key}`),
+  queryRelationField: camelCaseInsideUnderscores,
+  queryRelationFieldByKeyMethod: (collectionType: string, key: string): string => camelCaseInsideUnderscores(`${collectionType}_by_${key}`),
+
+  deleteMethod: (collectionType: string): string => camelCaseInsideUnderscores(`delete_${collectionType}`),
+  deleteType: (collectionType: string): string => pascalCaseInsideUnderscores(`delete_${collectionType}`),
+  deleteByKeyMethod: (collectionType: string, key: string): string =>  camelCase(`delete_${collectionType}_by_${key}`),
+  deletedID: (collectionType: string): string =>  camelCaseInsideUnderscores(`deleted_${collectionType}_id`),
+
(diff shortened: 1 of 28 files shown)


benjie:
(It would also make it a lot easier to review the code if it more tightly matched the existing code)

martinheidegger:
@benjie Thank you for the thorough reviewing, though I thought I mentioned that might be early ^_^.

To address your main issue:

```
api(...args:Array<Symbol | String>)
```
has a `...` api which is not so very good for JavaScript which is why i tend to avoid it. Also: The default case (current API) would be easy to implement but other orderings might be quite hard. The current solution has a lot of properties - yes - but those are also self explanatory.

src/graphql/utils/formats/default.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  createType: (collectionType: string): string =>  pascalCaseInsideUnderscores(`${collectionType}_input`),
+  createMethod: (collectionType: string): string => camelCaseInsideUnderscores(`create_${collectionType}`),
+
+  updateMethod: (fieldName: string): string => camelCaseInsideUnderscores(`update_${fieldName}`),
+  updateType: (collectionType: string): string =>  pascalCaseInsideUnderscores(`update_${collectionType}`),
+  updateByKeyMethod: (collectionType: string, key: string): string =>  pascalCaseInsideUnderscores(`update_${collectionType}_by_${key}`),
+  updatePatchType: (collectionType: string): string => pascalCaseInsideUnderscores(`${collectionType}-patch`),
+  updatePatchField: (fieldName: string): string => camelCaseInsideUnderscores(`${fieldName}_patch`),
+}
```
martinheidegger: Yes, totally should be equal/same. Lost the overview at some points.

src/graphql/utils/formats/default.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  mutationPayload: (collectionType: string): string =>  pascalCaseInsideUnderscores(`${collectionType}_payload`),
+
+  createType: (collectionType: string): string =>  pascalCaseInsideUnderscores(`${collectionType}_input`),
+  createMethod: (collectionType: string): string => camelCaseInsideUnderscores(`create_${collectionType}`),
+
+  updateMethod: (fieldName: string): string => camelCaseInsideUnderscores(`update_${fieldName}`),
+  updateType: (collectionType: string): string =>  pascalCaseInsideUnderscores(`update_${collectionType}`),
+  updateByKeyMethod: (collectionType: string, key: string): string =>  pascalCaseInsideUnderscores(`update_${collectionType}_by_${key}`),
+  updatePatchType: (collectionType: string): string => pascalCaseInsideUnderscores(`${collectionType}-patch`),
```
martinheidegger: Yes

src/graphql/utils/formats/default.ts
```diff
@@ -0,0 +1,48 @@
+import { camelCase, pascalCase, constantCase } from 'change-case'
+import FormatName from '../../schema/FormatName'
+
+const formatInsideUnderscores = (formatter: (name: string) => string) => (fullName: string) => {
+  const [, start, name, finish] = /^(_*)(.*?)(_*)$/.exec(fullName)!
+  return `${start}${formatter(name)}${finish}`
+}
```
martinheidegger: This is just the current logic: https://github.com/calebmer/postgraphql/blob/7eea049eea6703909dcb99fd7771f4032d6f4d09/src/graphql/utils/formatName.ts

calebmer:
This is a cool idea. I’m excited to see what it turns into 👍

I personally like the current approach vs. a symbol-based approach because an arbitrary function may make it easier to internationalize.

It is a lot more typing, but I think that may be fine for the increased flexibility and for the ability to see everything in one place. We could always also do a symbol-based approach on top of these functions. What do you think?

@martinheidegger make sure to add a schema integration test in [`postgraphqlIntegrationSchema-test.js`](https://github.com/calebmer/postgraphql/blob/master/src/postgraphql/__tests__/postgraphqlIntegrationSchema-test.js). I think that should be the only test you really need to add for this change 😊

If we could also decide on a few naming formats that we want, say: `camel`, `snake`, and `database` (which would keep the database formatting which would be helpful for @benjie) it would be great to add a CLI argument to allow users to configure that. In the JS API, it may also be advantageous to allow users to provide their own `FormatNames` object.

I’ll do a code review soon 👍

benjie:
If we allow users to provide their own formatname object then any additional formats we add at a later date would have to be considered a breaking change, which is what concerns me. Though I suppose we could just throw if their client doesn't have the required methods so they can immediately roll back or implement thr missing function.

The type safety is more assured by the currently proposed API, so it definitely has advantages. It also prevents typos that could result in additional formats accidentally being defined as currently.

Okay, I'm sold 👍

calebmer:
> If we allow users to provide their own `formatName` object then any additional formats we add at a later date would have to be considered a breaking change, which is what concerns me.

Interesting, could we allow users to supply a partial `FormatName` object with some keys missing? Then we could fall back to camel case or the default case the user prefers when a key is missing.

WoLfulus:
I'd love to be able to strip part of the name too. Some "filter" before the name gets processed by the engine...

For example, we do have hundreds of tables with prefixes (xxx_, yyy_, zzz_) and on PostGraphQL those prefixes appears on the QL queries.

For example: we need to write `allXxxUsers` which is bad.

benjie:
How about we do this with tagged template literals? e.g.

```js
formatName.type`${paginator.name}-order-by`
formatName.field`${type.name}-by-${collectionKey.name}`
formatName.field`delete-${collection.type.name}-by-${collectionKey.name}`
```

martinheidegger:
It is unlikely that I will continue to work on this PR.

benjie:
Custom naming support is now possible by extending (or replacing) the inflector in v4. We still need to tidy it up a bit and make it easier to extend (e.g. where you only pass in certain things, like the case-changing functions) but it's a fair start. If anyone wants to work on making it easier to use/easier to extend, that would be welcome.

PR #339, opened 2017-02-01
fix(postgres): fix processing money type

benjie:
Finally got around to figuring out what was wrong with the tests for me...

I'm concerned that the removal of commas might be localization dependent too, for example in continental Europe they tend to place commas where we place dots and vice versa; e.g. €1.234,56 - if this is the case for the money type then things are going to go wrong... 😜

### DIFF
--- a/src/postgres/inventory/type/scalar/pgFloatType.ts
+++ b/src/postgres/inventory/type/scalar/pgFloatType.ts
@@ -2,6 +2,11 @@ import { AdapterType, floatType } from '../../../../interface'
 import { sql } from '../../../utils'
 import PgType from '../PgType'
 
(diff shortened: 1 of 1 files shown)


calebmer:
Do you think it would be worth using something like https://github.com/openexchangerates/accounting.js for parsing monetary formats? Because you’re right about the localization thing 😉

benjie:
Seems overkill; could we just cast the column to float or decimal when selecting?

benjie:
@calebmer Should we merge this as-is and revisit accounting.js at a later date? I'm just aware that currently the money type doesn't work outside the US for anyone...

calebmer:
Yep 👍

Just waiting for CI to pass now…

Issue #265, opened 2016-12-06
Performance of computed columns in connections


ferdinandsalis:
I am lacking the proper terminology to express the problem correctly but here you go.

For every row in a table the computed column procedure is executed, hence every other row contributes significantly to the execution time of the query. I am just guessing here but could we maybe use `with` queries to 'inline' the queries 🙈

You can find a 'test case' here https://github.com/ferdinandsalis/computed-columns-perf

@calebmer please let me know if you were expecting something else for the issue and the test case.

benjie:
For the query:

```
{
  allPosts {
    edges {
      node {
        id
        body
        summary #computed column
      }
    }
  }
}
```

we could potentially solve it with a single SQL statement along the lines of:

```sql
select
  to_json(__local_0__) as value,
  json_build_object(
    'summary',
    to_json("perf"."post_summary"(__local_0__))
  ) as computed_value
from "perf"."post" as __local_0__
where "id" is not null
order by "id" using <
```

and then combine value/computed value with Object.assign or similar. You could go further with explicit column lists, minimising encoding data that isn't used, e.g. something like:

```sql
select
  json_build_object(
    'id',
    to_json(__local_0__."id"),
    'body',
    to_json(__local_0__."body"),
    'summary',
    to_json("perf"."post_summary"(__local_0__))
  ) as computed_value
from "perf"."post" as __local_0__
where "id" is not null
order by "id" using <
```

(I'm not sure if the `to_json`s inside the `json_build_object`s are actually required, I suspect not.)

benjie:
Also, if we were to do this we'd also be able to sort by computed columns as requested in #262 by @valorize:

```sql
select
  json_build_object(
    'id',
    to_json(__local_0__."id"),
    'body',
    to_json(__local_0__."body"),
    'summary',
    to_json("perf"."post_summary"(__local_0__))
  ) as computed_value
from "perf"."post" as __local_0__
where "id" is not null
order by "perf"."post_summary"(__local_0__) using <
```

(this would only work for computed columns that take no additional arguments, as the ones that take arguments cannot be expressed as static strings in our `orderBy` API currently)

benjie:
So I spent an hour digging through how PostGraphQL works, thought it might be useful to share some of my scattered learnings. I'm a GraphQL newbie, so apologies if some of this is obvious.

[...]

PR #340, opened 2017-02-02
WIP: improving performance by selecting only required fields and placing function calls into initial SQL query

benjie:
Re: #265

I've managed to process the 4th argument to `resolve` (i.e. `resolveInfo`) using a pattern heavily inspired by [join-monster](https://github.com/stems/join-monster/blob/master/src/queryASTToSqlAST.js) and doing so I've managed to make it so we don't select fields that aren't needed and we do the function calls inline. It's mostly working; ~~however :warning:because I had to disable the memoization of the DataLoader we're now running the same batchable SQL queries over and over:warning: (and thus there's no immediate performance improvement)~~

~~I need help squeezing `resolveInfo` into the DataLoader (I'm thinking we could pass them all together, extract the various fields for each request, merge them all together dropping duplicates) and also this breaks a tonne of assumptions which I need help solving (e.g. I'm currently returning the database columns using their fieldNames (camelCase) rather than their database_column_names (snake_case) which is breaking things like functions which return setof and some relations.~~

Loads of stuff will need adapting to work well with this.

It also destroys the tests.

I'm certain the performance improvements once we have it smoothed off will be worth it though!

# Before

```
  postgraphql:request GraphQL query request has begun. +7s
  postgraphql:request GraphQL query is parsed. +7ms
  postgraphql:request GraphQL query is validated. +39ms
  postgraphql:postgres begin +0ms
  postgraphql:postgres select set_config($1, $2, true) +1ms
  postgraphql:postgres select row_to_json(__local_0__) as object from "hookhaven_data"."sketch" as __local_0__ where "id" = any($1) limit $2 +7ms
  postgraphql:postgres select to_json("hookhaven_data"."sketch_is_favorite"((($1 + 0), ($2 + 0), ($3 + 0), null, $4, $5, $6, $7, $8, ($9 + 0), null, null)::"hookhaven_data"."sketch")) as value +5ms
  postgraphql:postgres select to_json("hookhaven_data"."sketch_settings"((($1 + 0), ($2 + 0), ($3 + 0), null, $4, $5, $6, $7, $8, ($9 + 0), null, null)::"hookhaven_data"."sketch")) as value +1ms
  postgraphql:postgres select to_json("hookhaven_data"."sketch_configuration"((($1 + 0), ($2 + 0), ($3 + 0), null, $4, $5, $6, $7, $8, ($9 + 0), null, null)::"hookhaven_data"."sketch")) as value +1ms
  postgraphql:postgres select to_json("hookhaven_data"."sketch_is_editable"((($1 + 0), ($2 + 0), ($3 + 0), null, $4, $5, $6, $7, $8, ($9 + 0), null, null)::"hookhaven_data"."sketch")) as value +1ms
  postgraphql:postgres select to_json("hookhaven_data"."sketch_favorite_count"((($1 + 0), ($2 + 0), ($3 + 0), null, $4, $5, $6, $7, $8, ($9 + 0), null, null)::"hookhaven_data"."sketch")) as value +1ms
  postgraphql:postgres select to_json("hookhaven_data"."sketch_fork_count"((($1 + 0), ($2 + 0), ($3 + 0), null, $4, $5, $6, $7, $8, ($9 + 0), null, null)::"hookhaven_data"."sketch")) as value +1ms
  postgraphql:postgres select to_json(__local_0__) as value from "hookhaven_data"."sketch_integration_event" as __local_0__ where "id" is not null and true and true and ("sketch_id" = $1) order by "id" using > limit $2 +4ms
  postgraphql:postgres select to_json(__local_0__) as value from "hookhaven_data"."sketch_version" as __local_0__ where "version_padded" is not null and "id" is not null and true and true and ("sketch_id" = $1) order by "version_padded" using >, "id" using > limit $2 +0ms
  postgraphql:postgres select to_json(__local_0__) as value from "hookhaven_data"."sketch_deploy" as __local_0__ where "id" is not null and true and true and ("sketch_id" = $1) order by "id" using > limit $2 +0ms
  postgraphql:postgres select to_json(__local_0__) as value from "hookhaven_data"."sketch_integration_event_run" as __local_0__ where "created_at" is not null and "id" is not null and true and true and ("sketch_id" = $1) order by "created_at" using >, "id" using > limit $2 +1ms
  postgraphql:postgres select row_to_json(__local_0__) as object from "hookhaven_data"."sketch" as __local_0__ where "id" = any($1) limit $2 +0ms
  postgraphql:postgres select row_to_json(__local_0__) as object from "hookhaven_data"."team" as __local_0__ where "id" = any($1) limit $2 +0ms
  postgraphql:postgres select row_to_json(__local_0__) as object from "hookhaven_data"."sketch_deploy" as __local_0__ where "id" = any($1) limit $2 +1ms
  postgraphql:postgres select null from "hookhaven_data"."sketch_integration_event" where ("id") < ($1) and ("sketch_id" = $2) limit 1 +58ms
  postgraphql:postgres select null from "hookhaven_data"."sketch_integration_event" where ("id") > ($1) and ("sketch_id" = $2) limit 1 +0ms
  postgraphql:postgres select row_to_json(__local_0__) as object from "hookhaven_data"."integration" as __local_0__ where "id" = any($1) limit $2 +2ms
  postgraphql:postgres select null from "hookhaven_data"."sketch_version" where ("version_padded", "id") < ($1, $2) and ("sketch_id" = $3) limit 1 +1ms
  postgraphql:postgres select null from "hookhaven_data"."sketch_version" where ("version_padded", "id") > ($1, $2) and ("sketch_id" = $3) limit 1 +0ms
  postgraphql:postgres select row_to_json(__local_0__) as object from "hookhaven_data"."sketch" as __local_0__ where "id" = any($1) limit $2 +0ms
  postgraphql:postgres select to_json("hookhaven_data"."sketch_deploy_is_ready"((($1 + 0), ($2 + 0), $3, $4, $5, $6, ($7 + 0), $8, null, null)::"hookhaven_data"."sketch_deploy")) as value +2ms
  postgraphql:postgres select null from "hookhaven_data"."sketch_deploy" where ("id") < ($1) and ("sketch_id" = $2) limit 1 +0ms
  postgraphql:postgres select null from "hookhaven_data"."sketch_deploy" where ("id") > ($1) and ("sketch_id" = $2) limit 1 +1ms
  postgraphql:postgres select row_to_json(__local_0__) as object from "hookhaven_data"."sketch_version" as __local_0__ where "id" = any($1) limit $2 +0ms
  postgraphql:postgres select row_to_json(__local_0__) as object from "hookhaven_data"."sketch" as __local_0__ where "id" = any($1) limit $2 +0ms
  postgraphql:postgres select null from "hookhaven_data"."sketch_integration_event_run" where ("created_at", "id") < ($1, $2) and ("sketch_id" = $3) limit 1 +20ms
  postgraphql:postgres select null from "hookhaven_data"."sketch_integration_event_run" where ("created_at", "id") > ($1, $2) and ("sketch_id" = $3) limit 1 +1ms
  postgraphql:postgres select row_to_json(__local_0__) as object from "hookhaven_data"."sketch_integration_event" as __local_0__ where "id" = any($1) limit $2 +1ms
  postgraphql:postgres select to_json(__local_0__) as value from "hookhaven_data"."integration" as __local_0__ where "id" is not null and true and true and ("team_id" = $1) order by "id" using < limit $2 +2ms
  postgraphql:postgres select row_to_json(__local_0__) as object from "hookhaven_data"."sketch_version" as __local_0__ where "id" = any($1) limit $2 +1ms
  postgraphql:postgres select row_to_json(__local_0__) as object from "hookhaven_data"."integration" as __local_0__ where "id" = any($1) limit $2 +49ms
  postgraphql:postgres select null from "hookhaven_data"."integration" where ("id") > ($1) and ("team_id" = $2) limit 1 +4ms
  postgraphql:postgres select null from "hookhaven_data"."integration" where ("id") < ($1) and ("team_id" = $2) limit 1 +0ms
  postgraphql:postgres commit +11ms
  postgraphql:request GraphQL query has been executed. +1ms
  postgraphql:request GraphQL query request finished. +2ms
0 error(s) in 230.46ms :: query ...
```

# After

```
  postgraphql:request GraphQL query request has begun. +0ms
  postgraphql:request GraphQL query is parsed. +25ms
  postgraphql:request GraphQL query is validated. +65ms
  postgraphql:postgres begin +2ms
  postgraphql:postgres select set_config($1, $2, true) +3ms
  postgraphql:postgres select json_build_object($1::text, __local_0__."id", $2::text, __local_0__."updated_at", $3::text, __local_0__."owner_id", $4::text, __local_0__."team_id", $5::text, "hookhaven_data"."sketch_is_favorite"(__local_0__), $6::text, __local_0__."visibility", $7::text, __local_0__."name", $8::text, __local_0__."body", $9::text, "hookhaven_data"."sketch_settings"(__local_0__), $10::text, "hookhaven_data"."sketch_configuration"(__local_0__), $11::text, "hookhaven_data"."sketch_is_editable"(__local_0__), $12::text, __local_0__."sketch_deploy_id", $13::text, "hookhaven_data"."sketch_favorite_count"(__local_0__), $14::text, "hookhaven_data"."sketch_fork_count"(__local_0__), $15::text, __local_0__."readme", $16::text, __local_0__."upstream_id") as object from "hookhaven_data"."sketch" as __local_0__ where "id" = any($17) limit $18 +21ms
  postgraphql:postgres select json_build_object($1::text, __local_0__."id", $2::text, __local_0__."payload", $3::text, __local_0__."meta", $4::text, __local_0__."headers", $5::text, __local_0__."query", $6::text, __local_0__."type", $7::text, __local_0__."integration_id", $8::text, __local_0__."created_at", $9::text, __local_0__."within_allowance", $10::text, __local_0__."skipped", $11::text, __local_0__."skip_reason", $12::text, __local_0__."sketch_id", $13::text, __local_0__."processed_with_id") as value from "hookhaven_data"."sketch_integration_event" as __local_0__ where "id" is not null and true and true and ("sketch_id" = $14) order by "id" using > limit $15 +30ms
  postgraphql:postgres select json_build_object($1::text, __local_0__."id", $2::text, __local_0__."created_at", $3::text, __local_0__."version", $4::text, __local_0__."published", $5::text, __local_0__."yanked", $6::text, __local_0__."sketch_id", $7::text, __local_0__."upstream_id") as value from "hookhaven_data"."sketch_version" as __local_0__ where "version_padded" is not null and "id" is not null and true and true and ("sketch_id" = $8) order by "version_padded" using >, "id" using > limit $9 +1ms
  postgraphql:postgres select json_build_object($1::text, __local_0__."id", $2::text, "hookhaven_data"."sketch_deploy_is_ready"(__local_0__), $3::text, __local_0__."failure_reason", $4::text, __local_0__."updated_at", $5::text, __local_0__."created_at", $6::text, __local_0__."sketch_version_id", $7::text, __local_0__."sketch_id") as value from "hookhaven_data"."sketch_deploy" as __local_0__ where "id" is not null and true and true and ("sketch_id" = $8) order by "id" using > limit $9 +2ms
  postgraphql:postgres select json_build_object($1::text, __local_0__."id", $2::text, __local_0__."created_at", $3::text, __local_0__."error", $4::text, __local_0__."sketch_integration_event_id", $5::text, __local_0__."sketch_id", $6::text, __local_0__."sketch_deploy_id", $7::text, "hookhaven_data"."sketch_integration_event_run_last_block_id"(__local_0__)) as value from "hookhaven_data"."sketch_integration_event_run" as __local_0__ where "created_at" is not null and "id" is not null and true and true and ("sketch_id" = $8) order by "created_at" using >, "id" using > limit $9 +1ms
  postgraphql:postgres select json_build_object($1::text, __local_0__."id", $2::text, __local_0__."team_id", $3::text, __local_0__."owner_id", $4::text, __local_0__."upstream_id", $5::text, __local_0__."sketch_deploy_id") as object from "hookhaven_data"."sketch" as __local_0__ where "id" = any($6) limit $7 +2ms
  postgraphql:postgres select json_build_object($1::text, __local_0__."name", $2::text, __local_0__."id") as object from "hookhaven_data"."team" as __local_0__ where "id" = any($3) limit $4 +1ms
  postgraphql:postgres select json_build_object($1::text, __local_0__."id", $2::text, __local_0__."updated_at", $3::text, __local_0__."sketch_version_id", $4::text, __local_0__."sketch_id") as object from "hookhaven_data"."sketch_deploy" as __local_0__ where "id" = any($5) limit $6 +0ms
  postgraphql:postgres select null from "hookhaven_data"."sketch_integration_event" where ("id") < ($1) and ("sketch_id" = $2) limit 1 +70ms
  postgraphql:postgres select null from "hookhaven_data"."sketch_integration_event" where ("id") > ($1) and ("sketch_id" = $2) limit 1 +1ms
  postgraphql:postgres select json_build_object($1::text, __local_0__."name", $2::text, __local_0__."id", $3::text, __local_0__."user_id", $4::text, __local_0__."team_id") as object from "hookhaven_data"."integration" as __local_0__ where "id" = any($5) limit $6 +8ms
  postgraphql:postgres select null from "hookhaven_data"."sketch_version" where ("version_padded", "id") < ($1, $2) and ("sketch_id" = $3) limit 1 +3ms
  postgraphql:postgres select null from "hookhaven_data"."sketch_version" where ("version_padded", "id") > ($1, $2) and ("sketch_id" = $3) limit 1 +1ms
  postgraphql:postgres select json_build_object($1::text, __local_0__."id", $2::text, __local_0__."sketch_deploy_id", $3::text, __local_0__."updated_at", $4::text, __local_0__."team_id", $5::text, __local_0__."owner_id", $6::text, __local_0__."upstream_id") as object from "hookhaven_data"."sketch" as __local_0__ where "id" = any($7) limit $8 +0ms
  postgraphql:postgres select null from "hookhaven_data"."sketch_deploy" where ("id") < ($1) and ("sketch_id" = $2) limit 1 +1ms
  postgraphql:postgres select null from "hookhaven_data"."sketch_deploy" where ("id") > ($1) and ("sketch_id" = $2) limit 1 +1ms
  postgraphql:postgres select json_build_object($1::text, __local_0__."version", $2::text, __local_0__."id", $3::text, __local_0__."sketch_id", $4::text, __local_0__."upstream_id") as object from "hookhaven_data"."sketch_version" as __local_0__ where "id" = any($5) limit $6 +0ms
  postgraphql:postgres select json_build_object($1::text, __local_0__."id", $2::text, __local_0__."sketch_deploy_id", $3::text, __local_0__."updated_at", $4::text, __local_0__."team_id", $5::text, __local_0__."owner_id", $6::text, __local_0__."upstream_id") as object from "hookhaven_data"."sketch" as __local_0__ where "id" = any($7) limit $8 +0ms
  postgraphql:postgres select null from "hookhaven_data"."sketch_integration_event_run" where ("created_at", "id") < ($1, $2) and ("sketch_id" = $3) limit 1 +37ms
  postgraphql:postgres select null from "hookhaven_data"."sketch_integration_event_run" where ("created_at", "id") > ($1, $2) and ("sketch_id" = $3) limit 1 +1ms
  postgraphql:postgres select json_build_object($1::text, __local_0__."id", $2::text, __local_0__."created_at", $3::text, __local_0__."integration_id", $4::text, __local_0__."type", $5::text, __local_0__."query", $6::text, __local_0__."headers", $7::text, __local_0__."payload", $8::text, __local_0__."meta", $9::text, __local_0__."within_allowance", $10::text, __local_0__."skipped", $11::text, __local_0__."skip_reason", $12::text, __local_0__."sketch_id", $13::text, __local_0__."processed_with_id") as object from "hookhaven_data"."sketch_integration_event" as __local_0__ where "id" = any($14) limit $15 +22ms
  postgraphql:postgres select json_build_object($1::text, __local_0__."id", $2::text, __local_0__."team_id", $3::text, __local_0__."type", $4::text, __local_0__."name", $5::text, __local_0__."user_id") as value from "hookhaven_data"."integration" as __local_0__ where "id" is not null and true and true and ("team_id" = $6) order by "id" using < limit $7 +3ms
  postgraphql:postgres select json_build_object($1::text, __local_0__."version", $2::text, __local_0__."id", $3::text, __local_0__."sketch_id", $4::text, __local_0__."upstream_id") as object from "hookhaven_data"."sketch_version" as __local_0__ where "id" = any($5) limit $6 +1ms
  postgraphql:postgres select json_build_object($1::text, __local_0__."name", $2::text, __local_0__."id", $3::text, __local_0__."user_id", $4::text, __local_0__."team_id") as object from "hookhaven_data"."integration" as __local_0__ where "id" = any($5) limit $6 +68ms
  postgraphql:postgres select null from "hookhaven_data"."integration" where ("id") > ($1) and ("team_id" = $2) limit 1 +1ms
  postgraphql:postgres select null from "hookhaven_data"."integration" where ("id") < ($1) and ("team_id" = $2) limit 1 +0ms
  postgraphql:postgres commit +13ms
  postgraphql:request GraphQL query has been executed. +1ms
  postgraphql:request GraphQL query request finished. +2ms
0 error(s) in 398.86ms ::
```

### DIFF
--- a/src/graphql/schema/collection/createCollectionGqlType.ts
+++ b/src/graphql/schema/collection/createCollectionGqlType.ts
@@ -46,13 +46,16 @@ export default function createCollectionGqlType<TValue> (
       Array.from(type.fields).map(
         <TFieldValue>([fieldName, field]: [string, ObjectType.Field<TValue, TFieldValue>]) => {
           const { gqlType, intoGqlOutput } = getGqlOutputType(buildToken, field.type)
+          const formattedFieldName = formatName.field(fieldName)
           return {
-            key: formatName.field(fieldName),
+            key: formattedFieldName
             value: {
               description: field.description,
               type: gqlType,
               sqlExpression: (aliasIdentifier) => sql.query`${sql.identifier(aliasIdentifier)}.${sql.identifier(fieldName)}`,
-              resolve: (value: TValue): mixed => intoGqlOutput(field.getValue(value)),
+              resolve: (value) => {
+                return value.get(formattedFieldName)
+              }
             },
           }
         },
--- a/src/graphql/schema/collection/createCollectionGqlType.ts
+++ b/src/graphql/schema/collection/createCollectionGqlType.ts
@@ -7,6 +7,7 @@ import createConnectionGqlField from '../connection/createConnectionGqlField'
 import BuildToken from '../BuildToken'
 import createCollectionRelationTailGqlFieldEntries from './createCollectionRelationTailGqlFieldEntries'
 import getConditionGqlType from './getConditionGqlType'
+import { sql } from '../../../postgres/utils'
 
 /**
  * Creates the output object type for a collection. This type will include all
@@ -50,6 +51,7 @@ export default function createCollectionGqlType<TValue> (
             value: {
               description: field.description,
               type: gqlType,
+              sqlExpression: () => sql.identifier(fieldName),
               resolve: (value: TValue): mixed => intoGqlOutput(field.getValue(value)),
             },
           }
--- a/src/graphql/schema/collection/createCollectionGqlType.ts
+++ b/src/graphql/schema/collection/createCollectionGqlType.ts
@@ -51,7 +51,7 @@ export default function createCollectionGqlType<TValue> (
             value: {
               description: field.description,
               type: gqlType,
-              sqlExpression: () => sql.identifier(fieldName),
+              sqlExpression: (aliasIdentifier) => sql.query`${sql.identifier(aliasIdentifier)}.${sql.identifier(fieldName)}`,
               resolve: (value: TValue): mixed => intoGqlOutput(field.getValue(value)),
             },
           }
--- a/src/graphql/schema/collection/createCollectionGqlType.ts
+++ b/src/graphql/schema/collection/createCollectionGqlType.ts
@@ -52,9 +52,10 @@ export default function createCollectionGqlType<TValue> (
             value: {
               description: field.description,
               type: gqlType,
+              sqlName: () => fieldName,
               sqlExpression: (aliasIdentifier) => sql.query`${sql.identifier(aliasIdentifier)}.${sql.identifier(fieldName)}`,
               resolve: (value) => {
-                return value.get(formattedFieldName)
+                return value.get(fieldName)
               }
             },
           }
--- a/src/graphql/schema/collection/createCollectionQueryFieldEntries.ts
+++ b/src/graphql/schema/collection/createCollectionQueryFieldEntries.ts
@@ -93,11 +93,11 @@ function createCollectionPrimaryKeyField <TValue, TKey>(
       },
     },
 
-    async resolve (_source, args, context): Promise<mixed> {
+    async resolve (_source, args, context, resolveInfo): Promise<mixed> {
       const result = idSerde.deserialize(inventory, args[options.nodeIdFieldName] as string)
       if (result.collection !== collection) throw new Error(`The provided id is for collection '${result.collection.name}', not the expected collection '${collection.name}'.`)
       if (!keyType.isTypeOf(result.keyValue)) throw new Error(`The provided id is not of the correct type.`)
-      const value = await collectionKey.read!(context, result.keyValue)
+      const value = await collectionKey.read!(context, result.keyValue, resolveInfo)
       if (value == null) return
       return intoGqlOutput(value)
     },
@@ -123,9 +123,9 @@ function createCollectionKeyField <TValue, TKey>(
   return {
     type: collectionGqlType,
     args: buildObject(inputHelpers.fieldEntries),
-    async resolve (_source, args, context): Promise<mixed> {
+    async resolve (_source, args, context, resolveInfo): Promise<mixed> {
       const key = inputHelpers.getKeyFromInput(args)
-      const value = await collectionKey.read!(context, key)
+      const value = await collectionKey.read!(context, key, resolveInfo)
       if (value == null) return
       return intoGqlOutput(value)
     },
--- a/src/graphql/schema/connection/createConnectionGqlField.ts
+++ b/src/graphql/schema/connection/createConnectionGqlField.ts
@@ -41,7 +41,6 @@ export default function createConnectionGqlField <TSource, TInput, TItemValue>(
   return {
     description: config.description || `Reads and enables paginatation through a set of ${scrib.type(gqlType)}.`,
     type: getConnectionGqlType(buildToken, paginator),
-    relatedGqlType: gqlType,
     args: buildObject<GraphQLArgumentConfig>([
       // Only include an `orderBy` field if there are ways in which we can
       // order.
@@ -158,14 +157,12 @@ export function _createConnectionGqlType <TInput, TItemValue>(
       },
       edges: {
         description: `A list of edges which contains the ${scrib.type(gqlType)} and cursor to aid in pagination.`,
-        relatedGqlType: gqlType,
         type: new GraphQLList(gqlEdgeType),
         resolve: <TCursor>({ orderingName, page }: Connection<TInput, TItemValue, TCursor>): Array<Edge<TInput, TItemValue, TCursor>> =>
           page.values.map(({ cursor, value }) => ({ paginator, orderingName, cursor, value })),
       },
       nodes: {
         description: `A list of ${scrib.type(gqlType)} objects.`,
-        relatedGqlType: gqlType,
         type: new GraphQLList(gqlType),
         resolve: ({ page }: Connection<TInput, TItemValue, mixed>): Array<TItemValue> =>
           page.values.map(({ value }) => value),
@@ -188,6 +185,7 @@ export function _createEdgeGqlType <TInput, TItemValue>(
   return new GraphQLObjectType({
     name: formatName.type(`${paginator.name}-edge`),
     description: `A ${scrib.type(gqlType)} edge in the connection.`,
+    relatedGqlType: gqlType,
     fields: () => ({
       cursor: {
         description: 'A cursor for use in pagination.',
--- a/src/graphql/schema/connection/createConnectionGqlField.ts
+++ b/src/graphql/schema/connection/createConnectionGqlField.ts
@@ -41,6 +41,7 @@ export default function createConnectionGqlField <TSource, TInput, TItemValue>(
   return {
     description: config.description || `Reads and enables paginatation through a set of ${scrib.type(gqlType)}.`,
     type: getConnectionGqlType(buildToken, paginator),
+    relatedGqlType: gqlType,
     args: buildObject<GraphQLArgumentConfig>([
       // Only include an `orderBy` field if there are ways in which we can
       // order.
@@ -115,7 +116,7 @@ export default function createConnectionGqlField <TSource, TInput, TItemValue>(
       const ordering = paginator.orderings.get(orderingName) as Paginator.Ordering<TInput, TItemValue, TCursor>
 
       // Finally, actually get the page data.
-      const page = await ordering.readPage(context, input, pageConfig, resolveInfo)
+      const page = await ordering.readPage(context, input, pageConfig, resolveInfo, gqlType)
 
       return {
         paginator,
@@ -142,6 +143,7 @@ export function _createConnectionGqlType <TInput, TItemValue>(
   return new GraphQLObjectType({
     name: formatName.type(`${paginator.name}-connection`),
     description: `A connection to a list of ${scrib.type(gqlType)} values.`,
+    relatedGqlType: gqlType,
     fields: () => ({
       pageInfo: {
         description: 'Information to aid in pagination.',
@@ -156,12 +158,14 @@ export function _createConnectionGqlType <TInput, TItemValue>(
       },
       edges: {
         description: `A list of edges which contains the ${scrib.type(gqlType)} and cursor to aid in pagination.`,
+        relatedGqlType: gqlType,
         type: new GraphQLList(gqlEdgeType),
         resolve: <TCursor>({ orderingName, page }: Connection<TInput, TItemValue, TCursor>): Array<Edge<TInput, TItemValue, TCursor>> =>
           page.values.map(({ cursor, value }) => ({ paginator, orderingName, cursor, value })),
       },
       nodes: {
         description: `A list of ${scrib.type(gqlType)} objects.`,
+        relatedGqlType: gqlType,
         type: new GraphQLList(gqlType),
         resolve: ({ page }: Connection<TInput, TItemValue, mixed>): Array<TItemValue> =>
           page.values.map(({ value }) => value),
--- a/src/graphql/schema/connection/createConnectionGqlField.ts
+++ b/src/graphql/schema/connection/createConnectionGqlField.ts
@@ -75,6 +75,7 @@ export default function createConnectionGqlField <TSource, TInput, TItemValue>(
       source: TSource,
       args: ConnectionArgs<TCursor>,
       context: mixed,
+      resolveInfo: mixed,
     ): Promise<Connection<TInput, TItemValue, TCursor>> {
       const {
         orderBy: orderingName,
@@ -114,7 +115,7 @@ export default function createConnectionGqlField <TSource, TInput, TItemValue>(
       const ordering = paginator.orderings.get(orderingName) as Paginator.Ordering<TInput, TItemValue, TCursor>
 
       // Finally, actually get the page data.
-      const page = await ordering.readPage(context, input, pageConfig)
+      const page = await ordering.readPage(context, input, pageConfig, resolveInfo)
 
       return {
         paginator,
--- a/src/graphql/schema/node/createNodeFieldEntry.ts
+++ b/src/graphql/schema/node/createNodeFieldEntry.ts
@@ -18,7 +18,7 @@ export default function createNodeFieldEntry (buildToken: BuildToken): [string, 
         type: new GraphQLNonNull(GraphQLID),
       },
     },
-    async resolve <TValue, TKey>(_source: mixed, args: { [key: string]: mixed }, context: mixed): Promise<TValue | symbol | null> {
+    async resolve <TValue, TKey>(_source: mixed, args: { [key: string]: mixed }, context: mixed, resolveInfo: mixed): Promise<TValue | symbol | null> {
       let deserializationResult: { collection: Collection<TValue>, keyValue: TKey }
       const idString = args[options.nodeIdFieldName]
 
@@ -46,7 +46,7 @@ export default function createNodeFieldEntry (buildToken: BuildToken): [string, 
       if (!primaryKey || !primaryKey.read)
         throw new Error(`Invalid id, no readable primary key on collection named '${collection.name}'.`)
 
-      const value = await primaryKey.read(context, keyValue)
+      const value = await primaryKey.read(context, keyValue, resolveInfo)
 
       // If the value is null, end early.
       if (value == null)
--- a/src/graphql/schema/node/createNodeFieldEntry.ts
+++ b/src/graphql/schema/node/createNodeFieldEntry.ts
@@ -46,7 +46,7 @@ export default function createNodeFieldEntry (buildToken: BuildToken): [string, 
       if (!primaryKey || !primaryKey.read)
         throw new Error(`Invalid id, no readable primary key on collection named '${collection.name}'.`)
 
-      const value = await primaryKey.read(context, keyValue, resolveInfo)
+      const value = await primaryKey.read(context, keyValue, resolveInfo, getGqlOutputType(buildToken, collection.type).gqlType)
 
       // If the value is null, end early.
       if (value == null)
--- a/src/postgres/inventory/collection/PgCollectionKey.ts
+++ b/src/postgres/inventory/collection/PgCollectionKey.ts
@@ -137,6 +137,7 @@ class PgCollectionKey implements CollectionKey<PgClassType.Value, PgCollectionKe
   private _getSelectLoader (client: Client): DataLoader<PgClassType.Value, PgClassType.Value | null> {
     return new DataLoader<PgClassType.Value, PgClassType.Value | null>(
       async (keys: Array<PgClassType.Value>): Promise<Array<PgClassType.Value | null>> => {
+        console.dir(keys)
         const aliasIdentifier = Symbol()
 
         // For every key we have, generate a select statement then combine
@@ -187,7 +188,16 @@ class PgCollectionKey implements CollectionKey<PgClassType.Value, PgCollectionKe
           limit ${sql.value(keys.length)}
         `)
 
-        const { rows } = await client.query(query)
+        let rows
+        try {
+          ({ rows } = await client.query(query))
+        } catch (e) {
+          console.log(__filename)
+          console.dir(e)
+          console.log("ERROR!!!")
+          console.log(query)
+          throw e
+        }
 
         const values = new Map(rows.map<[string, PgClassType.Value]>(({ object }) => {
           const value = this.collection.type.transformPgValueIntoValue(object)
--- a/src/postgres/inventory/collection/PgCollectionKey.ts
+++ b/src/postgres/inventory/collection/PgCollectionKey.ts
@@ -120,10 +120,10 @@ class PgCollectionKey implements CollectionKey<PgClassType.Value, PgCollectionKe
    * Reads a value if a user can select from this class. Batches requests to
    * the same client in the background.
    */
-  public read: ((context: mixed, key: PgClassType.Value) => Promise<PgClassType.Value | null>) | null = (
+  public read: ((context: mixed, key: PgClassType.Value, resolveInfo: mixed) => Promise<PgClassType.Value | null>) | null = (
     !this._pgClass.isSelectable
       ? null
-      : (context: mixed, key: PgClassType.Value): Promise<PgClassType.Value | null> =>
+      : (context: mixed, key: PgClassType.Value, resolveInfo: mixed): Promise<PgClassType.Value | null> =>
         this._getSelectLoader(pgClientFromContext(context)).load(key)
   )
 
--- a/src/postgres/inventory/collection/PgCollectionKey.ts
+++ b/src/postgres/inventory/collection/PgCollectionKey.ts
@@ -9,7 +9,7 @@ import PgType from '../type/PgType'
 import PgClassType from '../type/PgClassType'
 import getTypeFromPgType from '../type/getTypeFromPgType'
 import PgCollection from './PgCollection'
-import getSelectFragment from '../paginator/getSelectFragment'
+import {getFieldsFromResolveInfo, getSelectFragmentFromFields} from '../paginator/getSelectFragment'
 
 /**
  * Creates a key from some types of Postgres constraints including primary key
@@ -125,7 +125,7 @@ class PgCollectionKey implements CollectionKey<PgClassType.Value, PgCollectionKe
     !this._pgClass.isSelectable
       ? null
       : (context: mixed, key: PgClassType.Value, resolveInfo: mixed, collectionGqlType: mixed): Promise<PgClassType.Value | null> =>
-        this._getSelectLoader(pgClientFromContext(context), resolveInfo, collectionGqlType).load(key)
+        this._getSelectLoader(pgClientFromContext(context)).load({key, resolveInfo, collectionGqlType})
   )
 
   /**
@@ -134,10 +134,17 @@ class PgCollectionKey implements CollectionKey<PgClassType.Value, PgCollectionKe
    *
    * @private
    */
+  @memoizeMethod
   private _getSelectLoader (client: Client, resolveInfo: mixed, collectionGqlType: mixed): DataLoader<PgClassType.Value, PgClassType.Value | null> {
     return new DataLoader<PgClassType.Value, PgClassType.Value | null>(
-      async (keys: Array<PgClassType.Value>): Promise<Array<PgClassType.Value | null>> => {
+      async (keysAndStuff: mixed): Promise<Array<PgClassType.Value | null>> => {
         const aliasIdentifier = Symbol()
+        const keys = keysAndStuff.map(({key}) => key)
+        const fieldses = keysAndStuff.map(
+          ({resolveInfo, collectionGqlType}) =>
+            getFieldsFromResolveInfo(resolveInfo, aliasIdentifier, collectionGqlType)
+        )
+        const fields = Object.assign({}, ...fieldses)
 
         // For every key we have, generate a select statement then combine
         // those select statements with `union all`. This approach has a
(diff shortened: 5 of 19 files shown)


benjie:
Latest commits solve some of this, but there's some missing fields I need to get to the bottom of (seems some fragments aren't being processed...)

benjie:
Latest commits get everything working pretty well - outputs identical data for a complex query (except for one minor issue with a sort field not being fetched also which hopefully Caleb can help me fix). I've tidied it up a bit so it should be more accessible now. I do not recommend digging through the git history 😳

src/graphql/schema/collection/createCollectionGqlType.ts
```diff
@@ -45,12 +46,17 @@ export default function createCollectionGqlType<TValue> (
...
           const { gqlType, intoGqlOutput } = getGqlOutputType(buildToken, field.type)
+          const formattedFieldName = formatName.field(fieldName)
           return {
-            key: formatName.field(fieldName),
+            key: formattedFieldName
             value: {
               description: field.description,
               type: gqlType,
-              resolve: (value: TValue): mixed => intoGqlOutput(field.getValue(value)),
+              sqlName: () => fieldName,
```
calebmer: In my initial design for PostGraphQL, the GraphQL and SQL bits were kept completely separate. I’d like to see if we can maintain that pattern as it would make it easier to support other noSQL backends in the future. If it proves too difficult then we can scrap that ideal, but I’d like to try 😊

Here specifically we have `sqlName` and `sqlExpression` in the GraphQL type creation code, and we are also importing the SQL utility from the Postgres package. I’m going to briefly lay out how I saw this working, and we can discuss how to go from there.

I thought that instead of having fields like `sqlName` or `sqlExpression` that we could instead have a `dependencies` field which takes an array of objects from `src/interface`. So for example, the `dependencies` property could have an [`ObjectType.Field`](https://github.com/calebmer/postgraphql/blob/b54d05acafa8ada1d27e1a0e2a4cac091c620fe3/src/interface/type/ObjectType.ts#L34) item, or a [`CollectionKey`](https://github.com/calebmer/postgraphql/blob/b54d05acafa8ada1d27e1a0e2a4cac091c620fe3/src/interface/collection/CollectionKey.ts) item (for node ids). GraphQL would not know anything about these values, but there Postgres implementations (like [`PgObjectType.Field`](https://github.com/calebmer/postgraphql/blob/b54d05acafa8ada1d27e1a0e2a4cac091c620fe3/src/postgres/inventory/type/PgClassType.ts#L110)) would have the SQL information necessary to only select an optimal subset of information.

We could then also add to `src/interface` a `ReadOptimizations` type which would look something like:

```ts
interface ReadOptimizations {
  dependencies: {
    fields: Array<ObjectType.Field>,
    collectionKeys: Array<CollectionKey>,
    ...
  },
}
```

A `ReadOptimizations` object would be generated in `src/graphql` by looking at `resolveInfo` and the GraphQL type. `ReadOptimizations` would then be passed to the interface read methods instead of `resolveInfo` where Postgres could use the information to generate an optimized query.

Admittedly, I was not thinking about how we could optimize procedures when I first thought up this abstraction, but it shouldn’t be too hard, I think. We would just add a private, optional, `_pgProcedures` property to `ReadOptimizations.dependencies` that our Postgres implementation could pick up.

What do you think?

src/graphql/schema/collection/createCollectionGqlType.ts
```diff
@@ -45,12 +46,17 @@ export default function createCollectionGqlType<TValue> (
...
+              sqlExpression: (aliasIdentifier) => sql.query`${sql.identifier(aliasIdentifier)}.${sql.identifier(fieldName)}`,
+              resolve: (value) => {
+                return value.get(fieldName)
```
calebmer: This looks different than the code that was there, `resolve: (value: TValue): mixed => intoGqlOutput(field.getValue(value))`. Will I understand why as I continue? 😊

src/graphql/schema/collection/createCollectionQueryFieldEntries.ts
```diff
@@ -93,11 +93,11 @@ function createCollectionPrimaryKeyField <TValue, TKey>(
       },
     },
 
-    async resolve (_source, args, context): Promise<mixed> {
+    async resolve (_source, args, context, resolveInfo): Promise<mixed> {
       const result = idSerde.deserialize(inventory, args[options.nodeIdFieldName] as string)
       if (result.collection !== collection) throw new Error(`The provided id is for collection '${result.collection.name}', not the expected collection '${collection.name}'.`)
       if (!keyType.isTypeOf(result.keyValue)) throw new Error(`The provided id is not of the correct type.`)
-      const value = await collectionKey.read!(context, result.keyValue)
+      const value = await collectionKey.read!(context, result.keyValue, resolveInfo)
```
calebmer: This here is also where the nice GraphQL/Postgres separation breaks down. If possible, it would be great to pass in a `ReadOptimizations` object like I mention above.

src/graphql/schema/connection/createConnectionGqlField.ts
```diff
@@ -141,6 +142,7 @@ export function _createConnectionGqlType <TInput, TItemValue>(
   return new GraphQLObjectType({
     name: formatName.type(`${paginator.name}-connection`),
     description: `A connection to a list of ${scrib.type(gqlType)} values.`,
+    relatedGqlType: gqlType,
```
calebmer: Why was this added?

src/graphql/schema/node/createNodeFieldEntry.ts
```diff
@@ -46,7 +46,7 @@ export default function createNodeFieldEntry (buildToken: BuildToken): [string,
       if (!primaryKey || !primaryKey.read)
         throw new Error(`Invalid id, no readable primary key on collection named '${collection.name}'.`)
 
-      const value = await primaryKey.read(context, keyValue)
+      const value = await primaryKey.read(context, keyValue, resolveInfo, getGqlOutputType(buildToken, collection.type).gqlType)
```
calebmer: We should look to see if there is a better way to get the GraphQL type here instead of calling `getGqlOutputType`. Maybe there isn’t…

src/postgres/inventory/collection/PgCollectionKey.ts
```diff
@@ -134,10 +135,16 @@ class PgCollectionKey implements CollectionKey<PgClassType.Value, PgCollectionKe
    * @private
    */
   @memoizeMethod
-  private _getSelectLoader (client: Client): DataLoader<PgClassType.Value, PgClassType.Value | null> {
+  private _getSelectLoader (client: Client, resolveInfo: mixed, collectionGqlType: mixed): DataLoader<PgClassType.Value, PgClassType.Value | null> {
     return new DataLoader<PgClassType.Value, PgClassType.Value | null>(
-      async (keys: Array<PgClassType.Value>): Promise<Array<PgClassType.Value | null>> => {
+      async (keysAndStuff: mixed): Promise<Array<PgClassType.Value | null>> => {
```
calebmer: This `DataLoader` thing is definitely an interesting and complex problem…

It would be great if we could scan the GraphQL query in its entirety before executing and get a `ReadOptimizations`, or object of a similar concept, that will then be globally static across the request.

src/postgres/inventory/paginator/PgPaginatorOrderingAttributes.ts
```diff
@@ -99,7 +102,7 @@ implements Paginator.Ordering<TInput, PgClassType.Value, AttributesCursor> {
       ${_offset != null ? sql.query`offset ${sql.value(_offset)}` : sql.query``}
     `)
 
-    let { rows } = await client.query(query)
+    const { rows } = await client.query(query)
```
calebmer: Don’t we assign `rows` further down? Does this work?

src/postgres/inventory/type/PgClassType.ts
```diff
@@ -80,9 +80,21 @@ class PgClassType extends PgType<PgRow> implements ObjectType<PgRow> {
...
+    Array.from(this.fields).filter<[string, mixed]>(([fieldName, field]) => pgValue.hasOwnProperty(field.pgAttribute.name)).map<[string, mixed]>(([fieldName, field]) => {
+      obj[fieldName] = field.type.transformPgValueIntoValue(pgValue[field.pgAttribute.name])
+    })
+    // Pass through other fields unchanged
+    for (const key in pgValue) {
+      obj[key] = pgValue[key];
+    }
+
+    const pairs = [];
+    for (const key in obj) {
```
calebmer: You could instantiate the map where you instantiate `obj` and use `map.set` instead of doing this ceremony with the array pairs 😊

src/postgres/inventory/type/PgClassType.ts
```diff
@@ -80,9 +80,21 @@ class PgClassType extends PgType<PgRow> implements ObjectType<PgRow> {
...
-    return new Map(Array.from(this.fields).map<[string, mixed]>(([fieldName, field]) => {
-      return [fieldName, field.type.transformPgValueIntoValue(pgValue[field.pgAttribute.name])]
-    }))
+    // Parse official pg fields
+    const obj = {};
+    Array.from(this.fields).filter<[string, mixed]>(([fieldName, field]) => pgValue.hasOwnProperty(field.pgAttribute.name)).map<[string, mixed]>(([fieldName, field]) => {
+      obj[fieldName] = field.type.transformPgValueIntoValue(pgValue[field.pgAttribute.name])
+    })
+    // Pass through other fields unchanged
+    for (const key in pgValue) {
```
calebmer: Ok, so I think I understand why we are using `map.get` straight up now instead of `field.getValue`.

src/postgres/inventory/paginator/PgPaginatorOrderingAttributes.ts
benjie: Apparently TypeScript doesn't care!

benjie:
Thanks for the feedback; this is merely a proof of concept at the moment - I've not bothered with code quality as I'm just exploring the possible. As it happens the other PR I've built on this successfully uses DataLoader even with deep queries to efficiently fetch all the required data, so that's not the problem it was. I think I'm going to reimplement both PRs into one going forward; but there's a whole bunch of things I wanted help with first, such as determining the required (implicit) columns, such as for relations or __id; getting the gqlType without having to pass relatedType, etc.

src/graphql/schema/connection/createConnectionGqlField.ts
benjie: It was the easiest way to concretely determine whether to keep travelling down the AST looking for related nodes; without this we either over-select (making us fetch fields we don't need) or we have to do duck typing by looking for edges/nodes and inferring from there (which makes me uncomfortable); or placing constraints on the naming patterns so if it's looking for a Team then it'd know to look for TeamConnection/TeamEdge/etc. I suggest we keep this unless you know a better way.

src/graphql/schema/collection/createCollectionGqlType.ts
benjie: I agree with the desire for separation (I've not bothered yet because I was concentrating on just getting *something* working, but I will be looking at it during refactoring); but I think a full two-stage parse of the entire AST is a bit ambitious at this stage - I'd rather get a step in the right direction out than wait for perfection that may never come. (Incidentally this is the approach JoinMonster takes: transforming the entire AST and using it to build the SQL queries)

src/graphql/schema/collection/createCollectionGqlType.ts
benjie: Ah yeah, I had to change this in an earlier incarnation and forgot to change it back - good catch

src/graphql/schema/node/createNodeFieldEntry.ts
benjie: I was hoping you could help me out here! Also in a few places I've passed through the NotNull'd version of the type which I then have to strip in `getSelectFragment`. You've done an impressive job of writing very high quality code, but despite this because PostGraphQL is such a big project it does make a number of these things quite challenging to figure out - there's a lot going on!

src/postgres/inventory/collection/PgCollectionKey.ts
benjie: It gets more challenging when you factor in the relation fetching from #342, because then merging the fields from across different places would also merge the subqueries - fetching WAY more data than is necessary.

Since doing more work on #342 I actually think the better way to deal with this is to eliminate DataLoader entirely, and instead to inline all relationships via Join-Monster style queries that just pull down everything for a root node. Then the number of queries would be basically the same as the number of root nodes on the GraphQL query, which I think makes a lot of sense.

I don't think DataLoader is a good match once you start selecting specific fields and adding subqueries or JOINs into the mix

src/postgres/inventory/paginator/PgPaginatorOrderingAttributes.ts
benjie: (Incidentally the assigning lower down is unnecessary, `Array.prototype.reverse` is an in-place modification

src/postgres/inventory/type/PgClassType.ts
benjie: I've not checked - does field.getValue have some magic coercion in it? If so do make sure I don't lose that!

src/postgres/inventory/type/PgClassType.ts
benjie: Thanks, I've not played with maps yet so probably ought to spend some time reading the docs.

Incidentally why did you use maps here rather than just vanilla JS objects?

benjie:
In fact... I'm going to close this in favour of #342

src/graphql/schema/collection/createCollectionGqlType.ts
calebmer: > but I think a full two-stage parse of the entire AST is a bit ambitious at this stage

The more I think about it, you’re probably right for this case. What gets really tricky is if we have procedure fields. Do we call that procedure for every object of that type? We really shouldn’t…

Instead, we should see if there is anything we can do to create effectively one data loader for every field. One data loader per request turns out to be hard for us to use here, but one data loader per field could work well.

src/graphql/schema/connection/createConnectionGqlField.ts
calebmer: Gotcha. That makes sense 👍

Maybe another name or some documentation on this name could be helpful, though.

src/graphql/schema/node/createNodeFieldEntry.ts
calebmer: `getNullableType` from `graphql` is really great for getting the nullable GraphQL type, and I believe `src/interface` has a `getNonNullType` function because nullability is reversed in `src/interface` 😊

Here we could build an index keyed by the collection’s name of type `Map<string, GraphQLObjectType>`. I think I’d feel better if we did that. I don’t really want to make any guarantees that `getGqlOutputType` and similar functions are fast enough for runtime 😊

src/postgres/inventory/collection/PgCollectionKey.ts
calebmer: Yep, I think you’re right. We can remove `DataLoader` and have the same number of queries effectively. For the meantime, we can think about implementing a batch function on a field by field basis though.

src/postgres/inventory/paginator/PgPaginatorOrderingAttributes.ts
calebmer: Ok, let’s just remove the `rows =` then.

src/postgres/inventory/type/PgClassType.ts
calebmer: Theoretically, it could if we had a more complicated underlying data structure from a different database for example. It also allows us to get the correct type in TypeScript as `field.getValue` returns `TFieldValue` whereas `map.get` just returns `mixed` which is not helpful. The Postgres data structure is just a simple map though.

src/postgres/inventory/type/PgClassType.ts
calebmer: At one point the V8 team was exploring supporting a JavaScript [StrongMode](https://docs.google.com/document/d/1Qk0qC4s_XNCLemj42FqfsRLp49nDQMZ1y7fwf5YjaI4/view). The important parts from StrongMode in this case are as follows:

> - An object literal creates a strong object.
> - A TypeError is thrown if a property in an object literal has the same name as an earlier one, unless one is a getter and the other a setter.
> - A TypeError is thrown if an attempt is made, either via the '.', the '[]' syntax, or destructuring, to read a property that does not exist on an object or its prototypes. Destructuring does not throw if a default value is provided.
> - A TypeError is thrown if an attempt is made, either via the '.', the '[]' syntax, to assign to a property that does not exist on an object nor has a setter on its prototype chain. The only exception are assignments to properties of 'this' in a class’s constructor definition (see Classes).

The thesis is that using a JavaScript object as a map for arbitrary key/value pairs is a common source of errors, is bad for performance and engine optimizations, and is inferior to an [ES6 `Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) in every way. Maps have easy to use getters, setters, deletes, existence checks (`has`), iteration, arbitrary keys (so not just strings, but numbers and objects as well), more accurate TypeScript types, and a performant underlying engine implementation. All of these things behave weirdly in JavaScript objects and have a number of corner cases.

So StrongMode got me to consider always using maps for the cases where I’d normally use a JavaScript object (like this case), and after trying it I just think it’s all around superior for cases like this 😊

The objects we pass around in PostGraphQL we have no clue what the keys or values will be statically, so it doesn’t make sense to use JavaScript objects and instead makes sense to use `Map` which is much better for representing dynamic key/value pairs.

calebmer:
> such as determining the required (implicit) columns, such as for relations or `__id`

This is where a type like:

```ts
interface ReadOptimizations {
  dependencies: {
    fields: Array<ObjectType.Field>,
    collectionKeys: Array<CollectionKey>,
    ...
  },
}
```

Could be helpful 😉

The GraphQL code just lists the dependencies it needs, and then Postgres could internally determine what is needed. For example, I’m pretty sure `PgCollectionKey` has the Postgres constraint introspection information.

I’ll start discussing in #342.

src/graphql/schema/collection/createCollectionGqlType.ts
benjie: In #342 we basically just have one SQL query per root-level node; nested function calls are automatically inlined.

src/graphql/schema/connection/createConnectionGqlField.ts
benjie: Totally; suggestions welcome over on #342

src/graphql/schema/node/createNodeFieldEntry.ts
benjie: Yeah, performance is still worse than I'd like... But a lot of these things are still voodoo to me 😳

src/postgres/inventory/paginator/PgPaginatorOrderingAttributes.ts
benjie: Already done on #342 👍

src/postgres/inventory/type/PgClassType.ts
benjie: Awesome, thanks for the illumination 👍

benjie:
Yeah; this is going to need a major refactor before merge - but hopefully it's sufficient proof of concept to justify the effort!

calebmer:
For sure 👍

0x6368656174:
This change is very necessary to add the ability to use GRAND COLUM. Please merge to master, please.

calebmer:
@0x6368656174 This PR is being continued in https://github.com/postgraphql/postgraphql/pull/342. Try pulling down that branch and let us know if that works for you 😊

PR #342, opened 2017-02-04
WIP: Reduce the number of SQL queries

### LINKED ISSUE #219: Performance when embedding children
Doing some performance test for something of mine and along with it postgraphql.
Looking at the queries generated, dataloader is used optimally when embedding a parent relation (1 query), but when embedding children, you execute a separate query for each item from the first level. Because of this the performance is terrible when embedding children for a list that has even 100 items on the first level. I.e the number of queries depends on the number of results returned by the first level
This can (and should) be optimised.
The batching logic for embedding children can be done with a query similar to this
```
select project_id, json_agg(row_to_json(tasks.*)) as tasks from tasks
where project_id in (1,2,3,4)
group by project_id
```

this will give a result like
```
1	[{"id":235581,"name":"name_6623","project_id":1}]
2	[{"id":388001,"name":"name_292081_2","project_id":2}, {"id":405840,"name":"name_681506_6815","project_id":2}, {"id":479398,"name":"name_4139","project_id":2}]
4	[{"id":483932,"name":"name_6566","project_id":4}, {"id":241238,"name":"name_325151_","project_id":4}, {"id":50208,"name":"name_654111_654","project_id":4}]
```

### LINKED ISSUE #265: Performance of computed columns in connections

benjie:
- [x] calculate hasNextPage/hasPreviousPage in the same query as fetching the data
- [x] remove a source of N+1 queries relating to relations (foo.barsByBazId)

TODO:

- [ ] merge master
- [ ] try and remove additional parameters where possible (requires @calebmer's help)
- [ ] look into using CTEs to bypass performance issues, a la #396 

KNOWN ISSUES:

I've been informed by @cusxio that the following query returns null, but that including the `username` field fixes this - sounds like I need to add `username` to the list of implicit fields on the search endpoints:

```graphql
query {
  userByUsername(username:"cusx") {
    id
  }
}
```

Workaround:

```graphql
query {
  userByUsername(username:"cusx") {
    id
    username
  }
}
```

Fixes #219  
Fixes #265

### DIFF
--- a/src/graphql/schema/collection/createCollectionGqlType.ts
+++ b/src/graphql/schema/collection/createCollectionGqlType.ts
@@ -37,7 +37,7 @@ export default function createCollectionGqlType<TValue> (
         primaryKey && [options.nodeIdFieldName, {
           description: 'A globally unique identifier. Can be used in various places throughout the system to identify this single value.',
           type: new GraphQLNonNull(GraphQLID),
-          externalFieldNameDependencies: primaryKey._keyTypeFields.map(([fieldName, field]) => fieldName),
+          externalFieldNameDependencies: primaryKey._keyTypeFields.map(([fieldName, field]) => field.externalFieldName),
           resolve: (value: TValue) => idSerde.serialize(collection, value),
         }],
       ],
@@ -51,7 +51,7 @@ export default function createCollectionGqlType<TValue> (
             value: {
               description: field.description,
               type: gqlType,
-              externalFieldName: fieldName,
+              externalFieldName: field.externalFieldName,
               resolve: (value: TValue): mixed => intoGqlOutput(field.getValue(value)),
             },
           }
--- a/src/graphql/schema/collection/createCollectionGqlType.ts
+++ b/src/graphql/schema/collection/createCollectionGqlType.ts
@@ -7,6 +7,7 @@ import createConnectionGqlField from '../connection/createConnectionGqlField'
 import BuildToken from '../BuildToken'
 import createCollectionRelationTailGqlFieldEntries from './createCollectionRelationTailGqlFieldEntries'
 import getConditionGqlType from './getConditionGqlType'
+import { isSymbol } from 'lodash'
 
 /**
  * Creates the output object type for a collection. This type will include all
@@ -95,9 +96,11 @@ export default function createCollectionGqlType<TValue> (
               ],
               // We use the config when creating a connection field to inject
               // a condition that limits what we select from the paginator.
-              getPaginatorInput: (aliasIdentifier: mixed, args: { condition?: { [key: string]: mixed } }) =>
+              getPaginatorInput: (sourceOrAliasIdentifier: mixed, args: { condition?: { [key: string]: mixed } }) =>
                 conditionHelpers.and(
-                  relation.getTailConditionFromHeadAlias!(aliasIdentifier),
+                  isSymbol(sourceOrAliasIdentifier)
+                    ? relation.getTailConditionFromHeadAlias!(sourceOrAliasIdentifier)
+                    : relation.getTailConditionFromHeadValue!(sourceOrAliasIdentifier),
                   conditionFromGqlInput(args.condition),
                 ),
               subquery: true,
--- a/src/graphql/schema/collection/createCollectionGqlType.ts
+++ b/src/graphql/schema/collection/createCollectionGqlType.ts
@@ -100,7 +100,9 @@ export default function createCollectionGqlType<TValue> (
                   relation.getTailConditionFromHeadAlias!(aliasIdentifier),
                   conditionFromGqlInput(args.condition),
                 ),
-            }, true),
+              subquery: true,
+              relation,
+            }),
           ]
         }),
     ),
--- a/src/graphql/schema/collection/createCollectionGqlType.ts
+++ b/src/graphql/schema/collection/createCollectionGqlType.ts
@@ -52,8 +52,7 @@ export default function createCollectionGqlType<TValue> (
             value: {
               description: field.description,
               type: gqlType,
-              sqlName: () => fieldName,
-              sqlField: fieldName,
+              externalFieldName: fieldName,
               resolve: (value) => {
                 return value.get(fieldName)
               }
--- a/src/graphql/schema/collection/createCollectionGqlType.ts
+++ b/src/graphql/schema/collection/createCollectionGqlType.ts
@@ -37,6 +37,7 @@ export default function createCollectionGqlType<TValue> (
         primaryKey && [options.nodeIdFieldName, {
           description: 'A globally unique identifier. Can be used in various places throughout the system to identify this single value.',
           type: new GraphQLNonNull(GraphQLID),
+          externalFieldNameDependencies: primaryKey._keyTypeFields.map(([fieldName, field]) => fieldName),
           resolve: (value: TValue) => idSerde.serialize(collection, value),
         }],
       ],
--- a/src/graphql/schema/collection/createCollectionGqlType.ts
+++ b/src/graphql/schema/collection/createCollectionGqlType.ts
@@ -99,12 +99,12 @@ export default function createCollectionGqlType<TValue> (
               ],
               // We use the config when creating a connection field to inject
               // a condition that limits what we select from the paginator.
-              getPaginatorInput: (headValue: TValue, args: { condition?: { [key: string]: mixed } }) =>
+              getPaginatorInput: (aliasIdentifier: mixed, args: { condition?: { [key: string]: mixed } }) =>
                 conditionHelpers.and(
-                  relation.getTailConditionFromHeadValue!(headValue),
+                  relation.getTailConditionFromHeadAlias!(aliasIdentifier),
                   conditionFromGqlInput(args.condition),
                 ),
-            }),
+            }, true),
           ]
         }),
     ),
--- a/src/graphql/schema/collection/createCollectionGqlType.ts
+++ b/src/graphql/schema/collection/createCollectionGqlType.ts
@@ -53,7 +53,7 @@ export default function createCollectionGqlType<TValue> (
               description: field.description,
               type: gqlType,
               sqlName: () => fieldName,
-              sqlExpression: (aliasIdentifier) => sql.query`${sql.identifier(aliasIdentifier)}.${sql.identifier(fieldName)}`,
+              sqlField: fieldName,
               resolve: (value) => {
                 return value.get(fieldName)
               }
--- a/src/graphql/schema/collection/createCollectionGqlType.ts
+++ b/src/graphql/schema/collection/createCollectionGqlType.ts
@@ -7,7 +7,6 @@ import createConnectionGqlField from '../connection/createConnectionGqlField'
 import BuildToken from '../BuildToken'
 import createCollectionRelationTailGqlFieldEntries from './createCollectionRelationTailGqlFieldEntries'
 import getConditionGqlType from './getConditionGqlType'
-import { sql } from '../../../postgres/utils'
 
 /**
  * Creates the output object type for a collection. This type will include all
@@ -46,14 +45,13 @@ export default function createCollectionGqlType<TValue> (
       Array.from(type.fields).map(
         <TFieldValue>([fieldName, field]: [string, ObjectType.Field<TValue, TFieldValue>]) => {
           const { gqlType, intoGqlOutput } = getGqlOutputType(buildToken, field.type)
-          const formattedFieldName = formatName.field(fieldName)
           return {
-            key: formattedFieldName,
+            key: formatName.field(fieldName),
             value: {
               description: field.description,
               type: gqlType,
               externalFieldName: fieldName,
-              resolve: (value: TValue): mixed => intoGqlOutput(field.getValue(value),
+              resolve: (value: TValue): mixed => intoGqlOutput(field.getValue(value)),
             },
           }
         },
--- a/src/graphql/schema/collection/createCollectionGqlType.ts
+++ b/src/graphql/schema/collection/createCollectionGqlType.ts
@@ -48,7 +48,7 @@ export default function createCollectionGqlType<TValue> (
           const { gqlType, intoGqlOutput } = getGqlOutputType(buildToken, field.type)
           const formattedFieldName = formatName.field(fieldName)
           return {
-            key: formattedFieldName
+            key: formattedFieldName,
             value: {
               description: field.description,
               type: gqlType,
--- a/src/graphql/schema/collection/createCollectionGqlType.ts
+++ b/src/graphql/schema/collection/createCollectionGqlType.ts
@@ -53,9 +53,7 @@ export default function createCollectionGqlType<TValue> (
               description: field.description,
               type: gqlType,
               externalFieldName: fieldName,
-              resolve: (value) => {
-                return value.get(fieldName)
-              }
+              resolve: (value: TValue): mixed => intoGqlOutput(field.getValue(value),
             },
           }
         },
--- a/src/graphql/schema/collection/createCollectionRelationTailGqlFieldEntries.ts
+++ b/src/graphql/schema/collection/createCollectionRelationTailGqlFieldEntries.ts
@@ -46,7 +46,7 @@ export default function createCollectionRelationTailGqlFieldEntries <TSource, TV
             : formatName.field(`${headCollection.type.name}-by-${relation.name}`),
           {
             description: `Reads a single ${scrib.type(headCollectionGqlType)} that is related to this ${scrib.type(collectionGqlType)}.`,
-            externalFieldNameDependencies: relation._tailFieldNames,
+            externalFieldNameDependencies: relation._tailFieldNames, // 🔥 Needs to account for classicIds
             type: headCollectionGqlType,
             async resolve (source: TSource, _args: {}, context: mixed, resolveInfo: mixed): Promise<mixed> {
               const value = options.getCollectionValue(source)
--- a/src/graphql/schema/collection/createCollectionRelationTailGqlFieldEntries.ts
+++ b/src/graphql/schema/collection/createCollectionRelationTailGqlFieldEntries.ts
@@ -46,6 +46,7 @@ export default function createCollectionRelationTailGqlFieldEntries <TSource, TV
             : formatName.field(`${headCollection.type.name}-by-${relation.name}`),
           {
             description: `Reads a single ${scrib.type(headCollectionGqlType)} that is related to this ${scrib.type(collectionGqlType)}.`,
+            externalFieldNameDependencies: relation._tailFieldNames,
             type: headCollectionGqlType,
             async resolve (source: TSource, _args: {}, context: mixed, resolveInfo: mixed): Promise<mixed> {
               const value = options.getCollectionValue(source)
--- a/src/postgraphql/schema/procedures/createPgProcedureQueryGqlFieldEntry.ts
+++ b/src/postgraphql/schema/procedures/createPgProcedureQueryGqlFieldEntry.ts
@@ -8,6 +8,7 @@ import pgClientFromContext from '../../../postgres/inventory/pgClientFromContext
 import createPgProcedureFixtures from './createPgProcedureFixtures'
 import createPgProcedureSqlCall from './createPgProcedureSqlCall'
 import PgProcedurePaginator from './PgProcedurePaginator'
+import getSelectFragment from '../../../postgres/inventory/paginator/getSelectFragment'
 
 /**
  * Creates the fields for query procedures. Query procedures that return
@@ -51,10 +52,14 @@ function createPgSingleProcedureQueryGqlFieldEntry (
     type: fixtures.return.gqlType,
     args: buildObject(argEntries),
 
-    async resolve (_source, args, context): Promise<mixed> {
+    async resolve (_source, args, context, resolveInfo): Promise<mixed> {
       const client = pgClientFromContext(context)
       const input = argEntries.map(([argName], i) => fixtures.args[i].fromGqlInput(args[argName]))
-      const query = sql.compile(sql.query`select to_json(${createPgProcedureSqlCall(fixtures, input)}) as value`)
+      const aliasIdentifier = Symbol()
+      const query = sql.compile(sql.query`
+        select ${getSelectFragment(resolveInfo, aliasIdentifier, fixtures.return.gqlType)} as value
+        from ${createPgProcedureSqlCall(fixtures, input)} as ${sql.identifier(aliasIdentifier)}
+      `)
       const { rows: [row] } = await client.query(query)
       return row ? fixtures.return.intoGqlOutput(fixtures.return.type.transformPgValueIntoValue(row['value'])) : null
     },
--- a/src/postgraphql/schema/procedures/createPgProcedureQueryGqlFieldEntry.ts
+++ b/src/postgraphql/schema/procedures/createPgProcedureQueryGqlFieldEntry.ts
@@ -57,11 +57,12 @@ function createPgSingleProcedureQueryGqlFieldEntry (
       const input = argEntries.map(([argName], i) => fixtures.args[i].fromGqlInput(args[argName]))
       const aliasIdentifier = Symbol()
       const query = sql.compile(sql.query`
-        select ${getSelectFragment(resolveInfo, aliasIdentifier, fixtures.return.gqlType)} as value
+        select ${getSelectFragment(resolveInfo, aliasIdentifier, fixtures.return.gqlType)} as value,
+        (${sql.identifier(aliasIdentifier)} is null) as is_null
         from ${createPgProcedureSqlCall(fixtures, input)} as ${sql.identifier(aliasIdentifier)}
       `)
       const { rows: [row] } = await client.query(query)
-      return row ? fixtures.return.intoGqlOutput(fixtures.return.type.transformPgValueIntoValue(row['value'])) : null
+      return row && !row['is_null'] ? fixtures.return.intoGqlOutput(fixtures.return.type.transformPgValueIntoValue(row['value'])) : null
     },
   }]
 }
--- a/src/postgres/inventory/collection/PgRelation.ts
+++ b/src/postgres/inventory/collection/PgRelation.ts
@@ -3,6 +3,7 @@ import { PgCatalog, PgCatalogAttribute, PgCatalogForeignKeyConstraint } from '..
 import PgClassType from '../type/PgClassType'
 import PgCollection from './PgCollection'
 import PgCollectionKey from './PgCollectionKey'
+import { sql } from '../../utils'
 
 // TODO: This implementation is sketchy. Implement it better!
 // TODO: Tests
@@ -42,6 +43,12 @@ class PgRelation implements Relation<PgClassType.Value, PgCollectionKey.Value, P
   public getTailConditionFromHeadValue (headValue: PgClassType.Value): Condition {
     return conditionHelpers.and(...this._headFieldNames.map((headFieldName, i) =>
       conditionHelpers.fieldEquals(this._tailFieldNames[i], headValue.get(headFieldName)),
+    ))
+  }
+
+  public getTailConditionFromHeadAlias (aliasIdentifier): Condition {
+    return conditionHelpers.and(...this._headFieldNames.map((headFieldName, i) =>
+      conditionHelpers.fieldEqualsQuery(this._tailFieldNames[i], sql.query`${sql.identifier(aliasIdentifier)}.${sql.identifier(headFieldName)}`),
     ))
   }
 }
--- a/src/postgres/utils/sql.ts
+++ b/src/postgres/utils/sql.ts
@@ -1,5 +1,6 @@
 import { QueryConfig } from 'pg'
 import minify = require('pg-minify')
+import Client = require('pg/lib/client')
 
 namespace sql {
   type NestedArray<T> = Array<T> | Array<Array<T>> | Array<Array<Array<T>>> | Array<Array<Array<Array<T>>>>
@@ -10,6 +11,7 @@ namespace sql {
    */
   export type SqlItem =
     { type: 'RAW', text: string } |
+    { type: 'LITERAL', text: string } |
     { type: 'IDENTIFIER', names: Array<string | symbol> } |
     { type: 'VALUE', value: mixed }
 
@@ -58,6 +60,9 @@ namespace sql {
       switch (item.type) {
         case 'RAW':
           text += item.text
+          break
+        case 'LITERAL':
+          text += escapeSqlLiteral(item.text)
           break
         // TODO: Identifier escaping? I know `pg-promise` does some of this.
         // Most of the time identifiers are trusted, but this could be a source
@@ -111,6 +116,12 @@ namespace sql {
   export const raw = (text: string): SqlItem => ({ type: 'RAW', text })
 
   /**
+   * Creates a Sql item for a string that will be escaped so it is safe against
+   * Sql injection
+   */
+  export const literal = (text: string): SqlItem => ({ type: 'LITERAL', text })
+
+  /**
    * Creates a Sql item for a Sql identifier. A Sql identifier is anything like
    * a table, schema, or column name. An identifier may also have a namespace,
    * thus why many names are accepted.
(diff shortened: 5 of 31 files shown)


benjie:
I've just ran this against a large and relatively complex query for my own project;

master: 380ms, 435 SQL queries
this branch: 280ms, 9 SQL queries

That's with a local database (16GB RAM, quad core i7, virtually zero latency - ping says 0.06ms) - even if we factor in the network latency to the database to be just 1 millisecond we're looking at a significant speedup! (I just tested my servers and latency when unloaded comes in at an average about 1.2ms.)

calebmer:
Those numbers are pretty sweet 👍

Maybe wait on merging into https://github.com/calebmer/postgraphql/pull/340? I’d like to get a chance to look at it first 😊

benjie:
@calebmer Actually I think I'm going to merge this and #340 together - I was planning a two-stage thing (hence starting work on this separately) but as I progress I realise how much they are inter-dependent and not incorporating this into #340 may paint us into a corner preventing these advances from being implemented.

src/graphql/schema/collection/createCollectionGqlType.ts
```diff
@@ -53,7 +53,7 @@ export default function createCollectionGqlType<TValue> (
               description: field.description,
```
benjie: Huh, good catch. It seems the TypeScript compiler (at least the version I'm using) is hella forgiving. It's also really REALLY hard to figure out where syntax has gone wrong with TypeScript as compared to Babel which is so much more helpful with its errors.

valorize:
@benjie is in the flow :D Anything I can test or help you with?

benjie:
@valorize You're welcome to test it again; I'm currently aware that procedures returning sets are broken but most other things are working. I came up with a complex query for my dataset and then wrote the output of it on master to a JSON file, repeated on this branch and then diffed them. The aim is that the outputs are byte-for-byte identical.

benjie:
I think all the read operations work (unless I just broke something...) and mutations trigger but they don't return data yet. I don't anticipate fixing that taking very long; but I'm done for the day. Hopefully tomorrow I'll get this into a state ready to use.

benjie:
Most everything is working now; there's just one thing I need to fix which is reflective references - where a collection relates to itself; i.e.

```
Person {
  id
  name
  mother
  father
}


query {
  allPeople {
    nodes {
      name
      mother {
         name
      }
    }
  }
}
```

On it now...

benjie:
Sorted 👍

benjie:
Ah, there's an issue with procedures that return custom types... I gotta dash out but will fix later 👍

guigrpa:
@benjie I ❤️ the commit messages (I had to share this) 😂

valorize:
Hi Benji, I have a problem with authentication:
```
[{
"message": "Cannot read property 'selections' of null",
"locations": [{"line": 2, "column": 3}],
"path": ["authenticate"],
"stack": ["TypeError: Cannot read property 'selections' of null",
	"    at parseASTIntoFields (src/postgres/inventory/paginator/getSelectFragment.js:116:26)",
	"    at src/postgres/inventory/paginator/getSelectFragment.js:146:24", "    at Array.forEach (native)",
	"    at processFragment (src/postgres/inventory/paginator/getSelectFragment.js:145:42)",
	"    at parseASTIntoFields (src/postgres/inventory/paginator/getSelectFragment.js:133:5)",
	"    at src/postgres/inventory/paginator/getSelectFragment.js:118:9", "    at Array.forEach (native)",
	"    at parseASTIntoFields (src/postgres/inventory/paginator/getSelectFragment.js:116:38)",
	"    at src/postgres/inventory/paginator/getSelectFragment.js:15:7", "    at Array.forEach (native)"]
}]
```

benjie:
Okay I've done some more major fixes.

<details>
<summary>We now pass the integration tests, except for one.</summary>

```
$ npm run test -- src/postgraphql/__tests__/*-test.*
 PASS  src/postgraphql/__tests__/postgraphqlIntegrationSchema-test.js
 FAIL  src/postgraphql/__tests__/postgraphqlIntegrationOperations-test.js (13.064s)
  ● operation procedure-query.graphql

    - Snapshot
    + Received

    @@ -9,42 +9,11 @@
           "b": "easy cheesy baked potatoes",
           "c": "RED",
           "d": null,
           "fooBar": 8,
         },
    -    "intSetQuery": Object {
    -      "edges": Array [
    -        Object {
    -          "cursor": "WyJuYXR1cmFsIiwxXQ==",
    -          "node": 1,
    -        },
    -        Object {
    -          "cursor": "WyJuYXR1cmFsIiwyXQ==",
    -          "node": 2,
    -        },
    -        Object {
    -          "cursor": "WyJuYXR1cmFsIiwzXQ==",
    -          "node": 3,
    -        },
    -        Object {
    -          "cursor": "WyJuYXR1cmFsIiw0XQ==",
    -          "node": 4,
    -        },
    -        Object {
    -          "cursor": "WyJuYXR1cmFsIiw1XQ==",
    -          "node": 5,
    -        },
    -        Object {
    -          "cursor": "WyJuYXR1cmFsIiw2XQ==",
    -          "node": null,
    -        },
    -        Object {
    -          "cursor": "WyJuYXR1cmFsIiw3XQ==",
    -          "node": 6,
    -        },
    -      ],
    -    },
    +    "intSetQuery": null,
         "noArgsQuery": 2,
         "tableQuery": Object {
           "__id": "WyJwb3N0cyIsNV0=",
           "authorId": 5,
           "headline": "Please, Don-Bot… look into your hard drive, and open your mercy file!",
    @@ -84,6 +53,9 @@
             },
           ],
         },
         "typesQuery": false,
       },
    +  "errors": Array [
    +    [GraphQLError: Expected integer. Not 'object'.],
    +  ],
     }

      at Object.<anonymous> (src/postgraphql/__tests__/postgraphqlIntegrationOperations-test.js:50:36)
      at step (node_modules/tslib/tslib.js:114:27)
      at Object.next (node_modules/tslib/tslib.js:95:57)
      at fulfilled (node_modules/tslib/tslib.js:85:62)
      at process._tickCallback (internal/process/next_tick.js:103:7)

 PASS  src/postgraphql/__tests__/postgraphql-test.js
 PASS  src/postgraphql/__tests__/withPostGraphQLContext-test.js

Snapshot Summary
 › 1 snapshot test failed in 1 test suite. Inspect your code changes or run with `npm test -- -u` to update them.

Test Suites: 1 failed, 3 passed, 4 total
Tests:       1 failed, 42 passed, 43 total
Snapshots:   1 failed, 18 passed, 19 total
Time:        15.544s
Ran all test suites matching /src/postgraphql/__tests__/postgraphql-test.js|src/postgraphql/__tests__/postgraphqlIntegrationOperations-test.js|src/postgraphql/__tests__/postgraphqlIntegrationSchema-test.js|src/postgraphql/__tests__/withPostGraphQLContext-test.js/.
```
</details>

The one that's failing is the `intSetQuery()` custom procedure which returns a set of int

https://github.com/calebmer/postgraphql/blob/39cff1d123c282667e339064a9a5be9538e330a6/src/postgraphql/__tests__/fixtures/queries/procedure-query.graphql#L10
https://github.com/calebmer/postgraphql/blob/39cff1d123c282667e339064a9a5be9538e330a6/examples/kitchen-sink/schema.sql#L163

I've dug around a bunch but I can't figure out how to make specifically this one case return the node directly; at the moment it's getting returned a value that's like `{"__local_1__": 1}` rather than simply `1`. This should be solveable by putting the relevant `sqlExpression` in the right place or merely hacking the resolver to return the one and only property of the object; but I can't quite figure it out.

I also guarantee NOTHING when it comes to `classicIds`; the rewriting of fieldName messes with my assumptions and I'll need @calebmer's help figuring out how to correctly account for this. I've made a start, enough to pass the tests above, but I'm not confident in it.

I'm going to be testing this tomorrow on my own stuff, if you want to give it a go you can `npm install calebmer/postgraphql#ignore/bg2` but beware that branch changes frequently and is constantly rebased so DO NOT RELY ON THE COMMIT HASHES!

I've also noticed that because computed columns on rows you update via an SQL function use the value of the row BEFORE the update rather than after. E.g. this would be a problem:

```
mutation {
  favoritePost(id: $id) {
    post {
      isFavorite
    }
  }
}
```

benjie:
Unfortunately I've not managed to work around the SQL function out-of-sync issue with `WITH` clauses as I had hoped, because procedures that return simple values when wrapped by `WITH` clauses become structured.

```diff
diff --git i/src/postgraphql/schema/procedures/createPgProcedureMutationGqlFieldEntry.ts w/src/postgraphql/schema/procedures/createPgProcedureMutationGqlFieldEntry.ts
index bd39733..6d4cdb6 100644
--- i/src/postgraphql/schema/procedures/createPgProcedureMutationGqlFieldEntry.ts
+++ w/src/postgraphql/schema/procedures/createPgProcedureMutationGqlFieldEntry.ts
@@ -111,8 +111,9 @@ export default function createPgProcedureMutationGqlFieldEntry (
 
       const query = sql.compile(
         sql.query`
+          with ${sql.identifier(aliasIdentifier)} as (select value from ${procedureCall} as value)
           select ${getSelectFragment(resolveInfo, aliasIdentifier, fixtures.return.gqlType)} as value
-          from ${procedureCall} as ${sql.identifier(aliasIdentifier)}
+          from ${sql.identifier(aliasIdentifier)}
         `
       )
 
diff --git i/src/postgraphql/schema/procedures/createPgProcedureQueryGqlFieldEntry.ts w/src/postgraphql/schema/procedures/createPgProcedureQueryGqlFieldEntry.ts
index d575b30..c026b5a 100644
--- i/src/postgraphql/schema/procedures/createPgProcedureQueryGqlFieldEntry.ts
+++ w/src/postgraphql/schema/procedures/createPgProcedureQueryGqlFieldEntry.ts
@@ -57,9 +57,10 @@ function createPgSingleProcedureQueryGqlFieldEntry (
       const input = argEntries.map(([argName], i) => fixtures.args[i].fromGqlInput(args[argName]))
       const aliasIdentifier = Symbol()
       const query = sql.compile(sql.query`
+        with ${sql.identifier(aliasIdentifier)} as (select value from ${createPgProcedureSqlCall(fixtures, input)} as value)
         select ${getSelectFragment(resolveInfo, aliasIdentifier, fixtures.return.gqlType)} as value,
         (${sql.identifier(aliasIdentifier)} is null) as is_null
-        from ${createPgProcedureSqlCall(fixtures, input)} as ${sql.identifier(aliasIdentifier)}
+        from ${sql.identifier(aliasIdentifier)}
       `)
       const { rows: [row] } = await client.query(query)
       return row && !row['is_null'] ? fixtures.return.intoGqlOutput(fixtures.return.type.transformPgValueIntoValue(row['value'])) : null
```

```
    - Snapshot
    + Received

[...]
    -    "b": false,
    -    "c": 42,
    -    "d": 3.1415,
    -    "e": "hello, world!",
    +    "b": Object {
    +      "value": false,
    +    },
    +    "c": Object {
    +      "value": 42,
    +    },
    +    "d": Object {
    +      "value": 3.1415,
    +    },
    +    "e": Object {
    +      "value": "hello, world!",
    +    },
```

benjie:
I've reverted mutations to use the old method because of the issues above; have added notes to the source to explain why.

@calebmer The only thing preventing this being ready now, I think, is `intSetQuery` - could you give me a hand figuring that out?

benjie:
@valorize I've been running the latest on my staging environment all day and it seems to run nicely; so if you get a chance to give it a go and see if you have any issues that'd be another useful datapoint 👍

calebmer:
The progress is looking good! I’m going to do another review of the code and look into ways to fix the `intSetQuery` test, but first I have a couple of questions:

1. If in the future we allow users to provide custom types and/or extend the types we generate from Postgres, how would this setup affect them? Would the user be required to hook into the single query generation system?
2. Are there ways for us to “bail out” when we haven’t yet implemented query flattening. For example (and maybe you’ve already addressed this 😊), computed connection procedures like `person_latest_posts`. Is there a way to gracefully fallback to an N+1 mode in an instance like that? (Again, I’m assuming you haven’t already addressed that case. If you have, then awesome! But perhaps in the future, we may add a feature that requires a graceful fallback. Would it be possible to fallback in that time?)

src/graphql/schema/collection/createCollectionGqlType.ts
```diff
@@ -37,6 +37,7 @@ export default function createCollectionGqlType<TValue> (
         primaryKey && [options.nodeIdFieldName, {
           description: 'A globally unique identifier. Can be used in various places throughout the system to identify this single value.',
           type: new GraphQLNonNull(GraphQLID),
+          externalFieldNameDependencies: primaryKey._keyTypeFields.map(([fieldName, field]) => field.externalFieldName),
```
calebmer: `primaryKey._keyTypeFields` depends on an internal implementation detail, right? Any reason why we couldn’t just do: `externalKeyDependencies: [primaryKey]`. This would also work for some relations.

src/graphql/schema/collection/createCollectionGqlType.ts
```diff
@@ -50,6 +51,7 @@ export default function createCollectionGqlType<TValue> (
             value: {
               description: field.description,
               type: gqlType,
+              externalFieldName: field.externalFieldName,
```
calebmer: Any reason we can’t have just `externalField: field` here?

src/postgres/utils/sql.ts
```diff
@@ -153,22 +164,20 @@ namespace sql {
...
   function escapeSqlIdentifier (str: string): string {
-    let escaped = '"'
-
-    for (const c of str) {
-      if (c === '"') escaped += c + c
-      else escaped += c
-    }
-
-    escaped += '"'
+    return Client.prototype.escapeIdentifier(str)
```
calebmer: Nice.

src/postgraphql/schema/procedures/createPgProcedureQueryGqlFieldEntry.ts
```diff
@@ -51,12 +52,17 @@ function createPgSingleProcedureQueryGqlFieldEntry (
...
+        select ${getSelectFragment(resolveInfo, aliasIdentifier, fixtures.return.gqlType)} as value,
+        (${sql.identifier(aliasIdentifier)} is null) as is_null
+        from ${createPgProcedureSqlCall(fixtures, input)} as ${sql.identifier(aliasIdentifier)}
+      `)
       const { rows: [row] } = await client.query(query)
-      return row ? fixtures.return.intoGqlOutput(fixtures.return.type.transformPgValueIntoValue(row['value'])) : null
+      return row && !row['is_null'] ? fixtures.return.intoGqlOutput(fixtures.return.type.transformPgValueIntoValue(row['value'])) : null
     },
   }]
 }
```
calebmer: The function below, `createPgSetProcedureQueryGqlFieldEntry`, is probably where the error for `intSetQuery` is coming from. What does the query that runs for `intSetQuery` look like?

calebmer:
@benjie what do you think about adding to the operations integration test a query tracker snapshot? So whenever `client.query` gets run we take a snapshot of the query text and variables. This could be helpful to track changes and regressions.

If you could implement that on `master` and then merge that into this branch we would even get a diff, which would be awesome to see! My hunch is that this PR will make the tests faster 😉

src/graphql/schema/collection/createCollectionGqlType.ts
benjie: I didn't know where to grab this from, so just grabbed the nearest available thing. Also what if it's a compound primary key?

src/graphql/schema/collection/createCollectionGqlType.ts
benjie: I don't see why not.

src/postgres/utils/sql.ts
benjie: I'm not sure; it could be broken in a point release. I'm hoping if it is then the `require()` will fail making it very obvious though.

src/postgraphql/schema/procedures/createPgProcedureQueryGqlFieldEntry.ts
benjie: This:
https://github.com/calebmer/postgraphql/blob/333f513eeceb7c989509eee014009938aafd3a46/src/postgraphql/schema/procedures/createPgProcedureQueryGqlFieldEntry.ts#L94

calls into this:
https://github.com/calebmer/postgraphql/blob/333f513eeceb7c989509eee014009938aafd3a46/src/graphql/schema/connection/createConnectionGqlField.ts#L131

which calls into this:
https://github.com/calebmer/postgraphql/blob/333f513eeceb7c989509eee014009938aafd3a46/src/postgres/inventory/paginator/PgPaginatorOrderingOffset.ts#L133-L153

(or this:
https://github.com/calebmer/postgraphql/blob/333f513eeceb7c989509eee014009938aafd3a46/src/postgres/inventory/paginator/PgPaginatorOrderingAttributes.ts#L75-L89
)

which calls this:
https://github.com/calebmer/postgraphql/blob/333f513eeceb7c989509eee014009938aafd3a46/src/postgraphql/schema/procedures/PgProcedurePaginator.ts#L54

and finally:
https://github.com/calebmer/postgraphql/blob/333f513eeceb7c989509eee014009938aafd3a46/src/postgraphql/schema/procedures/createPgProcedureSqlCall.ts#L42


Man this stuff is complex.

benjie: Basically I've not found a way to work around the `with` query forcing a column alias. I think in the event of returning scalars (which we know up front, right?) we can just return the only value in a single-keyed object?

benjie:
As it happens I already exported the pgClient debugger so I could see the statements that the SQL runs; yes this totally makes sense. I think we should also build a *LOT* more integration tests (including way more complex ones) to ensure that the output still matches perfectly.

I wonder if someone else has time to do this? I went way WAY over time budget on this and can't afford to spend any more time on it right now.

(Also, quite a lot of time seems to be sent in JavaScript-land (rather than SQL-land) which means there's still some non-performant code around that could do with optimisation.)

benjie:
> If in the future we allow users to provide custom types and/or extend the types we generate from Postgres, how would this setup affect them? Would the user be required to hook into the single query generation system?

Via the `intoGqlOutput`? That's still called so should run in the same way.

> Are there ways for us to “bail out” when we haven’t yet implemented query flattening. For example (and maybe you’ve already addressed this 😊), computed connection procedures like person_latest_posts. Is there a way to gracefully fallback to a N+1 mode in an instance like that? Again, I’m assuming you haven’t already addressed that case. If you have, then awesome! But perhaps in the future, we may add a feature that requires a graceful fallback. Would it be possible to fallback in that time?

Sure, if you don't specify a `externalFieldName` or `sqlExpression` then we won't attempt to fetch it up front. The resolve method will always run; so if we don't fetch the data up front it can implement the fetch logic instead.

In fact, sometimes we have to hedge (e.g. for computed columns on mutations) - because of maps we can just say "if the result has this field already, use that, otherwise fetch it ourself" like we do here:

https://github.com/calebmer/postgraphql/pull/342/files#diff-68e6e47f66f4641b59bfbd00beff36acR68
```js
if (source.has(attrName)) {
  const value = source.get(attrName)
  return value != null ? fixtures.return.intoGqlOutput(value) : null
} else {
  // Subquery failed (e.g. computed function as subfield of compound type); fall back to old logic
  const client = pgClientFromContext(context)
  const input = [source, ...argEntries.map(([argName], i) => fixtures.args[i + 1].fromGqlInput(args[argName]))]
  const query = sql.compile(sql.query`select to_json(${createPgProcedureSqlCall(fixtures, input)}) as value`)
  const { rows: [row] } = await client.query(query)
  return row ? fixtures.return.intoGqlOutput(row['value']) : null
}
```

valorize:
Hi @benjie thanks for your efforts. I'm having problems regarding the classic ids (column __local_4__.row_id does not exist) and will try to nail it. But first did you also have similar outputs from npm test?

```
Test Suites: 4 failed, 21 passed, 25 total
Tests:       168 failed, 296 passed, 464 total
Snapshots:   1 failed, 19 passed, 20 total
Time:        91.217s
Ran all test suites.
```

benjie:
No - the classic IDs integration tests are not strong enough (and they're the only tests I run because I've basically broken all the other tests!). However I'm aware that classic IDs aren't fully supported because we rename the column to row_id internally, and I don't always handle that - you can see in the source where I've added the comment `// 🔥 Needs to account for classicIds` - that's where I'm using the re-written column names which are identical for normal mode, but `id` has mapped to `row_id` in classic IDs mode. There may be a couple places over and above this that I've missed too. I'm hoping someone can get the original column names through to the relevant places rather than using the hackish sources I'm using.

src/postgraphql/schema/procedures/createPgProcedureQueryGqlFieldEntry.ts
calebmer: Yep. That would work.

src/graphql/schema/collection/createCollectionRelationTailGqlFieldEntries.ts
```diff
@@ -46,11 +46,12 @@ export default function createCollectionRelationTailGqlFieldEntries <TSource, TV
             : formatName.field(`${headCollection.type.name}-by-${relation.name}`),
           {
             description: `Reads a single ${scrib.type(headCollectionGqlType)} that is related to this ${scrib.type(collectionGqlType)}.`,
+            externalFieldNameDependencies: relation._tailFieldNames, // 🔥 Needs to account for classicIds
```
valorize: Why do we need to account here for classicIds? The source of a relation is a column != id , it does not seem to address the destination of the relation here.

src/graphql/schema/collection/createCollectionGqlType.ts
valorize: external describes the database side or is external the graphql side?

src/graphql/schema/collection/createCollectionRelationTailGqlFieldEntries.ts
benjie: It's possible that your primary key is called "flibbledeedee" and that your "id" column is merely a foreign key.

src/graphql/schema/collection/createCollectionGqlType.ts
```diff
@@ -93,11 +95,13 @@ export default function createCollectionGqlType<TValue> (
               ],
               // We use the config when creating a connection field to inject
               // a condition that limits what we select from the paginator.
-              getPaginatorInput: (headValue: TValue, args: { condition?: { [key: string]: mixed } }) =>
+              getPaginatorInput: (aliasIdentifier: mixed, args: { condition?: { [key: string]: mixed } }) =>
                 conditionHelpers.and(
-                  relation.getTailConditionFromHeadValue!(headValue),
+                  relation.getTailConditionFromHeadAlias!(aliasIdentifier),
```
valorize: I think we have to handle classicIds here as well.

valorize: I think that PgRelation::getTailConditionFromHeadAlias does not convert the row_id of _headFieldNames to id?

benjie: The place to handle it would be here: https://github.com/calebmer/postgraphql/pull/342/files#diff-44b7c98267d37271bb1a324470ae1767R50 - but you're correct - it needs handling

src/postgres/inventory/collection/PgRelation.ts
```diff
@@ -44,6 +45,12 @@ class PgRelation implements Relation<PgClassType.Value, PgCollectionKey.Value, P
       conditionHelpers.fieldEquals(this._tailFieldNames[i], headValue.get(headFieldName)),
     ))
   }
+
+  public getTailConditionFromHeadAlias (aliasIdentifier): Condition {
+    return conditionHelpers.and(...this._headFieldNames.map((headFieldName, i) =>
```
benjie: Needs to handle classicIds

benjie:
⚠️  🔥  Just discovered a (very easy to work around) issue in this:

If you request a field using a unique constraint (not the primary key) and you don't REQUEST that same field then you won't get a result. The solution is simple: we just need to fetch the column that we're unique on.

Fail:
```
postBySlug(slug: 'dance-party') {
  id
}
```

Success:
```
postBySlug(slug: 'dance-party') {
  id
  slug
}
```

valorize:
In getSelectFragement.js (why is it js and not ts?) I completely commented out 71-84. After this change classicIds work for my test query. What is the conceptual difference between field.externalFieldName and field.externalFieldNameDependencies?
```
if (field.externalFieldNameDependencies) {
       field.externalFieldNameDependencies.forEach(
          externalFieldName => {
          	if(externalFieldName) {
	            fields[externalFieldName] = {
	              type: QUERY,
	              query: sql.query`${sql.identifier(aliasIdentifier)}.${sql.identifier(externalFieldName)}`,
	            }

            }
          }
        )
      }
```

valorize:
I was to eager. Throwing out the externalFieldNameDependencies ended in not fetching the sub sub queries. Ok so we rather account for classicIds with
```
query: sql.query`${sql.identifier(aliasIdentifier)}.${sql.identifier(externalFieldName === 'row_id' ? 'id' : externalFieldName)}`,```

But thats just a hack. @benjie do you have a better idea?

**Comment by valorize** [comment]
Subjective Evaluation: Request time cut in half 💛

**Comment by benjie** [comment]
`externalFieldName` is the name of the postgresql column (without id -> rowId shenanigans) this field represents

`externalFieldNameDependencies` is a list of external field names that we need to fetch in order for the system to work (because if we don't fetch them then we cannot generate the required field - e.g. the `__id` computed field (which is still __id because I've not merged the latest master yet), or relations that need to know the row's `fooId` so it can look up `barsByFooId`, and so on). When sorting we also have to add the sort criteria to the list of fields to fetch otherwise the logic in the paginators falls down I think.

**Comment by benjie** [comment]
I'm not sure how best to deal with rowId and I'm sure I'm getting various things from places I oughtn't since I just hacked it out as a proof of concept, so I'm awaiting Caleb's ideas on this point.

**Comment by benjie** [comment]
Just discovered a bug in this relating to one-to-many relations on sub-entries on mutation payloads.

**Comment by benjie** [comment]
Found it 👍

**Comment by 0x6368656174** [comment]
In this version, the GRAND KOLUMN does not work, in contrast to the version of bg / performance-experiment. And I really need the support of GRAND COLUMN = (((

**Comment by benjie** [comment]
@0x6368656174 Do you know why it doesn't work on this branch? Given it's based on the other (which didn't pass tests), I'm theorising it's because this branch pulls in the other required columns that are necessary for PostGraphQL internals, e.g. if you order by a field, that field gets implicitly added to the selection; similarly if you filter by a field.

**Comment by 0x6368656174** [comment]
@benjie if I execute the query:
````graphql
{
  allUsers {
    nodes {
      id
      name
    }
  }
}
````

In this branch, the query is selected:

````sql
select coalesce(json_agg(__local_0__), '[]'::json) as "rows", ( select count(*) from ( select * from "public"."users" as __local_1__ where "id" is not null and true ) as __local_2__ ) ::integer as "totalCount", ( select count(*) from ( select * from "public"."users" as __local_1__ where "id" is not null and true ) as __local_3__ where true ) > count(*) as "hasNextPage", false as "hasPreviousPage" from ( select json_build_object( 'value', json_build_object('id', __local_4__."id", 'name', __local_4__."name"), 'cursor', json_build_array("id") ) as __local_0__ from ( select * from "public"."users" as __local_1__ where "id" is not null and true ) as __local_4__ where true and true order by "id" using < limit all ) as __local_5__
````

And it should be:
````sql
select coalesce(json_agg(__local_0__), '[]'::json) as "rows", ( select count(*) from ( select "id", "name" from "public"."users" as __local_1__ where "id" is not null and true ) as __local_2__ ) ::integer as "totalCount", ( select count(*) from ( select "id", "name" from "public"."users" as __local_1__ where "id" is not null and true ) as __local_3__ where true ) > count(*) as "hasNextPage", false as "hasPreviousPage" from ( select json_build_object( 'value', json_build_object('id', __local_4__."id", 'name', __local_4__."name"), 'cursor', json_build_array("id") ) as __local_0__ from ( select "id", "name" from "public"."users" as __local_1__ where "id" is not null and true ) as __local_4__ where true and true order by "id" using < limit all ) as __local_5__;
````

Plus, I do not need `totalCount`, `hasNextPage`, `hasPreviosPage`, but in the SQL-request they are writing.

**Comment by benjie** [comment]
Formatted, so I can compare:

```sql
SELECT coalesce(json_agg(__local_0__), '[]'::json) AS "rows",

  (SELECT count(*)
   FROM
     (SELECT *
      FROM "public"."users" AS __local_1__
      WHERE "id" IS NOT NULL
        AND TRUE ) AS __local_2__) ::integer AS "totalCount",

  (SELECT count(*)
   FROM
     (SELECT *
      FROM "public"."users" AS __local_1__
      WHERE "id" IS NOT NULL
        AND TRUE ) AS __local_3__
   WHERE TRUE ) > count(*) AS "hasNextPage",
       FALSE AS "hasPreviousPage"
FROM
  (SELECT json_build_object('value', json_build_object('id', __local_4__."id", 'name', __local_4__."name"), 'cursor', json_build_array("id")) AS __local_0__
   FROM
     (SELECT *
      FROM "public"."users" AS __local_1__
      WHERE "id" IS NOT NULL
        AND TRUE ) AS __local_4__
   WHERE TRUE
     AND TRUE
   ORDER BY "id" USING <
   LIMIT ALL) AS __local_5__
```

```sql
SELECT coalesce(json_agg(__local_0__), '[]'::json) AS "rows",

  (SELECT count(*)
   FROM
     (SELECT "id",
             "name"
      FROM "public"."users" AS __local_1__
      WHERE "id" IS NOT NULL
        AND TRUE ) AS __local_2__) ::integer AS "totalCount",

  (SELECT count(*)
   FROM
     (SELECT "id",
             "name"
      FROM "public"."users" AS __local_1__
      WHERE "id" IS NOT NULL
        AND TRUE ) AS __local_3__
   WHERE TRUE ) > count(*) AS "hasNextPage",
       FALSE AS "hasPreviousPage"
FROM
  (SELECT json_build_object('value', json_build_object('id', __local_4__."id", 'name', __local_4__."name"), 'cursor', json_build_array("id")) AS __local_0__
   FROM
     (SELECT "id",
             "name"
      FROM "public"."users" AS __local_1__
      WHERE "id" IS NOT NULL
        AND TRUE ) AS __local_4__
   WHERE TRUE
     AND TRUE
   ORDER BY "id" USING <
   LIMIT ALL) AS __local_5__;
```

benjie:
@0x6368656174 So you're saying that the `select *` in the count statements and the source table are causing you trouble.

I'm aware that we over-fetch the hasPreviousPage/etc - I ran out of time whilst writing this proof of concept.

The intention of this PR was performance, rather than excluding fields for permissions reasons, and I'm hesitant to advance the code any further knowing that we'll probably be rebuilding a fair chunk of it. That said, you're welcome to build your own branch off of this one that solves your issue - if it's simple enough then I may consider merging, but you'll have to be careful to include all required fields, not just the ones that are directly requested (because PostGraphQL requires some internally for various things: sorting, relations, etc).

Personally I don't use column-level grants on select statements, I just split the tables based on permissions boundaries into sub-tables, e.g. instead of a monolithic `user` table, I might have `user`, `user_profile`, `user_email`, `user_auth`, etc.

0x6368656174:
@benjie Splitting one monolithic table into several is not a way out of the problem. First, there can be too many small tables that will degrade performance. Secondly, if one column is available for two user roles, but the rest are not available, then you get a whole table with only one column.

0x6368656174:
@benjie The easiest solution is to replace the `select (*)` with the `select ("id")`, and everything will work, because `"Id"` in PostGraphQL should always be available to the role, if it is available at least some other column.

benjie:
Well, for the counts simply `select 1` should suffice and doesn't require us to figure out the primary key(s).

I don't find performance degrades too much with splitting concerns into separate tables; one-to-one indexed joins are very fast.

benjie:
This code:

https://github.com/postgraphql/postgraphql/blob/bg/performance-experiment-2/src/graphql/schema/collection/createCollectionQueryFieldEntries.ts#L62

Needs to specify `externalFieldNameDependencies`

benjie:
There's also an issue here:

https://github.com/postgraphql/postgraphql/blob/bg/performance-experiment-2/src/postgraphql/schema/procedures/createPgProcedureObjectTypeGqlFieldEntry.ts#L70

If you have a procedure which returns a row from the database, this isn't sufficient to parse it into a Map as is required by much of the postgraphql internals - it just leaves it as a vanilla JSON object. If you then try and access attributes on this, it throws an error like `value.get is not a function`

benjie:
Using `fixtures.return.type.transformPgValueIntoValue` seems to fix the latter.

benjie:
Seems this same issue was fixed on master only a couple months ago.

benjie:
Closing in favour of V4: #506

Issue #313, opened 2017-01-18
GraphQL Schema generation


sudheerj:
When we try to integrate GraphQL with PostgresQL then we found postgraphql as best option.After running the examples folder forum project then we ended up with few questions.How GraphQL schema will be generated? Where GraphQL Schema/queries will be stored? How we can integrate this postgraphQL in a complete front end -backend runnable application?

Note: We are successfully able to integrate with MongoDB and Cassandra DBs with GraphQL-JS implementation.

kenota:
You should be able to get schema by using graphql introspection abilities.
That is how GraphiQl is able to do autocomplete and provide documentation
in right tab.

On 19 Jan 2017 04:46, "Dedoig" <notifications@github.com> wrote:

I actually had the same question. Is there a simple way to get the complete
GraphQL schema that was figured out by postgraphql as, say, a single file?

Am most impressed with postgraphql, wonderful work!

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/calebmer/postgraphql/issues/313#issuecomment-273681203>,
or mute the thread

sudheerj:
Thanks Kenota for the response. Yeah it is based on graphql introspection abilities but I'am not sure where it stored GraphQL Schema(as a file etc) to get more control on the schema.

I'am looking for complex filter features enhancements as well.
https://github.com/calebmer/postgraphql/issues/47

calebmer:
Check out the GraphQL-js [`introspectionQuery`](https://github.com/graphql/graphql-js/blob/5cf4cef688e115e08d9f53efba07bea174ec4446/src/utilities/introspectionQuery.js#L14-L106) to see the query that is recommended to get the schema PostGraphQL generates. I’d happily accept a PR that will automatically generate this schema in a JSON file for you on PostGraphQL startup, or whenever an update from the watch feature occurs 😊

@sudheerj make sure to check out the PostGraphQL [connection procedures](https://github.com/calebmer/postgraphql/blob/master/docs/procedures.md#connection-procedures). I think that’s just the thing you are looking for when it comes to complex filters 👍

calebmer:
A feature for auto-exporting the schema was added in [`3.0.0`](https://github.com/calebmer/postgraphql/releases/tag/v3.0.0) thanks to @MaienM :tada:

PR #347, opened 2017-02-06
Added support for auto-exporting the schema

### LINKED ISSUE #282: Creating a schema file on watch reload
Hello everyone,
given `watchPg` is enabled, is there any way I can call a callback function on schema changes?

I need this to recreate a schema file for [babel-plugin-graphql](https://github.com/ooflorent/babel-plugin-graphql) on each call.

Currently I'm using [fetch-graphql-schema](https://www.npmjs.com/package/fetch-graphql-schema) to fetch schema.json and schema.graphql, however any time I change my schema I need to go through the process manually to update these files. 

It would be great if `postgraphql` could generate these files automatically. It is very useful during development when the schema is changing rapidly.

MaienM:
Fixes #282 

Supports .json and .graphql formats.

--- 

I'm hoping someone more experienced with the testing framework used can help me out with the tests, as I ran into an issue I couldn't quite figure out. If I include the `jest.mock('fs')` line, all 3 of my tests give the following error:

```
Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.

  at ontimeout (timers.js:365:14)
  at tryOnTimeout (timers.js:237:5)
  at Timer.listOnTimeout (timers.js:207:5)
```

I found an [entry](https://facebook.github.io/jest/docs/troubleshooting.html#unresolved-promises) in the jest troubleshooting docs about it, but the suggested workaround didn't seem to work. 

I also attempted to use [mock-fs](https://github.com/tschaub/mock-fs), but I ran into an [issue](https://github.com/tschaub/mock-fs/issues/27).

Any ideas on how to fix this, or how to approach the tests in a manner that avoids this issue?


MaienM:
Looking at the CI failure, it looks like when running the full test suite, the test failure happens even without the mock line. While running only these tests using `--watch`, it only happens when the mock like is included.

MaienM:
Just noticed this also addresses #313.

docs/library.md
```diff
@@ -62,6 +62,8 @@ Arguments include:
   - `watchPg`: When true, PostGraphQL will watch your database schemas and re-create the GraphQL API whenever your schema changes, notifying you as it does. This feature requires an event trigger to be added to the database by a superuser. When enabled PostGraphQL will try to add this trigger, if you did not connect as a superuser you will get a warning and the trigger won’t be added.
   - `disableQueryLog`: Turns off GraphQL query logging. By default PostGraphQL will log every GraphQL query it processes along with some other information. Set this to `true` to disable that feature.
   - `enableCors`: Enables some generous CORS settings for the GraphQL endpoint. There are some costs associated when enabling this, if at all possible try to put your API behind a reverse proxy.
+  - `exportPathJSON`: Enables saving the detected schema, in JSON format, to the given location. The directories must exist already, if the file exists it will be overwritten.
+  - `exportPathSchema`: Enables saving the detected schema, in GraphQL schema format, to the given location. The directories must exist already, if the file exists it will be overwritten.
```
calebmer: Could these options be named `exportJsonSchemaPath` and `exportGqlSchemaPath`? It’s a little more descriptive 😊

src/postgraphql/__tests__/pathgraphqlIntegrationSchemaExport-test.js
```diff
@@ -0,0 +1,28 @@
+jest.mock('fs')
+
+import withPgClient from '../../postgres/__tests__/fixtures/withPgClient'
+import createPostGraphQLSchema from '../schema/createPostGraphQLSchema'
+import exportPostGraphQLSchema from '../schema/exportPostGraphQLSchema'
+import { writeFileSync } from 'fs'
+
+test('exports a schema as JSON', withPgClient(async pgClient => {
+  const gqlSchema = await createPostGraphQLSchema(pgClient, ['a', 'b', 'c'])
+  await exportPostGraphQLSchema(gqlSchema, { exportPathJSON: '/schema.json' })
```
calebmer: You’ll need to extend the jasmine timeout to get this to work, see:

https://github.com/calebmer/postgraphql/blob/2f25ce8a9567e956f6dac55c237a1bf2e9e38003/src/postgraphql/__tests__/postgraphqlIntegrationSchema-test.js#L8-L9

Unfortunately, in CI, on our kitchen sink schemas `createPostGraphQLSchema` can be slow 😣

calebmer: Also, is the `/schema.json` path here relative or absolute? Or is it just for mocking 😊

src/postgraphql/cli.ts
```diff
@@ -37,6 +37,8 @@ program
   .option('-a, --classic-ids', 'use classic global id field name. required to support Relay 1')
   .option('-j, --dynamic-json', 'enable dynamic JSON in GraphQL inputs and outputs. uses stringified JSON by default')
   .option('-M, --disable-default-mutations', 'disable default mutations, mutation will only be possible through Postgres functions')
+  .option('--export-json [path]', 'enables exporting the detected schema, in JSON format, to the given location. The directories must exist already, if the file exists it will be overwritten.')
+  .option('--export-schema [path]', 'enables exporting the detected schema, in GraphQL schema format, to the given location. The directories must exist already, if the file exists it will be overwritten.')
```
calebmer: Could these be `--export-schema-json` and `--export-schema-graphql`?

src/postgraphql/postgraphql.ts
```diff
@@ -130,6 +131,7 @@ export default function postgraphql (
     try {
       const pgClient = await pgPool.connect()
       const newGqlSchema = await createPostGraphQLSchema(pgClient, pgSchemas, options)
+      await exportPostGraphQLSchema(newGqlSchema, options)
```
calebmer: Sweet! So this supports watch mode out of the box, right? 😊

calebmer: I am a little concerned that we have to wait for the files to be written before resolving `createGqlSchema`. Since it looks like we already catch and log errors here, could we remove `await` and add a `.catch(() => { ... })` which logs the error and exits the process? (like below)

src/postgraphql/schema/exportPostGraphQLSchema.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+export default async function exportPostGraphQLSchema (
+  schema: GraphQLSchema,
+  options: {
+    exportPathJSON?: string,
+    exportPathSchema?: string,
+  } = {},
+): Promise<void> {
+  // JSON version
+  if (options.exportPathJSON) {
```
calebmer: Maybe this should be `typeof options.exportPathJSON === 'string'`, because it would fail on `''`? Not sure why a user would ever want `''` though.

src/postgraphql/schema/exportPostGraphQLSchema.ts
```diff
@@ -0,0 +1,26 @@
...
+    const result = await graphql(schema, introspectionQuery)
+    writeFileSync(options.exportPathJSON, JSON.stringify(result))
```
calebmer: Could we make this async? We don’t need another library or utility function, we could just do it inline. It’s a little bit of non-DRY code, but I think it’s worth it:

```js
await new Promise((resolve, reject) => {
  writeFile(options.exportPathJSON, JSON.stringify(result), error => {
    if (error) reject(error)
    else resolve()
  })
})
```

Same thing below for `options.exportPathSchema`.

docs/library.md
calebmer: Also, great job on updating the documentation! We recently added [`cli.md`](https://github.com/calebmer/postgraphql/blob/master/docs/cli.md), could you update that too? It’s pretty easy, just copy the CLI help message.

src/postgraphql/schema/exportPostGraphQLSchema.ts
calebmer: This would also mean you need to change the mock implementation in the tests…

calebmer:
As for the tests, see my comment about extending the timeout and we’ll see if that works. It also looks like there are a couple of TypeScript compile errors that are getting thrown in CI.

src/postgraphql/__tests__/pathgraphqlIntegrationSchemaExport-test.js
MaienM: For some reason, I never considered the possibility of the timeout just being too low. Oops!

The `/schema.json` path is just for mocking, as the fs module will be mocked.

src/postgraphql/postgraphql.ts
MaienM: Yes, it works for watch mode 😄

src/postgraphql/schema/exportPostGraphQLSchema.ts
MaienM: `''` is not a valid file path though, so do we want to accept that? Then again, there's definitely a case to be made for handling `''` the same as other invalid paths by just throwing an error when the file write fails.

src/postgraphql/schema/exportPostGraphQLSchema.ts
MaienM: Sure, I can do that. It's a shame the `fs` module doesn't use promises natively.

src/postgraphql/__tests__/pathgraphqlIntegrationSchemaExport-test.js
MaienM: Unfortunately, increasing the timeout does not seem to help. I went as far as 3 times the value used in the tests you linked, and it still times out.

When I do not mock `fs`, the timeout does not occur, and I get an error about not being able to read property `calls` on undefined, which is expected as the `mocks` property will be undefined.

Additionally, when I do mock `fs`, it seems to never even reach the `createPostGraphQLSchema` call. If I add a logging statement right above it, it never is called.

Any ideas?

MaienM: Never mind. I had to manually mock the `writeFile` method (as the callback needs to actually be called), and in doing this the tests now work 😄

src/postgraphql/__tests__/pathgraphqlIntegrationSchemaExport-test.js
```diff
@@ -0,0 +1,35 @@
+jest.mock('fs')
```
calebmer: Did you mean to call this file `pathgraphqlIntegrationSchemaExport` instead of `postgraphqlIntegrationSchemaExport`?

src/postgraphql/postgraphql.ts
```diff
@@ -118,6 +121,16 @@ export default function postgraphql (
     _emitter,
   }))
 
+  async function handleFatalError (error: Error): Promise<void> {
```
calebmer: Looks good 👍

calebmer: Could we return `Promise<never>`?

src/postgraphql/postgraphql.ts
```diff
@@ -139,13 +153,17 @@ export default function postgraphql (
     }
     // If we fail to build our schema, log the error and exit the process.
     catch (error) {
-      // tslint:disable no-console
-      console.error(`${error.stack}\n`)
-      process.exit(1)
+      return handleFatalError(error) as never
```
calebmer: If we return `Promise<never>` do we need the `as never?`

src/postgraphql/postgraphql.ts
calebmer: …or just `never` as this doesn’t need to be an async function.

calebmer: I’ll change this 👍

calebmer:
I made some small changes, so I think this is ready to merge 👍

I’ll release this in 3.0

calebmer:
Released in [`3.0.0`](https://github.com/calebmer/postgraphql/releases/tag/v3.0.0) 🎉

murpyslaw:
@calebmer @MaienM I should probably create a separate issue of pull request for this, but I wanted to let you guys know that with `v3.5.0` `--export-schema-json` works as expected, however `--export-schema-graphql` does not.

This is how I did it:
```
        async getApolloServerSchema(): Promise<GraphQLSchema> {
            // Create a graphql schema from the postgres backend
            CTEHRComputeWebServer.singleton.gqlSchema = await createPostGraphQLSchema(this.pgConnectionOptions, 'ctehr');
            return CTEHRComputeWebServer.singleton.gqlSchema;
        }

        async writeSchemaFile() : Promise<void> {
            let schema = await this.getApolloServerSchema();
            await writeFile('../common/query/api.schema.gql', printSchema(schema));
        }
```

with the obvious imports at the top. Otherwise I'd be guessing, how to fix it in [exportPostGraphQLSchema.ts](https://github.com/postgraphql/postgraphql/blob/master/src/postgraphql/schema/exportPostGraphQLSchema.ts).

Currently it just prints `Node` and `Query` ignoring all the other types.
Hope this helps!

Cheers
rivanov

MaienM:
You should indeed open a separate issue for this, describing in more detail what doesn't work, and perhaps how to reproduce. Looking at Travis I see no regression in the tests, so I'm curious to know what the underlying issue/cause is here.

Stradivario:
I can confirm also that this issue persist in the latest release

benjie:
@Stradivario Can you confirm what version you're running please? Via `--version`

Stradivario:
@benjie version is 4.0.0-beta.3

What i did is the following
```bash
npm install -g postgraphile
```
Doesn't work
```bash
postgraphile -c postgres://dbuser:dbuserpass@182.10.0.4/postgres \ --schema myschema --export-schema-graphql './test.graphql'
```
Works
```bash
postgraphile -c postgres://dbuser:dbuserpass@182.10.0.4/postgres \ --schema myschema --export-schema-json './test.json'
```

benjie:
`schema` vs `myschema`?

benjie:
Also note that using `-X` will make the process exit once it has exported the schema.

PR #348, opened 2017-02-07
let PG* environment variables setup test connection

jinty:
A simple change to not break expectations. I was surprised PGHOST did not work.


benjie:
@jinty Because the tests run all sorts of things, including destructive stuff, I fear that picking up the PG connection string from the environment is dangerous. That's why we use the `TEST_` prefix - to be explicit.

jinty:
sure, no worries it surprised me and I just thought it might be an oversight.

benjie:
@jinty A PR adding a warning explaining why it failed to connect and how to fix it would probably get merged, if you were offering 😉

PR #354, opened 2017-02-13
Remove explain from client

calebmer:
Removes the `explain analyse` functionality from the Postgres debugger. I never liked how it was enabled with `DEBUG=postgraphql*` as it is bad for production apps who may use that debugging statement for logs by executing a side effect. As it turns out this side effect could cause mutations to be run twice which is something we _never_ want to happen (see https://github.com/calebmer/postgraphql/issues/350#issuecomment-279177585).

The feature is more troublesome than useful. It is easy enough to copy-paste a SQL statement from the debug logs into `psql` and explain/analyse there.


calebmer:
Released in `3.0.1` 🎉

PR #365, opened 2017-02-20
Add Optional Param for JWT Audience

ugiacoman:
### Custom JWT Audience
Currently, Postgraphql requires that any JWT have an audience of `postgraphql`. With this PR, the JS API now allows for a custom audience to be set. If none is provided, it will default to `postgraphql`.

### Auth0 roles support
In a previous PR, I had added support for a `roles` object to be considered and for PostGraphQL to accept the first role in that object. This felt a little bit like a hack. I did not re-add that support in this PR. If this support is needed for an Auth0 implementation, I can update the PR. 

A discussion of this can be found in the following issue: #292


jamiemccrindle:
Thanks! I'm ok with not having roles support. You can easily add a role to the token in Auth0 using a rule. You can't override the audience easily, though, so this PR is awesome.

ugiacoman:
@jamiemccrindle Good catch, I added it. Also added it to `postgraphql.ts`. Admittedly, we should probably write some tests for this.

ugiacoman:
@calebmer Of course, I'll take a look into it this week!

EyMaddis:
Just a quick note, that `aud` in the JWT spec supports arrays:
https://tools.ietf.org/html/rfc7519#section-4.1.3

> In the general case, the "aud" value is an array of case-sensitive strings

mlegenhausen:
@ugiacoman Any progress on this pull request? I would continue to work if nobody else does?

mlegenhausen:
In generell why we do not expose the whole `jwt` configuration via a `jwt` configuration field. I think in a few week we would else have the next pull request that will expose the `algorithmen` or the `issuer` field aso.

ugiacoman:
@mlegenhausen go for it 👌🏼

EyMaddis:
> In generell why we do not expose the whole jwt configuration via a jwt configuration field.
@mlegenhausen why not exactly that? 
The config for the postgres client (pgClient) is also completely exposed via configuration.
I can see the point that this might suck for CLI integration, but that could be enhanced step by step. At least it would then be possible to configure it at all.

mlegenhausen:
See #403 for further progress on this.

benjie:
#403 was merged and that seems to include this and more, so I'm closing this.

PR #366, opened 2017-02-20
Emit listen port during startup

### LINKED ISSUE #364: Emit the actual listen port during startup
When starting PostGraphQL as a child process, I'd like to communicate with it over any available TCP port. I can invoke it with `-p0` to choose such a port, but from outside the process, it's not easy to find which port was used.

```
→ postgraphql -p0

PostGraphQL server listening on port 0 🚀
```

I would love it if this line printed [`address().port`](https://nodejs.org/api/net.html#net_server_address) of `http.Server` instead.

cbandy:
Fixes #364.


calebmer:
Released in `3.0.1` 🎉

PR #367, opened 2017-02-21
Only include columns in insert statements if needed

MaienM:
Columns that have no specific value set (are `undefined`) previously were explicitly set to `default`. However, this causes problems if you do not have permission to set a column, whereas if these columns are not specified at all it works fine, and they will still ends up with their default value.

This changes that, by only including columns in the insert statement if they have a value set. If the insert statement is for multiple rows, and not all rows have the same columns, all columns that are present in at
least one row will be included. Rows that don't have that column specified will revert to the old behaviour, being set to `default` explicitly.

An example: say you have a table `foo` with the columns `name` and `secret`. You're allowed to read all of them, but you're only allowed to set `name`. Previously, if you call the mutation `createFoo` with only a value for `name`, the query will look something like:

```sql
INSERT INTO "foo" ("name", "secret") VALUES ("test", default);
```

This will cause an error, as you're not allowed to set the column `secret`. After this change, the query will look something like this:

```sql
INSERT INTO "foo" ("name") VALUES ("test");
```

This will work fine.


benjie:
I don't think this implementation will work currently, because we use a DataLoader to allow you to do multiple insert statements and have them grouped together into one efficient statement, e.g. 

```graphql
mutation {
  addThing(...)
  addThing(...)
  addThing(...)
}
```

becomes

```sql
insert into thing (col1, col2) values (?, ?), (?, ?), (?, ?);
```

and the columns that are not `undefined` may differ for each of these calls.

benjie:
Nonsense, Benjie, Michon has already handled this. Read the code!

benjie:
This looks good to me 👍  Could you add a test that fails on master and passes on this branch for this code please, to avoid regressions?

MaienM:
Sure, I'll get on that.

src/postgres/inventory/collection/PgCollection.ts
```diff
@@ -133,23 +132,35 @@ class PgCollection implements Collection<PgClassType.Value> {
...
 
+        // Get the fields that actually have been used in any of the values, so
+        // we can leave the rest out of the query completely to prevent
+        // permission issues
+        const fields = new Map()
+        values.forEach((map) => {
+          map.forEach((value, key) => {
+            if (typeof value !== 'undefined') {
+              fields.set(key, this.type.fields.get(key))
+            }
```
calebmer: My one concern is that this is `O(vn)` where `v` is the number of values and `n` is the maximum number of fields in one of the values. Is there a more efficient algorithm for this? Can we do some other work while iterating? Think about these questions only for a bit, if you can’t think of a good answer I’m comfortable with merging this as an initial implementation.

Ideally I think we could take a similar approach here as we are taking in https://github.com/calebmer/postgraphql/pull/342, although this is a good first implementation. If you could put a `TODO` comment here that reads something like: `TODO: Use field information from GraphQL instead of iterating through values` we can address it when it becomes a problem 👍

src/postgres/inventory/collection/__tests__/PgCollection-test.js
```diff
@@ -80,6 +80,28 @@ test('create will insert new rows into the database', withPgClient(async client
...
+
+  client.query.mockClear()
+
+  await Promise.all([
+    collection1.create(context, value1),
+    collection1.create(context, value2),
+    collection1.create(context, value3),
+  ])
+
+  const query = client.query.mock.calls[0][0].text
```
calebmer: Could you either check the entire text of the query (vs. using `toMatch`) with either an inline string literal, or a snapshot? `expect(client.query.mock.calls[0][0].text).toMatchSnapshot()`. Snapshot is probably the easiest of course. In the future I’d like to be snapshotting more of the SQL queries we generate.

Could you also add a sanity check here: `expect(client.query.mock.calls.length).toBe(1)` just so that we can be sure that no other queries are being run?

src/postgres/inventory/collection/PgCollection.ts
MaienM: I hadn't really considered that yet, but I can see how that could become an issue. I unfortunately do not see a way to avoid that with this approach.

Another approach to this problem would be to detect which columns can be inserted on through introspection instead. This only has to be done at startup/watch updates, and does not involve extra looping, so this would effectively be `O(1)`. We would go back to sometimes specifying `default` for columns that we could leave out completely, but as long as you have the rights to set those columns this is fine. I am not (yet) familiar enough with the project to achieve this though, unfortunately.

MaienM: Another benefit to detecting these columns through introspection would be that this could also be reflected in the GraphQL schema/docs. We could simply leave columns that cannot be set out of the `Input` objects, thereby also preventing this from being a runtime issue.

src/postgres/inventory/collection/__tests__/PgCollection-test.js
MaienM: I had that in there initially for that exact reason, but I seem to have removed it accidentally. Whoops! Will add it back.

MaienM:
This is indeed needed for `INSERT` statements on tables with column-level `GRANTS`. This is what inspired me to do this, as I have a few columns that are managed on the database level that I do not want editable (mostly `created_at` and `updated_at` timestamps).

src/postgres/inventory/collection/PgCollection.ts
benjie: This won't work for my schema; we use multiple roles and those roles have different column access grants. Unless you enumerated all of the roles with access, and applied those learnings at run-time; but that seems complex.

MaienM: I failed to consider using multiple roles. That will indeed complicate things significantly, to the point where it may not be worth it at all. I think @calebmer's proposal to, if the performance of this code ever becomes an issue, use the field information from GraphQL is probably best then. For now I'll just add a `TODO` comment about it.

benjie:
Awesome, this resolves an issue I've had on another project also (again: updatedAt and createdAt). Thanks! 👍

calebmer:
Released in 3.1.0 🎉

PR #369, opened 2017-02-21
Fix startup crash when a non-superuser tries to watch

calebmer:
Fixes https://github.com/calebmer/postgraphql/issues/368


calebmer:
Wait, on second thought we aren’t in a transaction so this wouldn’t work, right? (I made this PR through the GitHub editor so I haven’t tested)

The watch fixtures are in a transaction, so we should auto-rollback: https://github.com/calebmer/postgraphql/blob/master/resources/watch-fixtures.sql

MaienM:
I can confirm that this does not fix the issue, unfortunately.

calebmer:
Closing in favor of: https://github.com/calebmer/postgraphql/pull/371

PR #370, opened 2017-02-21
Fixes graphiql so it works with koa

smh:
When run with koa, static files for graphiql would fail with error
message `Error: Can't set headers after they are sent.`

This is caused by pillarjs/send being incompatible with koa, at least in
the manner it was used here. See pillarjs/send#118 and pillarjs/send#119
for more discussion


mlegenhausen:
For everybody who is working for the pull request to land you can work around this issue with [koa-connect](https://github.com/vkurchatkin/koa-connect). Simply wrap the `postgraphql` call in `c2k` and the error is gone.

benjie:
CI issues: https://travis-ci.org/calebmer/postgraphql/jobs/205769637

benjie:
I've rebuilt a few times and cleared the relevant cache but it still fails on an npm issue.

smh:
Seems to be unrelated to the changes here, built locally and got the same error both for master branch and this branch (node v4.3.2 and npm v2.14.12).

Looks like it is caused by graphiql@0.8.0 not happy with graphql versions >= 0.9. Any reason we can't update the graphiql dependency to follow graphql?

smh:
@benjie Created a pull request to get the devDependencies in sync. See #377

benjie:
I defer to @calebmer on that; he's got a better grasp of the various versions and why they were chosen than I do.

calebmer:
CI was passing when I just looked. Updated the branch, and if CI still passes we can merge 😊

calebmer:
Released in 3.1.0 🎉

PR #371, opened 2017-02-22
Fix startup crash when a non-superuser tries to watch

### LINKED ISSUE #368: Server crashes when watch is used by non-super user
Hi,
when running PostGraphQL with `pgWatch` enabled, but without access to a super user, the server will not start and crash instead.

```
Failed to setup watch fixtures in Postgres database ️️⚠️
This is likely because your Postgres user is not a superuser. If the
fixtures already exist, the watch functionality may still work.
error: current transaction is aborted, commands ignored until end of transaction block
    at Connection.parseE (service-postgraphql/node_modules/pg/lib/connection.js:554:11)
    at Connection.parseMessage (service-postgraphql/node_modules/pg/lib/connection.js:381:17)
    at Socket.<anonymous> (service-postgraphql/node_modules/pg/lib/connection.js:117:22)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:189:7)
    at readableAddChunk (_stream_readable.js:176:18)
    at Socket.Readable.push (_stream_readable.js:134:10)
    at TCP.onread (net.js:551:20)
```

My guess is that there is automatic rollback missing.

Related question: which permissions are required to this to work?

MaienM:
The issue was that the transaction failed but was not rolled back, causing all subsequent commands to fail. This just adds a rollback in the catch, and it adjusts the tests to expect this.

Fixes #368, replaces #369.


benjie:
What voodoo is this?

> Issuing `ROLLBACK` outside of a transaction block emits a warning and otherwise has no effect.
> -- https://www.postgresql.org/docs/9.6/static/sql-rollback.html

benjie:
Ah; the query is from:

https://github.com/calebmer/postgraphql/blob/6b3374baa974d8efd110970e4bd9431b8b3771f3/resources/watch-fixtures.sql

This query issues a `begin;` it seems that the error thrown prevents further statements being executed. This is unexpected to me... I would expect pgClient to execute all of the statements issued, even if errors occur, ultimately culminating in the `commit;` (which would perform an rollback due to the errors) and then return the results.

PR #374, opened 2017-02-24
replace = with IS NOT DISTINCT FROM

valorize:


valorize:
The snapshots changed quite a lot and I'm not confident enough to update them.

valorize:
Tricky. Why do you write null -> false? As I understand null it is unknown so could be false and/or true. (Schrödinger's cat like). It is definitely different behaviour, but I think a better one. In the first examples we assume null to be something unknown (1,2,3...). And in the second is null rather a discrete Symbol. In order to use null as a condition it is necessary to see null as a dicrete symbol otherwise you would have to return all and/or nothing as you do not know which null is a matching value and which not.

benjie:
When used as part of a where clause, null is falsy. So previously rows that had null values for one of the conditions (whether equal, not equal, greater than, less than, etc) would be excluded. This change would make them be included. I agree it's probably a good thing, but it is a breaking change and thus needs a major version bump. Maybe we can bundle this, numeric as string, and the new performance stuff into a 4.0 release? There might be a few more breaking changes we want to make too.

benjie:
Correction: this change will make them be included in the case of equals.

valorize:
I agree with 4.0. Do we label this with a milestone or do you rather put it all in one 4.0 branch?

benjie:
Personally I'd have a 'next' branch which we keep up to date with master (bugfixes would still go to master) and release as a new major version when we're ready; but I'm not sure on Caleb's opinions on this.

calebmer:
Would this still be a breaking change if we used `x is null` instead of `=` when we are comparing against a null value? I consider the behavior of: `myConnection(field: null)` not matching null fields to be a bug, so the `is null` change would be a bug fix.

My current thought on breaking changes are that we should not be afraid to break things. If we have a breaking change that makes sense we should feel free to ship it and increment the major version number. If we treat the major version number as sacred it may slow us down 😊

I don’t like making breaking changes, but I don’t think that if we have a good breaking change we should wait till a large number of changes accumulate before releasing it. Does that make sense?

benjie:
There's quite a few suggestions for breaking changes at the moment, I'd suggest bundling them all together into a 'next' branch and releasing that in around a month's time (in the mean time we can release it with a 'next' tag on npm so people can opt into it)

This definitely feels like a bug; however it could cause issues for existing applications that are accidentally relying on the broken behaviour so I think it would be wise to class it as breaking.

valorize:
I think bundling up breaking changes into a next branch is a good idea. A ~ two monthly circle seams reasonable. @calebmer Yep `is null` checks would still be a breaking change and a bug fix as well.

0ff:
Sorry to just dump this here, but I noticed this passage in the [Postgres docs](https://www.postgresql.org/docs/9.6/static/plpgsql-statements.html) today and though I'd share with the appropriate people on this PR:

> (At present, IS NOT DISTINCT FROM is handled much less efficiently than =, so don't do this unless you must. See Section 9.2 for more information on nulls and IS DISTINCT.)

If you've taken that into account then please don't mind me.

benjie:
@valorize Given @0ff's comment above (which is totally legit) can we special-case null with an `IS NULL` check please. I'll then merge this into a `next` branch which will go out with the next major release.

valorize:
Ups, I merged bg/perf... into the branch. What will the next branch be based on?

benjie:
I'm not sure if we'll merge bg/perf, it's a very messy PR (it was my first real experience of the internals of GraphQL and I threw myself in at the deep end without knowing PostGraphQL's internals). Certainly need to tidy it up a bunch before it can be merged.

We should create a new branch `next` for all these breaking changes and keep it up to date with master. I'll probably use [this](https://bots.sh/s/0bTURxGpDO0S) to keep it up to date.

benjie:
I'd love to see this implemented as a plugin in v4; for now I'm going to close it.

PR #377, opened 2017-02-27
fix(ci): peerdeps for graphql/graphiql must match

smh:
Avoid unmet peer devDependencies between graphiql@0.8.0 and
graphql@0.9.1


smh:
Perhaps the devDependencies for graphql/graphiql should be ranged just like the dependency for graphql to avoid this happening again in the future?

calebmer:
One thing we need to worry about is that we do some dark magic with GraphiQL to get a good DX on live reloading schemas: https://github.com/calebmer/postgraphql/blob/53c6a3139b9c7247e5dc1d6761f610ddef61f384/src/postgraphql/graphiql/src/components/PostGraphiQL.js#L96-L181

We pinned the GraphiQL version number so that we could be sure that our “hack” always worked with the version of GraphiQL we were shipping.

It may be worth pinning GraphiQL again once we confirm that the hack still works.

calebmer:
Released in 3.1.0 🎉

Issue #372, opened 2017-02-23
Changing the limit for POST body size


kenota:
I am using postgraphql exposed Postgres function to process quite a bit amount of data (not storing binary files in db, use case is providing array of objects to db to process), and I have encountered following error reply when doing POST:

`{"errors":[{"message":"request entity too large"}]}`

It seems that I am hitting default limit of express post body parsing middleware which can be easily changed in code. 

Just wanted to check if postgraphql exposing an option/way to configure this which is not mentioned in doc? 

If not, I assume the best solution to use postgraphql as a library in custom small nodejs/express server?

calebmer:
We may want to discuss raising the default, otherwise I’m totally open to adding this option 👍

It looks like @guyemerson referenced this PR in a commit to his fork: https://github.com/guyemerson/postgraphql/commit/abbc8006719869e338f113277bcf3ea53443b086. That change looks good, although we may want to change the variable name.

calebmer:
Released in 3.1.0 🎉

PR #379, opened 2017-03-02
Changing the limit for JSON bodyParser (#372)

guyemerson:
Add sizeLimit as an option, which controls the maximum size of a JSON body that can can be accepted by bodyParser.json.  This option is available both at the command line and as middleware.


guyemerson:
The only check that is still failing is to do with the version of graphql, but that is not affected by this pull request.

docs/cli.md
```diff
@@ -38,6 +38,7 @@ The usage of the `postgraphql` binary is as follows. To pull up this documentati
     -a, --classic-ids                use classic global id field name. required to support Relay 1
     -j, --dynamic-json               enable dynamic JSON in GraphQL inputs and outputs. uses stringified JSON by default
     -M, --disable-default-mutations  disable default mutations, mutation will only be possible through Postgres functions
+    -l, --size-limit                 set the maximum size of JSON bodies that can be parsed (default 100kB)
```
calebmer: Could we change the name to `--body-size-limit`? Could we also add a note that you can define it using a pretty size like `200kB`.

docs/library.md
```diff
@@ -64,6 +64,7 @@ Arguments include:
   - `enableCors`: Enables some generous CORS settings for the GraphQL endpoint. There are some costs associated when enabling this, if at all possible try to put your API behind a reverse proxy.
   - `exportJsonSchemaPath`: Enables saving the detected schema, in JSON format, to the given location. The directories must exist already, if the file exists it will be overwritten.
   - `exportGqlSchemaPath`: Enables saving the detected schema, in GraphQL schema format, to the given location. The directories must exist already, if the file exists it will be overwritten.
+  - `sizeLimit`: Set the maximum size of JSON bodies that can be parsed (default 100kB).
```
calebmer: Could we change this to `bodySizeLimit` and add a similar note about pretty bytes?

calebmer:
I tried restarting the build. I’ll investigate if it keeps failing 😣

benjie:
Thanks! 👍

PR #382, opened 2017-03-03
Fix undefined values to be assumed as null values

mlegenhausen:
Problem:
When sending an graphql query that contains the only required values all undefined values in the query get replaced by null in the default sql query mutation builder. This makes the definition of default values when creating a table useless cause the default value is never
applied to the database.

Solution:
This fix makes sure that only defined values in the graphQL mutation get translated to the 
appropriate postgres type and the default pg query builder has a chance to use default instead of null.

The current unit tests do not cover this scenario. I need some generell help how to add such a scenario in postgraphql.

I will also try to improve the query generation in the insertLoader by only including columns defined in the graphql mutation.

### DIFF
--- a/src/graphql/schema/type/getGqlInputType.ts
+++ b/src/graphql/schema/type/getGqlInputType.ts
@@ -153,10 +153,17 @@ const createGqlInputType = <TValue>(buildToken: BuildToken, _type: Type<TValue>)
             },
           }))),
         })),
-        fromGqlInput: gqlInput =>
-          gqlInput != null && typeof gqlInput === 'object'
-            ? type.fromFields(new Map(fieldFixtures.map(({ fieldName, key, fromGqlInput }) => [fieldName, fromGqlInput(gqlInput[key])] as [string, mixed])))
-            : (() => { throw new Error(`Input value must be an object, not '${typeof gqlInput}'.`) })(),
+        fromGqlInput: gqlInput => {
+          if (gqlInput === null || typeof gqlInput !== 'object') {
+            throw new Error(`Input value must be an object, not '${typeof gqlInput}'.`)
+          }
+
+          const entries = fieldFixtures
+            .filter(({ key }) => key in gqlInput)
+            .map(({ fieldName, key, fromGqlInput }) => [fieldName, fromGqlInput(gqlInput[key])] as [string, mixed])
+
+          return type.fromFields(new Map(entries))
+        },
       }
     },
 


benjie:
In PgCollection we use the keyword `default` to insert not null but the default value for the column:

https://github.com/calebmer/postgraphql/blob/a88a9836722c5d42c7cae93fe901c095d7c704f6/src/postgres/inventory/collection/PgCollection.ts#L154

Where are you getting this issue?

mlegenhausen:
Example following schema:

```
create table test (
  id serial primary key,
  name text not null,
  created_at timestamp default now()
)
```

Now I use your default generated mutation.

```
mutation createTest($input: TestInput!) {
  createTest(input: $input) {
    test {
      id
      name
      createdAt
    }
  }
}
```

with following input

```
{
  input: {
    name: 'Test'
  }
}
```

The resulting query will look like (just an example).

```
with __local_0__ as ( insert into "public"."test" ("id", "name", "created_at") values (default, $1, default) returning * ) select row_to_json(__local_0__) as object from __local_0__
```

The id get correctly replaced by `default` I think cause it is a serial type. But the `created_at` value gets replaced by `null` which is not intended by my query. I never set it to `null`. Cause you internally choose a nullable type instead of simply ignoring it. Which makes `default` types useless. This is what my fix prevents by removing them from the values that should be handled.

The resulting query after my fix is:

```
with __local_0__ as ( insert into "public"."test" ("id", "name", "created_at") values (default, $1, null) returning * ) select row_to_json(__local_0__) as object from __local_0__
```

cause I removed it from the `fields` variable.

mlegenhausen:
I updated my comment to hopefully clarify the problem. If it is intended behaviour how can I then force to use the default value defined in my schema via the graphql query?

Thanks for your help and thanks for this project!

benjie:
Ahhhh, I see - thanks for expanding. So if I understand you right, you're saying:

"When an insert takes place (`createTest`), if you omit a nullable field from the `$input` variable then the value of that field will automatically be set to `null` (rather than left as `undefined` as would be the case for a non-nullable field); thus the ["use default value if undefined" logic](https://github.com/calebmer/postgraphql/blob/a88a9836722c5d42c7cae93fe901c095d7c704f6/src/postgres/inventory/collection/PgCollection.ts#L154) does not fire, preventing the `default` value from being used and causing `null` to be stored instead"

mlegenhausen:
Yes :)

benjie:
@mlegenhausen Yes, the advantage is we can bundle a lot of inserts together with varying inputs and execute via one SQL statement with multiple `VALUES()`.

@calebmer I'm not sure where the change should be made to handle this; is the issue coming from graphql-js or from us?

> When an insert takes place `mutation createTest($input: TestInput!) {...}`, if you omit a nullable field from the `$input` variable then the value of that field will automatically be set to `null` (rather than left as `undefined` as would be the case for a non-nullable field); thus the "[sql.query\`default\` if undefined](https://github.com/calebmer/postgraphql/blob/a88a9836722c5d42c7cae93fe901c095d7c704f6/src/postgres/inventory/collection/PgCollection.ts#L154)" logic does not fire, preventing the default value from being used and causing `null` to be stored instead.

I have not verified this issue myself.

calebmer:
@benjie I bet it’s us. There is some coercion going on, so I wouldn’t be surprised if we switched `undefined` to `null` somewhere 😊

@mlegenhausen we just merged: https://github.com/calebmer/postgraphql/pull/367 which contains a fix you mention in your OP that you wanted to make. Would this fix part of the issue?

mlegenhausen:
@calebmer it sounds like the same problem, I will evaluate the changes and give you feedback today.

mlegenhausen:
@calebmer it does not solve the issue. Now the query from my given example looks like

```sql
with __local_0__ as ( insert into "public"."test" ("name", "created_at") values ($1, null) returning * ) select row_to_json(__local_0__) as object from __local_0__
```

The default values where removed but still all `undefined` values in my mutation parameter gets replaced by `null`.

mlegenhausen:
Ok in some conditions I get the default type working when combining `not null` and `default` in my column definition. But it seems impossible to defined a column with `default` and `null` allowed (currently not a use case for me). My updated table looks like this

```sql
create type state as enum (
  'draft',
  'released'
);

create table test (
  id serial primary key,
  state state not null default 'draft',
  name text not null,
  created_at timestamp not null default now()
)
```

`created_at` does not work as expected. But when I now apply the same principal on `state` I got the following error back via `graphiql`.

```json
{
  "data": {
    "createTest": null
  },
  "errors": [
    {
      "message": "Input value must be an enum variant.",
      "locations": [
        {
          "line": 16,
          "column": 3
        }
      ],
      "path": [
        "createTest"
      ]
    }
  ]
}
```

This seems more of a graphql bug which is not in the scope of this pull request.

To sum up what this pull request does: It allows the definition of columns that have a `default` value and allow `null` by not falling not back to a `null` when the column/key was not explicitly set to `null` via the graphql mutation.

calebmer:
I’m going to have to think some more on whether or not this change will not have far reaching impacts. My initial intuition is that it would break some [`isTypeOf`](https://github.com/calebmer/postgraphql/blob/6cf9ac7d1b28440523d6938e0c993d007f0169ed/src/interface/type/BasicObjectType.ts#L51-L65 or https://github.com/calebmer/postgraphql/blob/6cf9ac7d1b28440523d6938e0c993d007f0169ed/src/postgres/inventory/type/PgClassType.ts#L65-L67) checks that currently exist, but I do want to either remove `isTypeOf` or find a way to allow much looser objects.

An alternative implementation which likely has less far reaching impacts would be to get `null`s instead returning `default`s here: https://github.com/calebmer/postgraphql/blob/6cf9ac7d1b28440523d6938e0c993d007f0169ed/src/postgres/inventory/collection/PgCollection.ts#L166. On second thought that wouldn’t work because a `null` is different from `undefined` from a GraphQL perspective, so that wouldn’t work 😣

Do you know why the tests are failing?

benjie:
@mlegenhausen I think the enum fail might be because for some reason PostGraphQL uppercases the enum values; so you need to provide them in uppercase? If this isn't the issue just dismiss this comment, but it caught me out when I started.

mlegenhausen:
@calebmer the last thought was mine too. The problem seems complex and I think is out of my current knowledge about postgraphql. I will investigate the test errors.

@benjie Same error even when I switch all my enum values to uppercase. I will open a separate issue for this one.

bradleyayers:
So I just ran into this issue, from the default enum scenario, but also from another not discussed — restricted insert permissions. I have something like the following:

```sql
create table test (
  id serial primary key,
  name text not null,
  deleted_at timestamp
);
grant insert ("name") on table "test" to "example_user";
```

The problem is that the `insert` query generated by postgraphql attempts to insert null into `deleted_at`:

```
insert into "public"."test" ("name", "deleted_at") values ($1, null) returning *
```

I thought about it a bit more, and decided it's fine to allow `insert` to include a `deleted_at` value. But regardless it was a bit of a gotcha. 

So I'm still stuck with the `Input value must be an enum variant.` error, and for the moment need to just pass through the correct value. What's the latest thinking on this issue? Is there anything I can do to help move this issue forward?

benjie:
This might already be addressed in v4: https://github.com/postgraphql/graphql-build/blob/7165585cea78c3f1f8192841de4199fdc3db4f2b/packages/graphql-build-pg/src/plugins/PgMutationCreatePlugin.js#L152-L178

One issue with the above code, however, is if you have a default on a nullable column but you want to explicitly set it to null the above code ignores that and just uses the default.

benjie:
(Sorry that prettier messes up the formatting of ternaries inside of template literals, I've not had a chance to look at that yet - I'll probably just move the ternaries outside.)

bradleyayers:
Interesting! It sounds like I should dive in and play around with v4 :smile:. Will report back when I've had some time to look into it.

benjie:
Does anyone know where we are on this issue currently?

benjie:
I'm closing this because I _believe_ it's fixed in v4; but please check and if it isn't raise an issue ASAP so we can fix it before release 👍

Issue #87, opened 2016-07-17
My vision for the future of PostGraphQL


calebmer:
Hi, I’m @calebmer, author of PostGraphQL a framework which turns your PostgreSQL database into a GraphQL server with minimal configuration. In this issue, I want to share my vision for the future PostGraphQL.

As you may know, PostGraphQL currently does one thing fairly well: turns a PostgreSQL database into a GraphQL API. You also may have heard of [PostgREST](http://postgrest.com/) which I contributed to in the past, it’s written in Haskell you should check it out. Similarly PostgREST also does one thing fairly well: turns a PostgreSQL database into a REST API.

Tools like these have a very high utility for certain use cases, but at the same time they are inflexible. What if I wanted a GraphQL API for my MySQL database using PostGraphQL? Well that wouldn’t be too easy. Every time we create another one of these tools we are repeating a lot of logic, because every API is generally just doing the same four things: creates, reads, updates, and deletes. CRUD.

It’s weird that we standardize on front-end frameworks like React or Angular, but our backends which are often just glorified CRUD interfaces get written from scratch.

I want PostGraphQL to be an opinionated tool that’s part of a more powerful ecosystem. An ecosystem that is focused on abstracting the data layer from the data interface. So instead of having one tightly coupled tool that look like:

PostgreSQL → GraphQL

We could have two loosely coupled tools that look like:

PostgreSQL → Interface → GraphQL

One tool that turns our PostgreSQL database into an abstract interface, and one that turns our abstract interface into a GraphQL API.

When we add that extra layer, we can start interchanging parts. Instead of PostgreSQL → Interface → GraphQL, we could have MySQL → Interface → GraphQL for half the effort. We also get the advantage of when our Interface → GraphQL tool improves, both toolchains benefit.

[...]

PR #384, opened 2017-03-03
Simplify access to getPgClientFromContext

mlegenhausen:
To get the most out of withPostGraphQLContext the getPgClientFromContext is required to allow the executing of queries inside the postgraphql context. This pull request simplifies the access to this function without ugly require paths that were already mentioned in other issues.


benjie:
(CI is passing in 2 out of 3 environments; we seem to be having consistent troubles with one of them)

benjie:
@calebmer 2 line change here; are you happy with an additional export being added?

smh:
@benjie @calebmer the CI troubles are caused by invalid peer dependencies between graphql and graphiql. See #377 for a fix.

calebmer:
We should evaluate exposing the client on the context without a symbol. That should be a simpler change and we wouldn’t have to add another export. Perhaps it could be something like: `context.pgClient`.

I do have an irrational negative reaction exporting a file nested within a sub-module 😊. If we decide we like exporting a function better it would make me feel a little better if we could add a level of indirection. Exporting from: https://github.com/calebmer/postgraphql/blob/53c6a3139b9c7247e5dc1d6761f610ddef61f384/src/postgres/index.ts first and then exporting from `./postgres`. This would also be good because I believe we use this function in `./postgraphql` a couple of times.

benjie:
I'm happy to defer to Caleb on this, though I can see advantages to both approaches I don't see much value in bikeshedding it.

mlegenhausen:
I would go for the `pgClient` approach. This makes the api much more easier and natural to use.

benjie:
@mlegenhausen Are you able to make the relevant changes?

mlegenhausen:
@benjie yes I will try it out.

mlegenhausen:
@benjie sorry for the long waiting of the pull request (some privat problems)

Here my refactoring. I was able to replace all `$$pgClient` and `getPgClientFromPostGraphqlContext` by a simple `pgClient` variable. I also introduced the `PostGraphQLContext` interface.

calebmer:
Great work! But I’m not sure I want to get rid of `getPgClientFromContext` internally 😣

Ideally it would be refactored to just return `context.pgClient` instead of `context[$$pgClient]`.

The reason for this is if we want PostGraphQL to evolve to support other databases then we can’t have a context that encodes a Postgres specific assumption (a `pgClient`).

mlegenhausen:
@calebmer How exactly will `getPgClientFromContext` help being more database independent?

calebmer:
The `context: mixed` change to `context: PostGraphQLContext` is what makes it harder to support other databases in the future. `mixed` represents any context type. It could have a `pgClient` instance, it could have a `mysqlClient` instance, or a `mongoClient` instance for example. Ideally the `interface` package wouldn’t know exactly what types could be on context because any other package could have its own context type.

A good rule of thumb when working on the PostGraphQL codebase is that:

- `src/interface` should not import from anything other then `src/interface`. Not from `src/postgres`, `src/graphql`, `pg` or anything other than the files in `src/interface`. This allows it to be completely separate and extensible by other packages in the future.
- `src/postgres` should not import anything from `src/graphql` or `src/postgraphql`.
- `src/graphql` should not import anything from `src/postgres` or `src/postgraphql`.
- `src/postgraphql` can import anything from anywhere.

If we decide these rules are too restrictive we can think about breaking them down, but even in a world where we don’t support multiple backends I believe this separation of concerns makes for a very good architecture in this large project 😊

A `PostGraphQLContext` type means `src/interface`, `src/graphql`, and `src/postgres` which are all low level packages must know about `src/postgraphql` which is the highest integration point. I like `getPgClientFromContext` because it moves a check which is very hard to do statically given our constraints to runtime. It will throw an error if `pgClient` doesn’t exist on context and it knows how to deal with `mixed`.

I’m open to finding different ways to add types to `context`, but this way breaks some of the fundamental assumptions we can currently make about the codebase.

mlegenhausen:
@calebmer thanks for your feedback. I will change the imports. I hoped for this feedback ;)

I am not really sure if this should be the way to go. As you see in my other pull request I am little type-nazi ;) cause I have the learned the struggled of having to lose typing the hard way.

My naive approach to support multiple databases would be pull everything postgres related out of the deeper levels and hide this database specifics behind a facade which other databases can implement. So the type of `pgClient` would be `DatabaseFacade` and of cause `pgClient` would be called only `client` or better `database` cause it included much more than a low level database client. So the setup of postgraphql would look like this:

```ts
postgraphql({
  database: new PostgresDatabase()
})

postgraphql({
  database: new MySqlDatabase()
})

postgraphql({
  database: new MongoDatabase()
})
```

and the fasade would look like this

```ts
interface DatabaseFacade {
  // Returns all database specific information to construct the graphql interfaces
  insepect(): DatabaseInsepectionResult;
  connect()
  setSetting(key: string, value: string)
  ...
}
```

But in all cases there would be no need for a `mixed` interface only to get the database access layer down the road.

I hope you get an idea what my current picture is of getting postgraphql database independent ;) Would be great to get some feedback about the ideas. Maybe I need to reread your vision in #87 are there any public design documents where it is possible to contribute to?

calebmer:
@mlegenhausen yep! That should certainly be something we explore. The thing that would make that approach difficult, however, is what if we wanted to support multiple databases? For instance, the user could have a few tables from Postgres and a few collections from Mongo. Or a few collections from a REST API even.

The safest way to introduce better type-checking to context without making major changes would to be start making everything in interface generic and accept a `TContext` type.

For instance:

```ts
interface Collection<TContext, TValue> {
  create(context: TContext, value: TValue): Promise<TValue>
}
```

I was waiting on TypeScript default types for generics before going all in on this. Perhaps we should start moving in this direction now, though?

calebmer:
Also, it doesn’t look like anything changed 😣

benjie:
This PR is way beyond my knowledge of TypeScript right now. We'll need Caleb's input if we are to advance it.

benjie:
v4 uses `pgClient` on the context object.

ed-g:
Apologies in advance, for 3.3.0 users here is what I used to pull out the context.

```javascript
const _ = require('lodash')
function hacketyHackHackGetPgClientFromContext ( context ) {
    // dig out the pgClient property so we can use it!
    var $$pgClient = _.find(
            Object.getOwnPropertySymbols(context), 
            x => { return x.toString() == 'Symbol(postgres/client)' });
    return context[$$pgClient];
}
```

benjie:
I'm so sorry I let this PR stagnate. I'd rather keep any change to v3 to a minimum but if this is still desired functionality for the v3 branch and using v4 isn't a possibility then I'd like to see as small a PR as possible that implements it.

In the mean time I'm going to close this to help me manage the repo.

Thanks 🙏

PR #391, opened 2017-03-14
Production dependencies fixed

mlegenhausen:
This commit fixes the problem that `graphql` is missing when running `npm prune --production` before starting postgraphql cause it was also defined in `devDependencies`. Also is `graphiql` moved to `dependencies` cause it is needed when enabling it via the postgraphql options.


calebmer:
When are you running `npm prune --production`? It’s important that we can control the version of `graphql` when testing, and `graphiql` isn’t a production dependency because it gets bundled into the build. See: https://unpkg.com/postgraphql@3.0.1/build/postgraphql/graphiql/public/static/js/

benjie:

Issue #380, opened 2017-03-02
Slow performance when using orderBy


danielbuechele:
I have table of articles with about 100k rows. When I sort this table using `orderBy: PUBLICATION_DATE_DESC` the response is pretty slow. Without the `orderBy` it's fast.
I started debugging it:
`SELECT * FROM articles ORDER BY publication_date DESC LIMIT 10` is super fast. The query generated by PostGraphQL looks like this and is slow.
```
select to_json(__local_0__) as value from "public"."articles" as __local_0__  order by "publication_date" limit 10;
```
Once I remove `to_json` or the ordering it's fast again. Currently there is no index on the publication_date column. I don't understand what's happening there. A first guess would be, that all rows are converted to json before the LIMIT is applied, but I don't understand how this is related to the `ORDER BY`.

benjie:
@danielbuechele Please submit an `explain analyze select...` and durations (run each one 5 times and give the lowest result) for the four queries:

1. `select * from articles limit 10;`
2. `select * from articles order by publication_date desc limit 10;`
3. `select to_json(articles) from articles limit 10;`
4. `select to_json(articles) from articles order by publication_date desc limit 10;`

I suspect the missing index is the issue.

danielbuechele:
Really appreciate your help:

1. `select * from articles limit 10;`
```
Limit  (cost=0.00..1.17 rows=10 width=1072) (actual time=0.013..0.022 rows=10 loops=1)
  ->  Seq Scan on articles  (cost=0.00..10259.08 rows=87908 width=1072) (actual time=0.013..0.019 rows=10 loops=1)
Planning time: 0.082 ms
Execution time: 0.063 ms
```

2. `select * from articles order by publication_date desc limit 10`
```
Limit  (cost=12158.74..12158.77 rows=10 width=1072) (actual time=94.997..95.003 rows=10 loops=1)
  ->  Sort  (cost=12158.74..12378.51 rows=87908 width=1072) (actual time=94.996..95.001 rows=10 loops=1)
        Sort Key: publication_date DESC
        Sort Method: top-N heapsort  Memory: 36kB
        ->  Seq Scan on articles  (cost=0.00..10259.08 rows=87908 width=1072) (actual time=0.014..51.099 rows=87908 loops=1)
Planning time: 0.124 ms
Execution time: 95.045 ms
```

3. `select to_json(articles) from articles limit 10;`
```
[...]

PR #396, opened 2017-03-16
fix(postgres): use CTEs for paginators #380

danielbuechele:
This uses Postgres common table expressions to make sure the `to_json`-function is only called on the results of a pagination, as discussed in #380

---

@Benjie's explanation for this PR:

It seems that PostgreSQL sometimes performs the `to_json()` whilst ordering the rows and before applying the limit - this means that a `select to_json(tbl) from tbl order by col limit 10` is **significantly** slower than both `select tbl from tbl order by col limit 10` and `select to_json(tbl) from tbl limit 10`. To work around this we use a common table expression (CTE) as an optimisation barrier to first find the matching rows, and only after doing so perform to_json on the relevant rows (and not the irrelevant ones). This yields a significant performance improvement.


benjie:
I have an issue with `intSetQuery` on #342 also.

src/postgres/inventory/paginator/PgPaginatorOrderingAttributes.ts
```diff
@@ -66,37 +66,40 @@ implements Paginator.Ordering<TInput, PgClassType.Value, AttributesCursor> {
       throw new Error('Before cursor must be a value tuple of the correct length.')
 
     const aliasIdentifier = Symbol()
+    const aliasIdentifier2 = Symbol()
```
calebmer: Could we think of a better name for these variables? How about `cteIdentifier`?

src/postgres/inventory/paginator/PgPaginatorOrderingOffset.ts
```diff
@@ -121,17 +121,20 @@ implements Paginator.Ordering<TInput, TItemValue, OffsetCursor> {
     }
 
     const aliasIdentifier = Symbol()
+    const aliasIdentifier2 = Symbol()
```
calebmer: How about `cteIdentifier` here as well?

src/postgres/inventory/paginator/PgPaginatorOrderingAttributes.ts
```diff
@@ -66,37 +66,40 @@ implements Paginator.Ordering<TInput, PgClassType.Value, AttributesCursor> {
...
+        order by ${sql.join(pgAttributes.map(pgAttribute =>
+          sql.query`${sql.identifier(pgAttribute.name)} using ${sql.raw((last != null ? !descending : descending) ? '>' : '<')}`,
+        ), ', ')}
+
+        -- Finally, apply the appropriate limit.
+        limit ${first != null ? sql.value(first) : last != null ? sql.value(last) : sql.raw('all')}
+
+        -- If we have an offset, add that as well.
+        ${_offset != null ? sql.query`offset ${sql.value(_offset)}` : sql.query``}
+      ) select to_json(${sql.identifier(aliasIdentifier2)}) as value from ${sql.identifier(aliasIdentifier2)};
```
calebmer: Could we add a comment here explaining why we have the `WITH` clause? My fear is that in the future we will be looking to migrate away from `to_json` (which could be sooner rather then later with https://github.com/calebmer/postgraphql/pull/342!) and we won’t know why the `WITH` is hear and we’ll just keep the `WITH`. It also helps to explain to future contributors why it is there.

src/postgres/inventory/paginator/PgPaginatorOrderingOffset.ts
```diff
@@ -121,17 +121,20 @@ implements Paginator.Ordering<TInput, TItemValue, OffsetCursor> {
...
-      offset ${sql.value(offset)}
-      limit ${limit != null ? sql.value(limit) : sql.query`all`}
+      with ${sql.identifier(aliasIdentifier2)} as (
+        select ${sql.identifier(aliasIdentifier)}.*
+        from ${fromSql} as ${sql.identifier(aliasIdentifier)}
+        where ${conditionSql}
+        ${this.orderBy ? sql.query`order by ${this.orderBy}` : sql.query``}
+        offset ${sql.value(offset)}
+        limit ${limit != null ? sql.value(limit) : sql.query`all`}
+      ) select to_json(${sql.identifier(aliasIdentifier2)}) as value from ${sql.identifier(aliasIdentifier2)};
```
calebmer: Could we get a comment here too?

src/postgres/inventory/paginator/PgPaginatorOrderingAttributes.ts
```diff
@@ -66,37 +66,43 @@ implements Paginator.Ordering<TInput, PgClassType.Value, AttributesCursor> {
...
-      limit ${first != null ? sql.value(first) : last != null ? sql.value(last) : sql.raw('all')}
-
-      -- If we have an offset, add that as well.
-      ${_offset != null ? sql.query`offset ${sql.value(_offset)}` : sql.query``}
+      -- The query is wrapped with a CTE on which the to_json function is
+      -- applied. This ensures the to_json function is only called on the final
+      -- results of the query.
+      with ${sql.identifier(cteIdentifier)} as (
+        -- The standard select/from clauses up top.
+        select ${sql.identifier(aliasIdentifier)}.*
```
calebmer: I think I know how to fix the failing tests. Try changing this to:

```
select ${sql.identifier(aliasIdentifier)} as value
```

src/postgres/inventory/paginator/PgPaginatorOrderingAttributes.ts
```diff
@@ -66,37 +66,43 @@ implements Paginator.Ordering<TInput, PgClassType.Value, AttributesCursor> {
...
+        order by ${sql.join(pgAttributes.map(pgAttribute =>
+          sql.query`${sql.identifier(pgAttribute.name)} using ${sql.raw((last != null ? !descending : descending) ? '>' : '<')}`,
+        ), ', ')}
+
+        -- Finally, apply the appropriate limit.
+        limit ${first != null ? sql.value(first) : last != null ? sql.value(last) : sql.raw('all')}
+
+        -- If we have an offset, add that as well.
+        ${_offset != null ? sql.query`offset ${sql.value(_offset)}` : sql.query``}
+      ) select to_json(${sql.identifier(cteIdentifier)}) as value from ${sql.identifier(cteIdentifier)};
```
calebmer: Then change this here to:

```
select to_json(${sql.identifier(cteIdentifier)}.value) as value from ${sql.identifier(cteIdentifier)}
```

calebmer: Of course you should also probably test that this change does not break the performance. If it does then we may need to special case times when a query does not return a composite type and instead just returns a scalar type.

benjie: It does break the performance, sadly: https://gist.github.com/benjie/02df5fdbd7ee8315712fff83007f3cd1

benjie: More specifically:

```sql
select sketch_integration_event_run_step.* from sketch_integration_event_run_step order by type limit 10;
 Planning time: 0.084 ms
 Execution time: 556.536 ms

select sketch_integration_event_run_step from sketch_integration_event_run_step order by type limit 10;
 Planning time: 0.089 ms
 Execution time: 7939.289 ms
(7 rows)
```

calebmer: Wow. Perhaps we will have to special case this then.

benjie:
This is still failing the tests on `intSetQuery` - I have the same issue on #342 where I am making heavy use of CTEs. I believe the issue is that each row is returned as an object with one property, the value for which is an integer. I believe if you were to detect this case (where the function returns a setof scalars) you could handle it by mapping `const rows = rawRows.map(row => Object.values(row)[0])`.

danielbuechele:
I added the requested changes and removed the usage of `.*` which seemed to be a performance issue in some cases.

benjie:
Does this give a quantifiable performance improvement? My tests previously suggested it wasn't just the to_json that caused the issue, but also the referencing the table without the `.*` (which seemed to be processed before the filter rather than after, as is the case when an index is present):

https://gist.github.com/benjie/02df5fdbd7ee8315712fff83007f3cd1 explain.sql; i.e.:

with `.*`:

```
[test] # explain analyse select sketch_integration_event_run_step.* from sketch_integration_event_run_step order by type limit 10;
  | QUERY PLAN
  | ---------------------------------------------------------------------------------------------------------------------------------------------------------
  | Limit  (cost=79913.03..79913.06 rows=10 width=807) (actual time=556.499..556.500 rows=10 loops=1)
  | ->  Sort  (cost=79913.03..82010.18 rows=838859 width=807) (actual time=556.497..556.498 rows=10 loops=1)
  | Sort Key: type
  | Sort Method: top-N heapsort  Memory: 26kB
  | ->  Seq Scan on sketch_integration_event_run_step  (cost=0.00..61785.59 rows=838859 width=807) (actual time=0.019..219.488 rows=838708 loops=1)
  | Planning time: 0.084 ms
  | Execution time: 556.536 ms
  | (7 rows)
  |  
  | Time: 556.919 ms
```

without `.*`:

```
  | [test] # explain analyse select sketch_integration_event_run_step from sketch_integration_event_run_step order by type limit 10;
  | QUERY PLAN
  | ----------------------------------------------------------------------------------------------------------------------------------------------------------
  | Limit  (cost=79913.03..79913.06 rows=10 width=837) (actual time=7939.256..7939.259 rows=10 loops=1)
  | ->  Sort  (cost=79913.03..82010.18 rows=838859 width=837) (actual time=7939.255..7939.258 rows=10 loops=1)
  | Sort Key: type
  | Sort Method: top-N heapsort  Memory: 26kB
  | ->  Seq Scan on sketch_integration_event_run_step  (cost=0.00..61785.59 rows=838859 width=837) (actual time=0.033..7150.607 rows=838708 loops=1)
  | Planning time: 0.089 ms
  | Execution time: 7939.289 ms
  | (7 rows)
  |  
  | Time: 7939.653 ms
```

danielbuechele:
In my case (using row level security) the `.*` operator doesn't make any difference. However using CTEs over not using them makes a big difference.

**no `.*`, no CTE**
```
Limit  (cost=31989.50..31989.63 rows=50 width=62) (actual time=3917.921..3917.937 rows=50 loops=1)
  ->  Sort  (cost=31989.50..32277.03 rows=115009 width=62) (actual time=3917.920..3917.934 rows=50 loops=1)
        Sort Key: __local_0__.id
        Sort Method: top-N heapsort  Memory: 161kB
        ->  Subquery Scan on __local_0__  (cost=0.00..28168.99 rows=115009 width=62) (actual time=0.343..3839.865 rows=110587 loops=1)
              ->  Seq Scan on articles __local_0___1  (cost=0.00..26731.38 rows=115009 width=1930) (actual time=0.262..221.573 rows=110587 loops=1)
                    Filter: (status = 'PUBLISHED'::status)
                    Rows Removed by Filter: 6208
Planning time: 2.721 ms
Execution time: 3917.965 ms
```

**with `.*`, no CTE**
```
Limit  (cost=31989.50..31989.63 rows=50 width=62) (actual time=4030.565..4030.584 rows=50 loops=1)
  ->  Sort  (cost=31989.50..32277.03 rows=115009 width=62) (actual time=4030.563..4030.579 rows=50 loops=1)
        Sort Key: __local_0__.id
        Sort Method: top-N heapsort  Memory: 161kB
        ->  Subquery Scan on __local_0__  (cost=0.00..28168.99 rows=115009 width=62) (actual time=0.451..3947.324 rows=110587 loops=1)
              ->  Seq Scan on articles __local_0___1  (cost=0.00..26731.38 rows=115009 width=1930) (actual time=0.374..244.449 rows=110587 loops=1)
                    Filter: (status = 'PUBLISHED'::status)
                    Rows Removed by Filter: 6208
Planning time: 3.815 ms
Execution time: 4030.613 ms
```

**no `.*`, with CTE**
```
CTE Scan on test  (cost=31702.11..31703.23 rows=50 width=32) (actual time=201.214..203.636 rows=50 loops=1)
  CTE test
    ->  Limit  (cost=31701.98..31702.11 rows=50 width=1876) (actual time=201.107..201.133 rows=50 loops=1)
          ->  Sort  (cost=31701.98..31989.50 rows=115009 width=1876) (actual time=201.106..201.127 rows=50 loops=1)
                Sort Key: __local_0__.id
                Sort Method: top-N heapsort  Memory: 124kB
                ->  Seq Scan on articles __local_0__  (cost=0.00..26731.38 rows=115009 width=1876) (actual time=0.353..110.675 rows=110587 loops=1)
                      Filter: (status = 'PUBLISHED'::status)
                      Rows Removed by Filter: 6208
Planning time: 4.358 ms
Execution time: 203.685 ms
```

**with `.*`, with CTE**
```
CTE Scan on test  (cost=31702.11..31703.23 rows=50 width=32) (actual time=196.084..197.368 rows=50 loops=1)
  CTE test
    ->  Limit  (cost=31701.98..31702.11 rows=50 width=1930) (actual time=196.015..196.030 rows=50 loops=1)
          ->  Sort  (cost=31701.98..31989.50 rows=115009 width=1930) (actual time=196.013..196.025 rows=50 loops=1)
                Sort Key: __local_0__.id
                Sort Method: top-N heapsort  Memory: 128kB
                ->  Subquery Scan on __local_0__  (cost=0.00..27881.47 rows=115009 width=1930) (actual time=0.313..146.215 rows=110587 loops=1)
                      ->  Seq Scan on articles __local_0___1  (cost=0.00..26731.38 rows=115009 width=1930) (actual time=0.313..138.437 rows=110587 loops=1)
                            Filter: (status = 'PUBLISHED'::status)
                            Rows Removed by Filter: 6208
Planning time: 29.157 ms
Execution time: 197.405 ms
```

So, my suggestion would be with `.*` and with CTE. Should I update the PR accordingly?

benjie:
I strongly suspect if you use `.*` it'll cause issues with intSetQuery. Thanks 🙏

PR #399, opened 2017-03-17
feat(postgraphql): inject custom settings into the database

mlegenhausen:
This pull request adds the `pgSettings` configuration parameter to `postgraphql` middleware function. With this parameter it is possible to set custom config values to the postgres connection. These settings can then be accessed via `current_setting('my.custom.setting')` in all stored procedures. This is analog to the usage of the `jwt.claims` settings but can be defined at middleware initialisation time.

I hope you like it.

### DIFF
--- a/src/postgraphql/withPostGraphQLContext.ts
+++ b/src/postgraphql/withPostGraphQLContext.ts
@@ -144,7 +144,7 @@ async function setupPgClientTransaction ({
   // this prevents an accidentional overwriting
   if (typeof pgSettings === 'object') {
     for (const key of Object.keys(pgSettings)) {
-      localSettings.set(key, pgSettings[key])
+      localSettings.set(key, String(pgSettings[key]))
     }
   }
 
--- a/src/postgraphql/withPostGraphQLContext.ts
+++ b/src/postgraphql/withPostGraphQLContext.ts
@@ -36,11 +36,13 @@ export default async function withPostGraphQLContext(
     jwtToken,
     jwtSecret,
     pgDefaultRole,
+    pgSettings,
   }: {
     pgPool: Pool,
     jwtToken?: string,
     jwtSecret?: string,
     pgDefaultRole?: string,
+    pgSettings?: { [key: string]: mixed },
   },
   callback: (context: mixed) => Promise<ExecutionResult>,
 ): Promise<ExecutionResult> {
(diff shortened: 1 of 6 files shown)


src/postgraphql/withPostGraphQLContext.ts
```diff
@@ -135,6 +140,14 @@ async function setupPgClientTransaction ({
   // Sql query.
   const localSettings = new Map<string, mixed>()
 
+  // Set the custom provided settings before jwt claims and role are set
+  // this prevents an accidentional overwriting
+  if (typeof pgSettings === 'object') {
+    for (const key of Object.keys(pgSettings)) {
+      localSettings.set(key, pgSettings[key])
```
benjie: Please wrap the value with String constructor: localSettings.set(key, String(pgSettings[key]))

mlegenhausen:
A very simple use case is to inject node process related information in your postgres database like the current process id. My specific use case is to make the configuration of my postgres functions configurable from the node process like the configuration of an url prefix which gets concatinated via a  postgres function which gets mapped to a graphql field.

This is the first preparation for a second pull request I maybe come up to allow the generation of the settings via a middleware like function. Example:

```
pgSettingsMiddleware: (req, defaultPgSettings) => ({ ...defaultPgSettings, 'req.ip': req.ip })
```

which can then be accessed via `current_settings('req.ip')`.

Or another approach I have in mind is to open up the middleware stack of `postgraphql` to allow this modification in a more generell way and to allow a much broader modification of the postgraphql data. I am thinking about a multer integration in postgraphql to allow file uploads. But that for the future ;)

src/postgraphql/withPostGraphQLContext.ts
mlegenhausen: I will. Any thoughts about objects that can be serialized to a json string and injected in postgres?

benjie: I think the user should handle that themself with JSON.serialize. I just want to ensure that the types are clear so that `set_config(text, text, bool)` isn't mistaken for a potential different user-defined function such as `set_config(test, json, bool)` etc.

benjie:
Merged 🎉  - thanks @mlegenhausen

calebmer:
Released in 3.1.0 🎉

PR #403, opened 2017-03-23
Expose jwt.verify options parameter to the api and cli

mlegenhausen:
Here is my current state of work. I will add some unit tests after the API changes are confirmed by you. Currently the CLI parameter is named `jwt-audience` other suggestions for a name?


EyMaddis:
I would suggest jwt.audience as it could be the audience setting on a jwt options object. 
This would be a direct mapping between the cli parameter and settings in JavaScript, especially for further settings

mlegenhausen:
This would introduce a namespacing as first the parameter for the cli options. Would be then an `jwt.secret` option the consequence? Just for future improvements of the cli.

EyMaddis:
> Would be then an jwt.secret option the consequence?

I would propose that, yes.
Even if this could be considered a breaking change as the options were previously named `jwtSecret`.

I think it is worth it, as the cli arguments could be grouped into an object like this
```
const postgraphqlSettings = {
  jwt: {
    secret: 'xyz',
    audience: 'postgraphql',
    futureSetting: ....
  }
}
```

with the settings  `--jwt.secret xyz --jwt.audience postgraphql ...`

I love it when the CLI parameters of a function directly map to the option keys when using it as a library. A great example is [Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/2.3/setup-configuration.html) where the configuration is a JSON/YAML object and it is possible to set every value via its traversal path in env variables.

Just my two cents. In the long term, this could lead to a new convention for CLI parameters. Could be a bit too much opinion for such a simple PR (sorry).


Calling @calebmer for an opinion.

benjie:
I think we can support both rather than having a breaking change, and simply deprecate the old flag.

mlegenhausen:
@EyMaddis I am totally on your side this is how the api should look like. If @benjie is ok with that I would implement it this way and guarantee backwards compability. Should we add deprecation output?

For long term there should be external JSON and environment variable configuration possibilities. The latter one is especially useful for docker.

EyMaddis:
yargs supports that kind of grouping: http://yargs.js.org/docs/#parsing-tricks-dot-notation
But I think it breaks with the current cli parameters

benjie:
I'm not really going to weigh in here as I bypass the JWT stuff in my stack anyway... Just noting it should be possible to say `if (args.jwtSecret) { args.jwt = args.jwt || {}; args.jwt.secret = args.jwt.secret || args.jwtSecret }` to maintain backwards compatibility.

mlegenhausen:
New proposed backwards compatible cli style added. All JWT parameters have now a `jwt` prefix. This includes also the `secret` parameter. For backwards compatibility there is also the old `secret` parameter which I marked as deprecated in the help text.

There is currently no test file for the CLI. So is manuell testing enough or should I try to automate this also?

@benjie @calebmer if you are fine with the API? I would go on and add all missing CLI parameters for the JWT options and would add all needed tests.

angelosarto:
@mlegenhausen This looks great -- I am using postgraphql completely vanilla from the command line, and right now my only block is the audience check; so this is awesome for me personally.  I think the deprecated flag makes sense.  

I didn't see anything in the PR that didn't make sense, but I am not sure I have the necessary depth in the project yet (or write access) to be the reviewer.

calebmer:
@EyMaddis 

> I think it is worth it, as the cli arguments could be grouped into an object like this
>
> ```js
> const postgraphqlSettings = {
>   jwt: {
>     secret: 'xyz',
>     audience: 'postgraphql',
>     futureSetting: ....
>   }
> }
> ```
>
> with the settings --jwt.secret xyz --jwt.audience postgraphql ...

I like the idea! But if we are going to move in this direction I want to see the style applied consistently. So perhaps we don’t add that style to this PR, but then we can follow up with a second PR that introduces a better way to configure PostGraphQL with a better mapping to CLI options. (and perhaps a configuration file as well as a result?)

For this PR let’s just call cli option `--jwt-audience`. Yes it means we’ll have to deprecate that as well (if we do a release in between this and the new config style), but that’s something I can live with 😊

* * *

@mlegenhausen what’s the reasoning behind letting the user configure all of the `VerifyOptions`? Generally I like to mask as many implementation details as possible, so in this case it would mean perhaps allowing users to configure the `audience` and `algorithms` for instance, but not permitting any other options. I like to do this because it gives us the freedom to swap out our underlying implementation at any time. Otherwise we are tied to supporting the past and future `jsonwebtoken` API. Whitelisting options also helps us with documentation. Admittedly the likelihood of us switching out the `jsonwebtoken` package is _very_ low, but it’s a principle I like to apply 😊

I’m not entirely against using `VerifyOptions`, I just want to hear an argument for it, and/or decide if perhaps we only want to expose one or two options instead of all the options and any future options `jsonwebtoken` adds.

Also, there are some options like `jwtId` that perhaps should be set on a request-by-request basis, and `algorithms` which perhaps shouldn’t be configurable because currently we only allow a string secret 😣

src/postgraphql/cli.ts
```diff
@@ -31,13 +33,15 @@ program
...
   .option('-o, --cors', 'enable generous CORS settings. this is disabled by default, if possible use a proxy instead')
   .option('-a, --classic-ids', 'use classic global id field name. required to support Relay 1')
   .option('-j, --dynamic-json', 'enable dynamic JSON in GraphQL inputs and outputs. uses stringified JSON by default')
   .option('-M, --disable-default-mutations', 'disable default mutations, mutation will only be possible through Postgres functions')
   .option('-l, --body-size-limit <string>', 'set the maximum size of JSON bodies that can be parsed (default 100kB) The size can be given as a human-readable string, such as \'200kB\' or \'5MB\' (case insensitive).')
+  .option('--secret <string>', 'DEPRECATED: See jwt.secret')
+  .option('-e, --jwt.secret <string>', 'the secret to be used when creating and verifying JWTs. if none is provided auth will be disabled')
+  .option('-A, --jwt.audience <list>', 'a comma separated list of audiences your jwt token can contain. If no audience is given the audience defaults to `postgraphql`', assign('audience', list), jwtOptions)
```
calebmer: Could we build something a little more custom here for parsing? Like perhaps what we do for the `--schema` option:

```js
(option: string) => ({ audience: option.split(', ') })
```

It’s just a little shorter and we don’t have to define functions we may never re-use 😊

src/postgraphql/cli.ts
```diff
@@ -31,13 +33,15 @@ program
...
   .option('-b, --disable-graphiql', 'disables the GraphiQL interface. overrides the GraphiQL route option')
-  .option('-e, --secret <string>', 'the secret to be used when creating and verifying JWTs. if none is provided auth will be disabled')
   .option('-t, --token <identifier>', 'the Postgres identifier for a composite type that will be used to create tokens')
   .option('-o, --cors', 'enable generous CORS settings. this is disabled by default, if possible use a proxy instead')
   .option('-a, --classic-ids', 'use classic global id field name. required to support Relay 1')
   .option('-j, --dynamic-json', 'enable dynamic JSON in GraphQL inputs and outputs. uses stringified JSON by default')
   .option('-M, --disable-default-mutations', 'disable default mutations, mutation will only be possible through Postgres functions')
   .option('-l, --body-size-limit <string>', 'set the maximum size of JSON bodies that can be parsed (default 100kB) The size can be given as a human-readable string, such as \'200kB\' or \'5MB\' (case insensitive).')
+  .option('--secret <string>', 'DEPRECATED: See jwt.secret')
```
calebmer: Let’s deprecate this and introduce the `.` style in another PR. There are a couple of things we could apply this style to like the `export-*` properties 😊

src/postgraphql/http/createPostGraphQLHttpRequestHandler.d.ts
```diff
@@ -48,6 +49,10 @@ export default function createPostGraphQLHttpRequestHandler (config: {
   // value.
   jwtSecret?: string,
 
+  // Custom options given to jwt.verify function. When no custom options were given
+  // the default audience will be set to 'postgraphql'
+  jwtOptions?: VerifyOptions,
```
calebmer: It would be more accurate to say when no `audience` property is given. Unless I’m mistaken about the implementation? 😊

mlegenhausen:
@calebmer Thanks for your feedback. I will roll back to `jwt-audience` as cli parameter and will stick with it. It is the most important option for JWT configuration. Personally I have no use case for the other parameters and they can be easily be added afterwards. You are absolutely right that is always the best to only support what is currently needed. I simply stepped again in the "you ain't gonna need it" trap were using an existing interface seemed obvious :wink:

So lets keep this pull request small and focused on the audience and having a new configuration format would be great.

src/postgraphql/cli.ts
```diff
@@ -39,9 +37,9 @@ program
   .option('-j, --dynamic-json', 'enable dynamic JSON in GraphQL inputs and outputs. uses stringified JSON by default')
   .option('-M, --disable-default-mutations', 'disable default mutations, mutation will only be possible through Postgres functions')
   .option('-l, --body-size-limit <string>', 'set the maximum size of JSON bodies that can be parsed (default 100kB) The size can be given as a human-readable string, such as \'200kB\' or \'5MB\' (case insensitive).')
-  .option('--secret <string>', 'DEPRECATED: See jwt.secret')
-  .option('-e, --jwt.secret <string>', 'the secret to be used when creating and verifying JWTs. if none is provided auth will be disabled')
-  .option('-A, --jwt.audience <list>', 'a comma separated list of audiences your jwt token can contain. If no audience is given the audience defaults to `postgraphql`', assign('audience', list), jwtOptions)
+  .option('secret <string>', 'DEPRECATED: Use jwt-secret instead')
```
calebmer: Looks like the `--` was removed in `--secret`. Is that intentional?

src/postgraphql/cli.ts
```diff
@@ -39,9 +37,9 @@ program
...
+  .option('-e, --jwt-secret <string>', 'the secret to be used when creating and verifying JWTs. if none is provided auth will be disabled')
+  .option('-A, --jwt-audience <list>', 'a comma separated list of audiences your jwt token can contain. If no audience is given the audience defaults to `postgraphql`', (option: string) => option.split(','))
```
calebmer: Maybe instead of `<list>` here we just do `<string>`? I don’t people to mistakenly think that they have to put in a list 😣

We do this for `--schema` as well.

src/postgraphql/cli.ts
```diff
@@ -117,7 +116,9 @@ const server = createServer(postgraphql(pgConfig, schemas, {
   graphiql: !disableGraphiql,
   jwtPgTypeIdentifier,
   jwtSecret: jwtSecret || deprecatedJwtSecret,
-  jwtOptions,
+  jwtOptions: {
+    audience: jwtAudience,
```
calebmer: Could we flatten this out for now into just `jwtAudience` and not `jwtOptions: { audience }`? In the future we could put all of the jwt settings (for which there are more then a few couple…) in a `jwtOptions` property. It also removes the need for the type.

calebmer:
Hmm, my GitHub review comments are rendering as outdated for me, but the code looks the same in https://github.com/postgraphql/postgraphql/pull/403/files

If you could also just really quickly run `--help` and paste the output into https://github.com/postgraphql/postgraphql/blob/722eb0d01c8dff87daeaeec844be67211a7290dd/docs/cli.md so that the documentation stays up-to-date that would be great! 😊

src/postgraphql/cli.ts
mlegenhausen: Good catch. Of cause not.

src/postgraphql/cli.ts
mlegenhausen: Sure just used the example from the commander website.

mlegenhausen:
@calebmer Yeah I forgot to mark the `secret` as deprecated and forced pushed.

src/postgraphql/cli.ts
```diff
@@ -110,8 +114,11 @@ const server = createServer(postgraphql(pgConfig, schemas, {
   graphqlRoute,
   graphiqlRoute,
   graphiql: !disableGraphiql,
-  jwtSecret,
   jwtPgTypeIdentifier,
+  jwtSecret: jwtSecret || deprecatedJwtSecret,
+  jwtOptions: {
```
calebmer: Weren’t we going to flatten this out into just `jwtAudience`?

mlegenhausen: I thought I only should flatten the CLI options. I changed it again.

mlegenhausen:
I renamed the attribute to `jwtAudiences` cause it is defined as `Array<string>`.

docs/library.md
```diff
@@ -58,6 +58,7 @@ Arguments include:
   - `graphiqlRoute`: The endpoint the GraphiQL query interface will listen on (**NOTE:** GraphiQL will not be enabled unless the `graphiql` option is set to `true`). Defaults to `/graphiql`.
   - `pgDefaultRole`: The default Postgres role to use. If no role was provided in a provided JWT token, this role will be used.
   - `jwtSecret`: The secret for your JSON web tokens. This will be used to verify tokens in the `Authorization` header, and signing JWT tokens you return in procedures.
+  - `jwtOptions`: The options for the `jwt.verify(jwtToken, jwtSecret, jwtOptions)` call. See the [jsonwebtoken documentation](https://github.com/auth0/node-jsonwebtoken#jwtverifytoken-secretorpublickey-options-callback) for further details.
```
calebmer: Awesome! I just think this needs to be updated and I’ll merge and release ASAP 😊

calebmer:
Alright, thanks so much @mlegenhausen! Sorry this took a bit 😊. I’ll merge and release once CI passes 👍

calebmer:
Released in `3.2.0` 🎉

angelosarto:
Has anyone else tried the new command line options using the package on npm?

As far as I can tell the new options aren't working.  The package appears to be the correct version and the new options are in the help file:
```
postgraphql --version && postgraphql --help
3.2.0
  Usage: postgraphql [options...]
<SNIP...>
    --secret <string>                DEPRECATED: Use jwt-secret instead
    -e, --jwt-secret <string>        the secret to be used when creating and verifying JWTs. if none is provided auth will be disabled
    -A, --jwt-audiences <string>     a comma separated list of audiences your jwt token can contain. If no audience is given the audience defaults to `postgraphql`
```

However running it doesn't seem to work, if you supply the deprecated `--secret` option it works ok:
```
postgraphql --connection postgres://ls_external:ls_external_password@127.0.0.1:32002/local_dev_db --secret local_secret
```

However, changing this to the short form `-e` or to the new long form `--jwt-secret` like this:
```
postgraphql --connection postgres://ls_external:ls_external_password@127.0.0.1:32002/local_dev_db --jwt-secret local_secret
```
And then attempting to supply the authorization header results in this which I am pretty sure is what happens if no secret is configured.
```
"errors": [
    {
      "message": "Not allowed to provide a JWT token."
    }
  ]
```

Can someone else confirm what I am seeing?

Zeppelin456:
Yes, I get exactly the same :  "errors": [
    {
      "message": "Not allowed to provide a JWT token."
    }
  ]

benjie:
Ref: #437

PR #407, opened 2017-03-25
fix(postgres): use literal syntax for empty arrays

### LINKED ISSUE #406: Error: cannot determine type of empty array
When I call the following mutation with an empty array, I get an error from postgres about an untyped array.

```sql
CREATE TABLE graphql.role (
  role_name  text PRIMARY KEY,
  operations text[] NOT NULL
);
```

```graphql
mutation CreateRole($operations: [String!]!) {
  createRole(input: {role: {roleName: "chris", operations: $operations}}) {
    role {
      roleName
      operations
    }
  }
}
```

```
STATEMENT:  with __local_0__ as ( insert into "graphql"."role" ("role_name", "operations") values ($1, array[]) returning * ) select row_to_json(__local_0__) as object from __local_0__
ERROR:  cannot determine type of empty array at character 92
HINT:  Explicitly cast to the desired type, for example ARRAY[]::integer[].
```

Using an array literal works well in `psql`:

```
psql> with __local_0__ as ( insert into "graphql"."role" ("role_name", "operations") values ('chris', '{}') returning * ) select row_to_json(__local_0__) as object from __local_0__;
                object                 
---------------------------------------
 {"role_name":"chris","operations":[]}
(1 row)
```

When I call the mutation with a non-empty array, I can see the generated query is

```sql
with __local_0__ as ( insert into "graphql"."role" ("role_name", "operations") values ($1, array[$2, $3]) returning * ) select row_to_json(__local_0__) as object from __local_0__
```

Is it possible for `postgraphql` to use `pg` to serialize arrays?

(I'm willing, though perhaps not able, to contribute a fix. Where should I look for the generation of this query?)

cbandy:
Fixes #406


cbandy:
I wasn't sure where to add a test for this. I've verified that it solves the specific scenario in #406.

cbandy:
It did work, though Travis did not like when I [updated the test snapshot](https://travis-ci.org/cbandy/postgraphql/jobs/215078801). I've removed it from the amended commit.

calebmer:
Yep. You’ll need to update the snapshot on your end. Jest is nice with these snapshot tests because after you update we can review the diff of the snapshot to make sure its what we expect 😊

To update the snapshots run:

```bash
./scripts/test -u
```

cbandy:
Done. (Something about Vim swapfiles was messing with the snapshot on my end.)

calebmer:
Released in 3.1.0 🎉

PR #408, opened 2017-03-25
Make the tests faster

calebmer:
Some of the test suites are pretty slow, so this PR aims to speed up our tests. With this PR tests go from 42 seconds to 12 seconds on my machine 🎉

The secret to pretty much all of the perf wins is parallelization. Jest does not automatically parallelize tests which is fine when testing synchronous code, but is not ok for some of the heavy asynchronous tests we end up running.

We were also repeating a lot of introspection work so I tried to get clients and introspections to be shared as much as possible.


calebmer:
It looks like tests on CI went from 30s to 10s.

benjie:
Nice. Looks like we can't use snapshot tests when we're expecting errors though:

```
 FAIL  src/postgraphql/__tests__/postgraphqlIntegrationQueries-test.js
  ● dynamic-json.graphql
    
    - Snapshot
    + Received
    
    @@ -1,5 +1,5 @@
     Object {
       "errors": Array [
    -    [SyntaxError: Unexpected token h in JSON at position 0],
    +    [SyntaxError: Unexpected token h],
       ],
     }
```

src/__tests__/utils/createTestInParallel.ts
```diff
@@ -0,0 +1,31 @@
...
+    testFns.push(fn)
+
+    test(name, async () => {
+      // If the tests have not yet been run then run all of our tests.
+      if (!testResults) {
+        testResults = testFns.map(testFn => Promise.resolve(testFn()))
+      }
+      // Await the result.
+      await testResults[index]
+    })
```
benjie: Nice.

calebmer:
Snapshot tests are still fine. That one snapshot test failed because it looks like the error message is different in Node.js 4 then in Node.js 6, however seeing that helped me catch a bug I forgot to fix 😊

The second error causing the tests to crash is because Koa ships some ES2015 syntax in its source code that Node.js 4 doesn’t like. So the fix was to not test Koa if the Node.js major version is less than or equal to 4.

PR #409, opened 2017-03-26
Add missing type coercion in computed procedure field

### LINKED ISSUE #390: 3.0.0: error on custom functions
First of all, great library, thanks for all of the great work.

There seems to be an error that started occurring in the 3.0.0 branch related to custom functions. 

How to reproduce this issue:
## 1. create the tables, functions and populate the database
```SQL
# A first table
create table public.foo (
    id serial not null primary key,
    name text
);

# A second table
create table public.bar (
    id serial not null primary key,
    created_at timestamp default now(),
    foo_id int not null references public.foo(id)
);

# A function that returns the last item of the second table
create function public.foo_last_bar(foo public.foo) returns bar as $$
    select *
    from bar
    where bar.foo_id = foo.id
    order by created_at desc
    limit 1
$$ language sql stable;

# Some data
insert into foo (name) values ('postgraphql');
insert into bar (foo_id) values (1);
insert into bar (foo_id) values (1);
```

## 2. Execute a query in GraphiQL:
```GQL
fooById(id: 1) {
    lastBar{id}
}
```

## 3.A: The 3.0.0 branches' output:
```
{
  "data": {
    "fooById": {
      "lastBar": null
    }
  },
  "errors": [
    {
      "message": "value.get is not a function",
      "locations": [
        {
          "line": 2,
          "column": 14
        }
      ],
      "path": [
        "fooById",
        "lastBar",
        "id"
      ]
    }
  ]
}
```

## 3.B: The 2.6.0 branches' output:
```
{
  "data": {
    "fooById": {
      "lastBar": {
        "id": 2
      }
    }
  }
}
```

Any idea what might be causing this?

calebmer:
Fixes #390

This PR builds off of https://github.com/postgraphql/postgraphql/pull/408 because I wanted the tests to be faster as I fixed this issue 😊

Here is the diff that actually fixes the issue: https://github.com/postgraphql/postgraphql/commit/02973986af0cf65773d482a935c6a00da80d04d9

Specifically this line: https://github.com/postgraphql/postgraphql/commit/02973986af0cf65773d482a935c6a00da80d04d9#diff-68e6e47f66f4641b59bfbd00beff36acR62

When doing the types refactor I forgot to add a coercion at that point from the Postgres value to the internal PostGraphQL value 😣. This PR makes sure that coercion happens. I’m surprised there was no test case for this, so a test case has also been added.


benjie:
Merge when it passes CI

PR #413, opened 2017-03-27
Add video and add section about watch mode

calebmer:
I gave a talk on PostGraphQL at GraphQL Summit which I am pretty proud of, so we should probably add that to the readme.

Also, PostGraphQL has a pretty advanced watch mode that provides the best GraphQL API development experience in the entire GraphQL ecosystem. We should probably mention that 😉


calebmer:
To see how the changes rendered, check out: https://github.com/postgraphql/postgraphql/tree/update-readme

PR #433, opened 2017-04-09
fix(postgraphql): don't read auth header if auth not enabled

nbushak:
The documentation says "And if you don’t want authorization, just don’t set the --secret argument and PostGraphQL will ignore all authorization information!"

It turns out postgraphql _does_ read auth information and throws an error if the format is not as expected. This prevents folks from using the Authorization header for non-JWT, like Basic auth.


nbushak:
Ah actually I think I need to do something a bit different here, if we want this code to still work:
https://github.com/postgraphql/postgraphql/blob/d4fd6a4009fea75dbcaa00d743c985148050475e/src/postgraphql/withPostGraphQLContext.ts#L121

I'll make the change differently and resubmit.

PR #436, opened 2017-04-11
fix(graphqli): pass along credentials in fetcher

nbushak:
I'm using basic authentication on my postgraphql endpoint. This configures fetch() to pass along the browser's credentials, which seems to be the norm for other graphiql users like express-graphql:
https://github.com/graphql/express-graphql/blob/master/src/renderGraphiQL.js#L114

Based on my reading of the code, it doesn't appear that `jwtToken` is ever passed in, so I didn't test passing that in. But changing "credentials" to "include" shouldn't result in it being overwritten if it is passed in, according to the fetch spec:
https://fetch.spec.whatwg.org/#http-network-or-cache-fetch
"If httpRequest’s header list contains `Authorization`, then terminate these substeps."

### DIFF
--- a/src/postgraphql/graphiql/src/components/PostGraphiQL.js
+++ b/src/postgraphql/graphiql/src/components/PostGraphiQL.js
@@ -61,6 +61,7 @@ class PostGraphiQL extends React.Component {
       }, jwtToken ? {
         'Authorization': `Bearer ${jwtToken}`,
       } : {}),
+      credentials: 'same-origin',
       body: JSON.stringify(graphQLParams),
     })
 


benjie:
I'm often faced with this issue also (I use `same-origin` rather than `include` though); however I don't think it should be on by default as that seems to be bad practice. It would be nice to make it a config option though, so you can pass e.g. `--graphiql-credentials=same-origin` on the CLI and have it enabled.

calebmer:
Instead of adding a bunch of CLI options we should alternatively consider how to allow users to provide their own custom `fetcher` function to GraphiQL. Or we should encourage users to build their own copy of GraphiQL which is not too hard. You would just miss out on some of PostGraphQL’s watch features.

The latter option of using your own GraphiQL instance might be the best idea because at some point I want to add a JWT input to the PostGraphQL GraphiQL UI which may conflict with people’s non-JWT auth implementations.

We could also publish the PostGraphQL GraphiQL UI as a separate package that you could configure.

calebmer:
This one change is pretty simple and small in scope. What is the downside to merging this?

benjie:
I'd be happy to merge it, does it open any XSS vulnerabilities?

nbushak:
I can change this to `same-origin` instead. While I don't think this opens any XSS vulnerabilities `same-origin` definitely shouldn't. Does that work?

benjie:
Works for me; I'd love this change. (I've been manually patching the GraphiQL compiled file in node_modules ☺)

Issue #300, opened 2017-01-06
Expose hooks to enable extending the GraphQL schema


cellis:
Let's say I have a query and I need to call out to some api (either internal or external) to fulfill it? What's the best way to accomplish this? Does postgraphql expose resolve hooks? For some things, obviously I can precompute the data by saving it to postgres, but i can forsee problems where it wouldn't work.

benjie:
We don't currently support doing that through postgraphql, though if you understand the caveats you could try doing it through postgresql directly using foreign data wrappers. Definitely something that we have in mind through.

idibidiart:
Very new to GraphQL so this may not make sense but I assume in all these types of SQL-to-GraphQL-Schema solutions that we get an actual GraphL schema generated statically and as such we should be able to extend that schema with new types and resolvers etc. Or does postgraphql implement its own graphql server that consumes the database schema directly and will ignore or throw in case the graphql schema does not map 1:1 to the db schema?

danvk:
It would be great to have some hooks which make this possible. It's the difference between postgraphql being an all-or-nothing solution and one which can fit into a larger system.

@zamiang and my use case: we have a database which contains our own data as well as user IDs from an external service which exposes a REST API. We'd like to expose a unified GraphQL API over both services. To fill in data from the external REST API, we'd need to issue a second request after the Postgres result comes back.

I imagine defining the GraphQL schema for something like this would be tricky. @benjie any thoughts on a good workaround for this now? We're thinking about having two independent GraphQL endpoints: one for our own data with PostGraphQL and another which wraps the external REST API. Not ideal and not very GraphQLy, but it will let us keep moving.

benjie:
I'm afraid I can't give any solid advice on that; I've just added a couple tiny REST APIs where necessary for my requirements. You might try reading up on merging GraphQL APIs (I've no idea if this will be successful or not):

https://www.google.co.uk/search?q=combine+graphql+schemas&oq=combine+graphql+schemas

https://gist.github.com/icebob/553c1f9f1a9478d828bcb7a08d06790a

https://gist.github.com/voodooattack/b0484625a90191a4dde4aabae8884243

cellis:
[...]

PR #448, opened 2017-04-21
Adding schema injection

mikeburgh:
Looking for some feedback at this stage on the approach, as a way to solve #300 

Still todo
- Write some tests
- Clean up loadInjections.ts
- Validate the injections to ensure they are the right format (eg contain a schema).

You can load injections using a config param/CLI option that contains the path (supports globs) to your additional schema you want to inject: --schema-injection './injections/**/*.js'

Each file matched is exported to export the following format as an object: 

```javascript
{
    name: "testMutation1",
    type: "mutation",
    schema: {... a valid GraphQL Schema and resolver}
}
```

The type can be either a mutation or a query, and will be injected into the appropriate part of the GraphQL schema.

### DIFF
--- a/src/graphql/schema/getMutationGqlType.ts
+++ b/src/graphql/schema/getMutationGqlType.ts
@@ -1,5 +1,5 @@
 import { GraphQLObjectType, GraphQLFieldConfig } from 'graphql'
-import { buildObject, memoize1 } from '../utils'
+import { buildObject, memoize1, loadInjections } from '../utils'
 import BuildToken from './BuildToken'
 import createCollectionMutationFieldEntries from './collection/createCollectionMutationFieldEntries'
 
@@ -37,6 +37,9 @@ function createMutationGqlType (buildToken: BuildToken): GraphQLObjectType | und
             .getCollections()
             .map(collection => createCollectionMutationFieldEntries(buildToken, collection))
             .reduce((a, b) => a.concat(b), [])
+    ),
+    ...(loadInjections(options.schemaInjection, 'mutation')
+
     ),
   ]
 
--- a/src/graphql/schema/getMutationGqlType.ts
+++ b/src/graphql/schema/getMutationGqlType.ts
@@ -38,9 +38,7 @@ function createMutationGqlType (buildToken: BuildToken): GraphQLObjectType | und
             .map(collection => createCollectionMutationFieldEntries(buildToken, collection))
             .reduce((a, b) => a.concat(b), [])
     ),
-    ...(loadInjections(options.schemaInjection, 'mutation')
-
-    ),
+    ...(loadInjections(options.schemaInjection, 'mutation')),
   ]
 
   // If there are no mutation fields, just return to avoid errors.
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "pg-minify": "^0.4.1",
     "pluralize": "^3.0.0",
     "postgres-interval": "^1.0.2",
+    "require-glob": "^3.2.0",
     "send": "^0.14.1",
     "tslib": "^1.5.0"
   },
--- a/src/graphql/schema/BuildToken.ts
+++ b/src/graphql/schema/BuildToken.ts
@@ -29,6 +29,8 @@ interface BuildToken {
     // If true then the default mutations for tables (e.g. createMyTable) will
     // not be created
     readonly disableDefaultMutations: boolean,
+    // Path to read shcema injections from
+    readonly schemaInjection: string,
   },
   // Hooks for adding custom fields/types into our schema.
   readonly _hooks: _BuildTokenHooks,
--- a/src/graphql/schema/createGqlSchema.ts
+++ b/src/graphql/schema/createGqlSchema.ts
@@ -20,6 +20,8 @@ export type SchemaOptions = {
   // GraphQL types that override the default type generation. Currently this
   // API is private. Use at your own risk.
   _typeOverrides?: _BuildTokenTypeOverrides,
+  // The path to read our injections from
+  schemaInjection?: string,
 }
 
 /**
@@ -36,6 +38,7 @@ export default function createGqlSchema (inventory: Inventory, options: SchemaOp
       nodeIdFieldName: options.nodeIdFieldName || 'nodeId',
       dynamicJson: options.dynamicJson || false,
       disableDefaultMutations: options.disableDefaultMutations || false,
+      schemaInjection: options.schemaInjection || '',
     },
     _hooks: options._hooks || {},
     _typeOverrides: options._typeOverrides || new Map(),
--- a/src/graphql/schema/getQueryGqlType.ts
+++ b/src/graphql/schema/getQueryGqlType.ts
@@ -36,6 +36,7 @@ function createGqlQueryType (buildToken: BuildToken): GraphQLObjectType {
         .getCollections()
         .map(collection => createCollectionQueryFieldEntries(buildToken, collection))
         .reduce((a, b) => a.concat(b), []),
+      // Load injections
       loadInjections(options.schemaInjection, 'query'),
       [
         // The root query type is useful for Relay 1 as it limits what fields
--- a/src/graphql/schema/getQueryGqlType.ts
+++ b/src/graphql/schema/getQueryGqlType.ts
@@ -1,5 +1,5 @@
 import { GraphQLObjectType, GraphQLFieldConfig, GraphQLNonNull, GraphQLID } from 'graphql'
-import { buildObject, memoize1 } from '../utils'
+import { buildObject, memoize1, loadInjections } from '../utils'
 import createNodeFieldEntry from './node/createNodeFieldEntry'
 import getNodeInterfaceType from './node/getNodeInterfaceType'
 import createCollectionQueryFieldEntries from './collection/createCollectionQueryFieldEntries'
@@ -36,6 +36,7 @@ function createGqlQueryType (buildToken: BuildToken): GraphQLObjectType {
         .getCollections()
         .map(collection => createCollectionQueryFieldEntries(buildToken, collection))
         .reduce((a, b) => a.concat(b), []),
+      loadInjections(options.schemaInjection, 'query'),
       [
         // The root query type is useful for Relay 1 as it limits what fields
         // can be queried at the top level.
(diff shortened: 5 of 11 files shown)


mgallagher:
Really excited about this!

src/graphql/schema/getMutationGqlType.ts
```diff
@@ -38,6 +38,9 @@ function createMutationGqlType (buildToken: BuildToken): GraphQLObjectType | und
             .map(collection => createCollectionMutationFieldEntries(buildToken, collection))
             .reduce((a, b) => a.concat(b), [])
     ),
+    ...(loadInjections(options.schemaInjection, 'mutation')
+
+    ),
```
benjie: Weird whitespace?

benjie:
This is pretty cool and seems simple (I've not looked at it in depth yet). I feel like the files should export a function that returns the given format rather than returning the format directly because this allows us to pass options (e.g. PostGraphQLOptions) to the injected schemas which can, for example, give it access to the PostgreSQL database.

Or is this already accounted for?

benjie:
Is there another system that does something like this that we can compare to?

mikeburgh:
Thanks for the feedback.. I have not been able to find another GraphQL implementation that lets you inject into the schema, and I cannot think of any other systems that let you do similar. I am sure they are out there, just not something I have come across.

Exporting as a function is a lot better, I will sort that out, and clean this up.

I also want to look at getting access to the decoded/verified JWT in the methods, so the function approach will solve that.

mikeburgh:
@benjie I have cleaned up the loading process, and refactored it to expect a function from the loaded injections, rather than the resolved schema.

In terms of passing something into the schema function to allow for reuse, I passed the main postgraphql object, although not sure that is the best option given it's more of a constructor. To pass in the options object from the CLI might require some larger re working if your okay with that?

A couple of issues I have found: 
- When a custom injection is executed, commands for the transaction and claims are still sent to the DB. This may not be an issue if the injection is going to call into the DB (although getting access to the client from the context is not straight forward), but perhaps we need to let the injections specify if they need that, so in the case they are not going to call the DB directly that overhead is not present.

- The data from the JWT is not exposed on the context provided to the execute method. The role is present though. Any suggestions for the best way to get that jwt data into the resolve method ? I can build out the context object, but again want to check if you have any thoughts first. 

For reference here is a file that can be used as an injection
```javascript
var graphQL = require('graphql');
module.exports = {
  name: "testMutation",
  type: "mutation",
  schema: function (postgraphQL) {
    return {
      args: {
        input: {
          description: 'the input'
          , type: new graphQL.GraphQLNonNull(graphQL.GraphQLString)
        }
      }
      , description: "test"
      , resolve: function (_source, args, context, info) {
       //context at the moment only contains the role (hidden inside is a pgClient though)
        return {i: args.input, t: 're'};
      }
      , type: new graphQL.GraphQLObjectType({
          name: "testMutationPayload"
          , description: "The output of our mutation."
          , fields: {
            i: {type: graphQL.GraphQLString, description: 'Something out'},
            t: {type: graphQL.GraphQLString, description: 'Something else out'},
          }
        }
      )
    }
  }
}

```

benjie:
@mikeburgh Really sorry for the lack of communications on this - I've not had a chance to play with it but definitely extending the schema is something we (and specifically I!) want to do - so I will be taking a look at some point. If I've not got back to you by 7th May, please add a comment below to give me a nudge!

brysgo:
I played with this a little. Would love to figure out a way to use autogenerated types in injected schema.

aaronc:
Hey, I think this is a really cool idea and would be great to have. Would solve some cases I'm trying to handle.

I'm wondering why not implement this with a simple middleware function? This could be done by by adding `middleware?: (schema, pgClient, pgSchemas, options) => Promise<GraphQLSchema>` to `PostGraphQLOptions` and then calling it after right https://github.com/postgraphql/postgraphql/blob/master/src/postgraphql/postgraphql.ts#L138 like this:

```ts
const newGqlSchema = await createPostGraphQLSchema(pgClient, pgSchemas, options)
const enhancedGqlSchema = await middleware(newGqlSchema, pgClient, pgSchemas, options)
exportGqlSchema(enhancedGqlSchema)
```

 I think this would be the simplest and most flexible route.

mikeburgh:
@benjie no problems at all, I have not had a chance to get back to it (hoping this week). I am also keen on extending the schema as I think it gets you a really solid backend without having to do all the usual scaffolding of building a rest api that maps to a bunch of stored procs etc... the only thing missing is input validation, but I see talk about adding that into graphQL (https://github.com/graphql/graphql-js/issues/361)

@brysgo do you mean the other graphQL qureries and mutations that were discovered.. I think once we get the right object passed into the schema generator, you should be able to call those in a similar way to the [custom execution example](https://github.com/postgraphql/postgraphql/blob/master/docs/library.md#custom-execution)

@aaronc my original plan for using this was going to be via the command line, but I have since ended up wrapping it with express... so having the command line capability is not needed for me any more, so I guess it depends what the overall feeling on use case for this is ?

The big issue for me with this still (middleware or command line) is finding a neat way to mark methods as not needing the database and hence preventing a connection being opened to the DB.. in some cases it might make sense to have that, but not all.

brysgo:
@mikeburgh - first I have to say great work so far, though I think it is important to get at the API from inside the injection, I was actually talking about defining fields on existing types and defining an existing type as the return type of an injected query.

aaronc:
@mikeburgh yeah, it makes sense to take a directory string if you want to support command line. I would say maybe the config option could take either a string for cli usage or a function for embedded? The reason I'm suggesting middleware is because this would address some of the cases @brysgo is describing - you would get the full generated schema and be able to modify that if needed and or reuse some of the fields and types elsewhere.

The idea of not opening a pg client if not needed sounds useful. One approach would be to have context contain a function which opens a connection if needed. Would certainly require a bit of refactoring of the current code, but may allow more flexibility. I could see a scenario where a schema is shared between say pg and some other data store and you may want to inject support for that other data store into context.

mikeburgh:
@aaronc I was just checking out the middleware idea, with making the command line a file to require that exports a function (so both library and cli work as middleware). The problem with the suggestion for where to hook up the middleware is that the gqlSchema is already built there and they don't have a way to add to at that stage (https://github.com/graphql/graphql-js/issues/314).

I think we need to actually tie it in here around here (https://github.com/postgraphql/postgraphql/blob/master/src/graphql/schema/createGqlSchema.ts#L44) which then brings back the use of hooks which @calebmer mentioned in #300 I am just not clear on how he sees those being used to actually extend it, and still not sure this would get you access to all the types that were discovered during introspection from the DB

chadfurman:
The way I'd envision a useful middleware layer working would be:

1. I start postgraphql on the command-line and it looks for scripts in a folder and loads them by registering special endpoints (or queries?  or mutations?  Or both depending on config?) based on whatever is specified by the plugin file via an exported variable name or .json config file
2. a call to `/api/plugin` (or a graphQL query for the query/mutation) would hit the script first (like an express endpoint) which would then have the ability to trigger graphQL calls, get their results, and respond.

I think this would be a super clean and simple method of allowing us to write code in JS w/ all the power of libraries, external API calls, etc. while still maintaining the magic that is GraphQL.

chadfurman:
So, on second thought, I believe this is wasted effort in light of https://github.com/postgraphql/postgraphql/blob/master/docs/library.md

I can put a /graphql endpoint in a nodejs API, and then everything Postgraph can't do, my Node API can.

mgallagher:
@chadfurman I think serving PostGraphQL through Node and adding any needed REST endpoints is a great solution for some things, but don't we need a way to actually extend the GraphQL schema generated by PostGraphQL? I thought that was more of the goal here.

chadfurman:
I don't think PostGraphQL should do everything for the user.  I think there's 5 pages of active tickets on GitHub and, from what I can tell, two active developers.  I think the goal of this PR is to just have a way to deliver data from an external api to the user on request -- synchronously, without having to trigger a NOTIFY event, and as part of the same "API"

If, at that point, you want to bring the GraphQL library into your API to keep the request structure consistent, you could do that.

This was a feature that was asked for quite a bit, but I believe it's due to lack of knowledge / understanding.  That was certainly the case for me -- I don't expect Postgres to ever be able to run certain JS events, and PostGraph is about auto-generating a GraphQL API by introspecting schemas in a database.  At this point, I don't see why PostgraphQL would need to run any additional JS on the side -- injections, middleware, whatever -- because PostgraphQL itself can be the middleware and the JS handlers can simply wrap around Postgraph.

mikeburgh:
My initial needs for this functionality was to extend the GraphQL schema so that some JS functions to fetch remote data (and not touch the PG database) could be included. 

@chadfurman you are right, you can do this by consuming postgraphql as a lib, and then mixing GraphQL with another API (rest or even another graphQL endpoint).  

As some comments here have pointed out, there are some short falls with this currently:
- you cannot use any of the types the introspection creates. So if you want to inject methods that say augment the incoming data before going to PG.. you have to re define the types.

- a connection to PG is opened, even if the method is not going to use it.

I have not had a chance to get back to this, but I should in a week or two.. @benjie you mentioned to give you a nudge on it if you had not got back to it by June 7..

benjie:
I can't help wondering if there's a better way to extend the GraphQL schema... I'm working up an idea at the moment.

cellis:
@benjie @chadfurman you may want to look at [graphql-compose](https://github.com/nodkz/graphql-compose) for inspiration of how to extend graphql schemas.

benjie:
Thanks @cellis, I saw that on [Awesome GraphQL](https://github.com/chentsulin/awesome-graphql). It makes me a bit uncomfortable though because it uses internal GraphQL.js features (rather than just the public API) - there's no public API as far as I know to remove a field from a GraphQL schema.

bradleyayers:
I'm very enthusiastic about this issue, @benjie did you work up a plan for how this would work? Being able to contribute mutation queries backed by JavaScript logic would be great, but I'm curious if you've given any thought to subscription queries?

benjie:
@bradleyayers Yep, I've worked up a solution; I just want to get Caleb's take on it before I announce it...

As for subscriptions; I've not done any work on this yet... But I believe they can be implemented on top of my schema extending stuff 👍

benjie:
Caleb has given me his blessing, so here it is folks: #506

mikeburgh:
Closing this in favor of the v4 approach!

PR #452, opened 2017-04-27
fix(postgraphql): allow meteor-login-token request header when using enableCors option

rossmartin:
If you start a PostGraphQL instance using vanilla HTTP within a meteor server the `addCORSHeaders` function needs to allow the `meteor-login-token` request header.  Otherwise the request fails using a GraphQL client (I am using apollo-client).

Here is how I started the instance in Meteor.
```javascript
http.createServer(postgraphql(POSTGRES_CON_STRING, schema, {
  enableCors: true,
  graphiql: Meteor.isDevelopment,
  bodySizeLimit: '20MB'
})).listen(5000);
```

This is what I see when the `meteor-login-token` is now allowed - 

image

I am considering using PostGraphQL in node.js [as explained here](https://github.com/postgraphql/postgraphql/blob/master/docs/library.md#custom-execution) without going through HTTP but I think there are features that I would lose because I'm using apollo-client.

### DIFF
--- a/src/postgraphql/http/createPostGraphQLHttpRequestHandler.js
+++ b/src/postgraphql/http/createPostGraphQLHttpRequestHandler.js
@@ -515,6 +515,8 @@ function addCORSHeaders (res) {
     // like in a POST request.
     'Content-Type',
     'Content-Length',
+    // Allow meteor-login-token that is sent from the client of a meteor app
+    'meteor-login-token',
   ].join(', '))
 }
 


benjie:
Hi Ross, thanks for the pull request! Sorry you're having this trouble.

I don't think we should be adding framework-specific overrides to the source code like this; if necessary it should be provided by a command line (or in your case, library) option. I'm not sure why Apollo is sending this header? Or is Meteor intercepting it somewhere?

One option is to override this with an express middleware (literally remove this header before it hits postgraphql); but you'll have to be careful!

Another option is to stop Apollo sending this header in the first place.

Failing both of these, we'll happily accept a pull request to add additional CORS headers:

1. Add an option to PostGraphQLOptions: https://github.com/postgraphql/postgraphql/blob/d4fd6a4009fea75dbcaa00d743c985148050475e/src/postgraphql/postgraphql.ts#L11
2. Pass options to addCORSHeaders: https://github.com/rossmartin/postgraphql/blob/454ecb1d426278a08b7641e9a5d2bb029c4c8210/src/postgraphql/http/createPostGraphQLHttpRequestHandler.js#L127
3. Using the new option, add additional headers (e.g. with an object splat).

Good luck!

rossmartin:
@benjie Thanks for all the info and options.  I'll see if I can stop Apollo/Meteor from sending this header in the first place.  The [`applyMiddleware` for apollo-client](http://dev.apollodata.com/core/network.html#networkInterfaceMiddleware) might be what I can use to solve this.

benjie:
@rossmartin Good luck! If you come up with a solution, please share for anyone that comes across this in future :)

rossmartin:
### Update with an easy solution -

If you are using apollo-client you can prevent the `meteor-login-token` from being sent by setting the `useMeteorAccounts` option to false for `createMeteorNetworkInterface` -

```javascript
const networkInterface = createMeteorNetworkInterface({
  ...
  useMeteorAccounts: false
});
```

benjie:
Excellent, thanks for sharing the solution @rossmartin!

PR #453, opened 2017-05-02
Spelling Update: entriess -> entries

peterpme:
I'm assuming this is supposed to be `entries` instead of `entriess` :)

Thanks!

### DIFF
--- a/src/graphql/utils/buildObject.ts
+++ b/src/graphql/utils/buildObject.ts
@@ -30,11 +30,11 @@
  * ```
  */
 export default function buildObject <T>(
-  ...entriess: Array<Array<[string, T] | { key: string, value: T } | undefined | null | false>>,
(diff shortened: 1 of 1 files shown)


benjie:
Nope, entriess is an array of entries's which in turn is an array of entry's. Maybe should have called them entrieses but it looked even weirder to me.

peterpme:
Hey @benjie thanks for the response!

No problem it just looked like a spelling mistake to me. Makes sense :smile:

Perhaps `arrayOfEntries` may be more descriptive but that's entirely up to you, lol

Thanks again!

benjie:
Even that is ambiguous; it'd have to be something like:

```js
  arrayOfArrayOfEntries.forEach(arrayOfEntires => arrayOfEntries.forEach(entry => {
```

(currently:
```js
  entriess.forEach(entries => entries.forEach(entry => {
```
)

Issue #299, opened 2017-01-05
Accessing HTTP request data from PostgreSQL


jinty:
More of a question than anything else.

I'm writing an application which needs an audit log of changes,
I would like to implement that using triggers but I need to record
info from the HTTP request like the IP address of the client.

Is there any built-in way of doing that? Can I write some kind of plugin to pass additional data to postgresql (i.e. in addition to the JWT data)?

kenota:
@jinty You can put ip into JWT claim, use it for validation and Postgraphql expose it in session via SET.

calebmer:
@kenota’s JWT idea sounds good, is there any reason you couldn’t use the [Node.js API](https://github.com/calebmer/postgraphql/blob/master/docs/library.md) and add a middleware?

I’ve also thought about refactoring JWT claims to live on `request.jwt.claims...` instead of `jwt.claims...` so that we could also have `request.headers...`, `request.url...`, etc.

jinty:
@kenota Almost, but not quite, that gives me the IP which requested the JWT token, not the IP which is currently querying postgraphql.

@calebmer: Putting all that data into current settings sounds like a reasonable solution. I wouldn't mix the raw data (i.e. HTTP headers) with processed data (i.e. jwt.claims). I also wonder about encoding though when you get some real nasty junk coming in with the HTTP request: null bytes, non-UTF-8 encodings. Some people will want the raw data, others will want some level of pre-processing.

I was thinking more of a hook point in Javascript where I could ask graphql to run a function like this:

function before_request_hook(postgresql_connection, request) {
     ...
}

I've used pyramid's "tweens" quite a bit and find them very flexible/useful: http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/hooks.html#registering-tweens

calebmer:
That hook is actually a very good idea. It could open a door to all kinds of interesting things. If you wanted to make a PR for this, I’d be down for merging it :blush:

Here are some files you might want to look at:

- https://github.com/calebmer/postgraphql/blob/2b245daace3e25b9eb6fd73e6c1a09ba0e408969/src/postgraphql/http/setupRequestPgClientTransaction.js
- https://github.com/calebmer/postgraphql/blob/2b245daace3e25b9eb6fd73e6c1a09ba0e408969/src/postgraphql/http/createPostGraphQLHttpRequestHandler.js#L285-L288
[...]

PR #460, opened 2017-05-10
Exposing headers to context object and inside postgreSQL as req.headers

mikeburgh:
A solution for #299 which adds the req.headers using set_config.

Only thought I had was do we want to hide these behind an option, so that by default headers are not pushed into postgres ?

### DIFF
--- a/src/postgraphql/withPostGraphQLContext.ts
+++ b/src/postgraphql/withPostGraphQLContext.ts
@@ -38,6 +38,7 @@ export default async function withPostGraphQLContext(
     jwtAudiences = ['postgraphql'],
     pgDefaultRole,
     pgSettings,
+    reqHeaders,
   }: {
     pgPool: Pool,
     jwtToken?: string,
@@ -45,6 +46,7 @@ export default async function withPostGraphQLContext(
     jwtAudiences?: Array<string>,
     pgDefaultRole?: string,
     pgSettings?: { [key: string]: mixed },
+    reqHeaders: { [key: string]: mixed },
   },
   callback: (context: mixed) => Promise<ExecutionResult>,
 ): Promise<ExecutionResult> {
@@ -66,6 +68,7 @@ export default async function withPostGraphQLContext(
       jwtAudiences,
       pgDefaultRole,
       pgSettings,
+      reqHeaders,
     })
 
     return await callback({
@@ -97,6 +100,7 @@ async function setupPgClientTransaction ({
   jwtAudiences,
   pgDefaultRole,
   pgSettings,
+  reqHeaders,
 }: {
   pgClient: Client,
   jwtToken?: string,
@@ -104,6 +108,7 @@ async function setupPgClientTransaction ({
   jwtAudiences?: Array<string>,
   pgDefaultRole?: string,
   pgSettings?: { [key: string]: mixed },
+  reqHeaders: { [key: string]: mixed },
 }): Promise<string | undefined> {
   // Setup our default role. Once we decode our token, the role may change.
   let role = pgDefaultRole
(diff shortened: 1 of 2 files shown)


benjie:
Thanks for the PR 🙏

There needs to be a whitelist, really - or at least some sanity checks. Currently someone could send through an unlimited number of arbitrarily large header keys with arbitrary large values. As you say, there should also be an option to turn this on/off. I'm not sure what performance/memory concerns set_config has.

At least req.headers already does some normalisation for us 🙂

benjie:
I also don't like the idea of Options getting too many more keys; can we add a property that would include headers and other similar properties, e.g. "context". Or maybe just pass through the whole req object?

mikeburgh:
With the option, we could do a single context option, and give it a json struct for what you want on the context (there maybe a better format)

--context {req:{headers:true}} // would apply all the headers to the context and set_config

--context {req:{headers:'host,user-agent'}} // would apply the host and user-agent only headers to the context and set_config

--context {req:{body:true,headers:true}} // would apply all headers and body to the context and set_config

--context {req:true, custom:{abc:123,xyz:456}} // would apply the whole req object, and the custom options to the context and set_config

In this way, you could expand it if there were other things that came up that people wanted set, and you also get custom settings in the command line (I know you can apply custom ones via pgSettings, but I don't think you can get to that from the cli).

If we go this path, then we don't really need to do the whitelist approach as people will be in charge of that via the option (assuming they don't just do req:true).. and perhaps we limit each set_config item to 1000chars just to be safe ?

benjie:
Hi @mikeburgh, I've thought about this some more and I think we want to approach this from a different, less specific, angle.

We have a library option called 'pgSettings' already for passing through static settings. I'm thinking we should change this into a more flexible API by coercing it into an array (i.e. wrapping it with an array if it isn't already an array), where each element is either:

- an object: gets merged into the list of settings to set
- string that starts with a `{`: JSON.parse it (so we can accept it on the command line) and treat it as an object above
- function: gets called with a few interesting thing (e.g. `context`, which should have the `req` object on it), returns an object to be merged into the settings to set
- simple string: is a reference to a list of built in functions:
  - `req.headers` - `({context: {req}}) => mapKeys(req.headers, (value, key) => 'req.headers.' + key)` (but with some protections for number of keys/length of values)
  - `req.user` - `({context: {req}}) => mapKeys(req.user, (value, key) => 'req.user.' + key)` (with similar protections)
  - etc

then we can also expose this on the command line. e.g.

````
postgraphql --pgSettings req.headers --pgSettings req.user --pgSettings '{"a.b": "foo", "c.d": "bar"}'
```

Do you have any thoughts on this?

chadfurman:
@benjie I like your idea.  I think this goes nicely with https://github.com/postgraphql/postgraphql/issues/501   I think

benjie:
When using PostGraphile as a library this functionality is already available via `pgSettings` (though we only support the function and object forms). I'm going to leave this issue open as a reminder to add a CLI equivalent that does the above.

https://www.graphile.org/postgraphile/usage-library/#pgsettings-function

Anything that implements `req.headers` or similar will require significant scrutiny to avoid making the system vulnerable to attacks that set many or large headers.

src/postgraphql/withPostGraphQLContext.ts
```diff
@@ -167,6 +172,11 @@ async function setupPgClientTransaction ({
     localSettings.set(`jwt.claims.${key}`, jwtClaims[key])
   }
 
+  // If we have headers!
+  for (const key of Object.keys(reqHeaders)) {
+    localSettings.set(`req.headers.${key}`, reqHeaders[key])
```
chadfurman: Is this safe?  do we use localSettings in a way where we do not expect escaping?  Are these all type-cast to text?

benjie:
Replacing this with an issue: https://github.com/graphile/postgraphile/issues/769

PR #462, opened 2017-05-10
doc (examples) Clarify computed column function criteria

kenips:
It seems like the computed column functions must be prefixed by the table's name, not table first argument's name. I tried naming the first argument do something else and it still works. Is it a requirement indeed?


benjie:
Good spot, thanks 🙏

benjie:
Can't merge on mobile as CI failing; sorry 😞

benjie:
Thanks @kenips

Issue #417, opened 2017-03-29
Procedure with enum arrays


danielbuechele:
I created a procedure accepting an array of enums as argument, like this:
```
CREATE FUNCTION search_articles(type_filter article_type[]) ...
```
which creates a GraphQL field that can be queried like this:
```
searchArticles(typeFilter:[VIDEO,PHOTO])
```

However, this fails in SQL, saying function `searchArticles(text[])` does not exists. The problem seems to be, that the array of enums is translated to a text array. Probably because in JavaScript the argument creates an array `['VIDEO','PHOTO']`, because JavaScript is not aware of existing enums, which is then transformed to SQL. 

On a first look, it looks like we can add some special case here: https://github.com/postgraphql/postgraphql/blob/master/src/postgraphql/schema/procedures/createPgProcedureSqlCall.ts#L36

calebmer:
Another, simpler option would be to add type casting here: https://github.com/postgraphql/postgraphql/blob/master/src/postgres/inventory/type/PgEnumType.ts#L33 or here: https://github.com/postgraphql/postgraphql/blob/master/src/postgres/inventory/type/PgListType.ts#L40

So that it generates either:

```sql
array['x'::article_type, 'y'::article_type]
```

…or:

```sql
array['x', 'y']::article_type[]
```

Please feel free to submit a PR adding this casting! The former is easier and so probably what we want.

benjie:
![](https://github.trello.services/images/mini-trello-icon.png) [Procedure with enum[] arg](https://trello.com/c/HDHRMEBl/17-procedure-with-enum-arg)

benjie:
If someone fancied simply testing if this works in v4 that would be super helpful 👍

benjie:
(I think it will.)

benjie:
Yep; it works:

```
$ createdb test
$ psql test
[test] # create type test_enum as enum('ONE', 'TWO', 'THREE');
CREATE TYPE
Time: 63.477 ms
[test] # create function second(options test_enum[]) returns test_enum as $$ select options[2]; $$ language sql stable;
CREATE FUNCTION
Time: 8.991 ms
[test] # select second(ARRAY['TWO', 'THREE', 'ONE']::test_enum[]);
 second
--------
 THREE
(1 row)

Time: 0.429 ms
$ postgraphile -c postgres://localhost/test
```

```graphql
{
  second(options: [TWO, THREE, ONE])
}
```

```json
{
  "data": {
[...]

PR #463, opened 2017-05-11
fix(postgres): cast list of enums to their correct type (#417)

danielbuechele:


danielbuechele:
Now using `sql.identifier`. Also changed it to work not only with enums but custom classes, too.

benjie:
This looks great @danielbuechele! Are you able to add any tests for it?

danielbuechele:
Added test for the function I modified

danscan:
@benjie @danielbuechele What is the status of this?  Looks like it's just blocked by some tests.  I can add these if that'd be helpful.

benjie:
@danscan I think we just need CI to pass before we can merge.

benjie:
It seems to just be lint issues that need resolving.

danielbuechele:
I'll fix them tomorrow! (and get my linter working)

danielbuechele:
Huh, linter is now passing, but a lot of other test failing. I don't really know where this is coming from.

benjie:
The errors centre around:

`GraphQLError: function c.types_mutation(unknown, unknown, unknown, text[], unknown, numrange) does not exist`
`GraphQLError: function c.types_query(unknown, unknown, unknown, text[], unknown, numrange) does not exist`

Here's the relevant definitions:

https://github.com/postgraphql/postgraphql/blob/886f8752f03d3fa05bdbdd97eeabb153a4d0343e/examples/kitchen-sink/schema.sql#L155-L156

Seems the integer array is being interpreted as a text array instead.

PR #466, opened 2017-05-16
fix(postgraphql): Throw an error if jwtSecret is provided without jwtPgTypeIndentifer being provided

mikeburgh:
Throws an error if a secret and no token identifier is provided in the options.


src/postgraphql/postgraphql.ts
```diff
@@ -64,6 +64,12 @@ export default function postgraphql (
     options = schemaOrOptions
   }
 
+  // Check for a jwtSecret without a jwtPgTypeIdentifier
+  // a secret without a token identifier prevents JWT creation
+  if (options.jwtSecret && !options.jwtPgTypeIdentifier) {
+    throw new Error(`jwtSecret provided, however jwtPgTypeIdentifier (token identifier) not provided.\n`);
```
benjie: Why the `\n` on the error message?

benjie:
Apparently I can't merge this on mobile; sorry for delay.

benjie:
🙏

benjie:
So... It turns out that *I* am actually using jwtSecret without jwtPgTypeIdentifier - deliberately! I had totally forgotten that.

Reason is: I generate the JWT token *outside* of PostGraphQL, PostGraphQL just uses the jwtSecret to validate the incoming token.

I'm going to change this to a warning.

PR #471, opened 2017-05-23
Fix CI

benjie:


sandersn:
@benjie if you know what changed recently in the dependencies of postgraphql, can you open a bug on Microsoft/TypeScript? The fix to tsconfig indicates that one of the JS libraries that postgraphql depends on caused the Typescript compiler to crash.

benjie:
@sandersn I think it's https://github.com/Microsoft/TypeScript/issues/14628 as noted in ebf68c8a8bf1bec25cdcde73a9df87b93afa8a44 - at least that's where I got the solution from (i.e. using `include`). I'm not sure what changed - we didn't change the dependencies so it must have been a semver-incompatible change in one of our dependencies.

sandersn:
Yep, Microsoft/TypeScript#14628 was how I got here. Big Javascript packages tend to expose new scale bugs in the compiler, so I thought I'd check whether you had a suspect for which dependency changed so I could investigate that package. No worries if not.

sandersn:
(Being explicit about which files to include in the compilation is the right thing, for what my review of this change is worth.)

benjie:
Sorry, I've no theory on this (and the window for it having failed is pretty wide, so I suspect a lot of dependencies have changed in this time). Good luck!

PR #472, opened 2017-05-25
We should be sending a 401 status code not a 403 on expired token.

tim-field:
I noticed this issue when using the **authMiddleware** from https://github.com/nodkz/react-relay-network-layer 

authMiddleware provides the following functionality

> tokenRefreshPromise: - function(req, err) which must return promise or regular value with a new token. This function is called when server returns **401** status code. After receiving a new token, middleware re-run query to the server with it seamlessly for Relay.

However postgraphql only sends a 403 status code.  I was going to create a pull request for authMiddleware but after a bit of research I can see that postgrapqhl should really be sending a 401 in this case*.

Many people will be using authMiddleware with Relay, so this could make things easier all

*https://stackoverflow.com/questions/3297048/403-forbidden-vs-401-unauthorized-http-responses

*https://www.loggly.com/blog/http-status-code-diagram/

### DIFF
--- a/src/postgraphql/withPostGraphQLContext.ts
+++ b/src/postgraphql/withPostGraphQLContext.ts
@@ -115,7 +115,7 @@ async function setupPgClientTransaction ({
     // Try to run `jwt.verify`. If it fails, capture the error and re-throw it
     // as a 403 error because the token is not trustworthy.
     try {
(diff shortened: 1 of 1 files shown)


benjie:
Thanks 🙏 If you fancied following this up with a test, that would be great!

PR #479, opened 2017-05-31
feat(postgraphql) Added an option to allow the pgRole to be extracted from anywhere in the JWT

angelosarto:
When working with Auth0 as the source of the JWT it can be difficult to write directly to the root of the jwt (their rules engine allows writing to the jwt but it has to be namespaced with a url-like property name.)

I created this PR that allows you to specify a path in the jwt from which to set the pgRole; if unset it defaults to the current 'role' property on the root of the jwt.

Since this is my first commit I probably missed something, though I did add unit tests where appropriate, updated docs, ran all tests, and tried to follow the commit guidlines.


angelosarto:
oops - didn't run the linter... knew I missed something

benjie:
@angelosarto In future it's okay to just push your updated code to the branch - there's no need to close the PR and open another one 👍

benjie:
Replaced by #480

PR #480, opened 2017-05-31
feat(postgraphql): extract pgRole from arbitrary JWT path

angelosarto:
… from any location in the JWT, defaults to 'role' on the root of the JWT object.

When working with Auth0 as the source of the JWT it can be difficult to write directly to the root of the jwt (their rules engine allows writing to the jwt but it has to be namespaced with a url-like property name.)

I created this PR that allows you to specify a path in the jwt from which to set the pgRole; if unset it defaults to the current 'role' property on the root of the jwt.

Since this is my first commit I probably missed something, though I did add unit tests where appropriate, updated docs, ran all tests, (ran the linter) and tried to follow the commit guidelines.

### DIFF
--- a/docs/library.md
+++ b/docs/library.md
@@ -59,7 +59,7 @@ Arguments include:
   - `pgDefaultRole`: The default Postgres role to use. If no role was provided in a provided JWT token, this role will be used.
   - `jwtSecret`: The secret for your JSON web tokens. This will be used to verify tokens in the `Authorization` header, and signing JWT tokens you return in procedures.
   - `jwtAudiences`: The audiences to use when verifing the JWT token. If not set the audience will be `['postgraphql']`.
-  - `jwtRole`: a comma seperated list of strings that create a path in the jwt from which to extract the postgres role. if none is provided it will use the key `role` on the root of the jwt.
+  - `jwtRole`: a comma seperated list of strings that create a path in the jwt from which to extract the postgres role. If none is provided it will use the key `role` on the root of the jwt.
   - `jwtPgTypeIdentifier`: The Postgres type identifier for the compound type which will be signed as a JWT token if ever found as the return type of a procedure. Can be of the form: `my_schema.my_type`. You may use quotes as needed: `"my-special-schema".my_type`.
   - `watchPg`: When true, PostGraphQL will watch your database schemas and re-create the GraphQL API whenever your schema changes, notifying you as it does. This feature requires an event trigger to be added to the database by a superuser. When enabled PostGraphQL will try to add this trigger, if you did not connect as a superuser you will get a warning and the trigger won’t be added.
   - `disableQueryLog`: Turns off GraphQL query logging. By default PostGraphQL will log every GraphQL query it processes along with some other information. Set this to `true` to disable that feature.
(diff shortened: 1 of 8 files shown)


src/postgraphql/withPostGraphQLContext.ts
```diff
@@ -124,7 +129,16 @@ async function setupPgClientTransaction ({
...
+      let roleClaim = jwtClaims ? jwtClaims : ''
+      const rolePath = jwtRole ? jwtRole : []
+      for (let i = 0; i < rolePath.length; i++) {
+        try {
+          roleClaim = roleClaim[rolePath[i]]
+        } catch (e) {
+          i = rolePath.length
+          roleClaim = ''
+        }
+      }
```
benjie: I'd rather see the logic here extracted to utility function; e.g. the below is a slightly modified version of lodash's `get`:

```js
function getPath(inObject, path) {
  let object = inObject
  // From https://github.com/lodash/lodash/blob/master/.internal/baseGet.js
  let index = 0
  const length = path.length

  while (object && index < length) {
    object = object[path[index++]]
  }
  return (index && index === length) ? object : undefined
}

// ...

const roleClaim = getPath(jwtClaims, jwtRole);
```

angelosarto:
Don't merge yet -- accidentally hit update branch.  Changes not made yet.

benjie:
@angelosarto Cool; let me know when you're ready for me to re-review.

angelosarto:
Ok - I extracted the function as suggested - it actually works great as is so I added it as a private method in the same file.  all the tests still passed so I think we are good to go.

src/postgraphql/withPostGraphQLContext.ts
```diff
@@ -227,4 +232,22 @@ function debugPgClient (pgClient: Client): Client {
 
   return pgClient
 }
+
+/**
+ * Safely extract a nested object or undefined where inObject is any object and path is
+ * an array of indexes into an object
```
benjie:
```
Safely gets the value at `path` (array of keys) of `inObject`.
```

src/postgraphql/withPostGraphQLContext.ts
```diff
@@ -227,4 +232,22 @@ function debugPgClient (pgClient: Client): Client {
...
+ *
+ * @private
+ */
+function getPath(inObject: any, path: any): any {
```
benjie: Maybe something like `(inObject: mixed, path: Array<string>)`?

docs/library.md
```diff
@@ -59,6 +59,7 @@ Arguments include:
   - `pgDefaultRole`: The default Postgres role to use. If no role was provided in a provided JWT token, this role will be used.
   - `jwtSecret`: The secret for your JSON web tokens. This will be used to verify tokens in the `Authorization` header, and signing JWT tokens you return in procedures.
   - `jwtAudiences`: The audiences to use when verifing the JWT token. If not set the audience will be `['postgraphql']`.
+  - `jwtRole`: a comma seperated list of strings that create a path in the jwt from which to extract the postgres role. if none is provided it will use the key `role` on the root of the jwt.
```
benjie: `if none is provided` < missing capital

angelosarto:
Added the fixes to docs, and the typescript types on the function.   Also removed optional on the signature as a default is provided instead of undefined.

benjie:
And it's in 👍

doeg:
Hey @angelosarto,

Thank you for implementing this feature. I'm implementing Auth0 alongside postgraphql and your PR saved me a _ton_ of time and frustration (and hacky code). 😄 I'm running a build-from-source version to pick up your recent change and it's working like a charm.

And thanks @benjie for merging it. I really appreciate this project.

PR #504, opened 2017-07-08
add Quering

ppKrauss:
adding a little bit more, as example for "how to grow"


benjie:
Rather than merging this I want to make a point to document the type mappings.

benjie:
Moved to https://github.com/graphile/graphile.github.io/issues/51

PR #511, opened 2017-07-11
allow all numbers in jwt token type names

gkchestertron:
currently naming of the jwt_token type, as in the tutorial/example only allows for names that contain the numbers 1 and 2. I'm using this for multiple PTA organizations and would like to keep consistent naming and identify them by their district number (i.e. pta_dist_14.jwt_token). 

This fixes the issue - but I am unsure if there was a good reason that only 1 and two were allowed before.


benjie:
O_o this seems weird; surely that'd have to be deliberate... @calebmer any idea?

gkchestertron:
right? but I couldn't think of one.

benjie:
Tracking through various levels of git blame; this is where it was originally introduced:

https://github.com/postgraphql/postgraphql/commit/f8b2d42c4d17dfd099a1762f28f33d675b50f219

> "Type identifiers are a bit tricky, but this function uses a regular expression to do it. Fun."

Worrisome... I wonder what's tricky about them.

I had a scan through the docs and can't see any issues:

https://www.postgresql.org/docs/9.6/static/sql-createtype.html

Could you add some tests so we can be certain?

gkchestertron:
will do.

gkchestertron:
Let me know if I've complete missed on your conventions - I'm new to ES6, TypeScript, Jest, Travis, and ESLint. Also - sorry this is another suite of slow tests in the postgraphql space.

src/postgraphql/schema/auth/__tests__/getPgTokenTypeFromIdentifier-test.js
```diff
@@ -0,0 +1,39 @@
+import withPgClient from '../../../../__tests__/utils/withPgClient'
+import introspectDatabase from '../../../../postgres/introspection/introspectDatabase'
+import getPgTokenTypeFromIdentifier from '../getPgTokenTypeFromIdentifier'
+
+// good names start with a letter or underscore and may also contain numbers after position 0
+const goodNames = [
+  'test', 'test_0', 'test_1', 'test_2', 'test_3', 'test_4', 'test_5', 'test_6', 'test_7', 'test_8', 'test_9', '_test',
```
benjie: This only really needs to test edge cases, so 'test_0', 'test_9' and '_test' should suffice

src/postgraphql/schema/auth/__tests__/getPgTokenTypeFromIdentifier-test.js
```diff
@@ -0,0 +1,39 @@
...
+    let catalog   = await introspectDatabase(client, [goodName])
+    let tokenType = getPgTokenTypeFromIdentifier(catalog, `${goodName}.jwt_token`)
+
+    expect(tokenType.name).toBe('jwt_token')
+  }))
+})
+
+// schema, type, etc... names are not allowed to start with numbers
+const badNames = [
+  '0_test', '1_test', '2_test', '3_test', '4_test', '5_test', '6_test', '7_test', '8_test', '9_test',
```
benjie: similarly I'd only test '0_test' and '9_test'

src/postgraphql/schema/auth/__tests__/getPgTokenTypeFromIdentifier-test.js
```diff
@@ -0,0 +1,43 @@
...
+]
+
+for (var i in badNames) {
+  let badName = badNames[i]
+
+  test(`Bad schema name ${badName} should throw a syntax error on create`, withPgClient(async client => {
+    let badQuery = 'not an error'
+
+    try {
+      badQuery = await client.query(`create schema ${badName}; create type ${badName}.jwt_token as (role text, exp integer, a integer, b integer, c integer)`)
```
benjie:
```
create schema "0_test";
CREATE SCHEMA
Time: 35.303 ms
```

src/postgraphql/schema/auth/__tests__/getPgTokenTypeFromIdentifier-test.js
gkchestertron: fair enough

src/postgraphql/schema/auth/__tests__/getPgTokenTypeFromIdentifier-test.js
gkchestertron: you're right. works fine when escaped.

benjie:
@gkchestertron Are you good for this to be merged?

gkchestertron:
works for me

PR #512, opened 2017-07-12
chore(test): alphabetically order schema in snapshots

benjie:
In preparation for v4 this will make it easier to determine how close the generated schemas are.


benjie:
See https://github.com/graphql/graphql-js/issues/941 regarding alphabetic ordering.

Issue #485, opened 2017-06-08
Return more exception details than just the message?


zopf:
I am raising exceptions as part of my custom functions in PostgreSQL, and I'd like more details than just the message to be reported by PostgraphQL (for example, the `HINT`, `ERRCODE`, or `SQLSTATE` as specified in the [Postgres docs](https://www.postgresql.org/docs/9.5/static/plpgsql-errors-and-messages.html)).

Here's an example of an exception I'd throw:
```sql
-- check phone number format
if phone_number !~ '^\+[0-9]{8,20}$' then
  RAISE EXCEPTION 'Your phone number must start with a +, contain only numbers, and be at least 8 digits long.'
    USING ERRCODE = '12345';
  return null;
end if;
```

And here is the response that I currently see:
```json
{
  "data": {
    "authenticate": null
  },
  "errors": [
    {
      "message": "Your phone number must start with a +, contain only numbers, and be at least 8 digits long.",
      "locations": [
        {
          "line": 9,
          "column": 3
        }
      ],
      "path": [
        "authenticate"
      ]
    }
  ]
}
```

I'd love there to be a way for PostgraphQL to report the `HINT` or `ERRCODE` that my custom function might send as part of raising an exception.  Is there a good way to have PostgraphQL do so?

One suggestion could be to have optional `"hint"` and `"errcode"` fields in the same context as the `"message"` field.

benjie:
@zopf Looks like it should be possible to add that relatively easily by just modifying the `formatError` method here:

https://github.com/postgraphql/postgraphql/blob/b2155c9a5048b90f9f8c0357eeb2f3084241c60d/src/postgraphql/http/createPostGraphQLHttpRequestHandler.js#L78-L90

If that was wrapped in a CLI option that defaults false, perhaps something like `--extended-errors`, then I'd be happy to see a PR implementing this functionality!

zopf:
Making progress on this today.  Expect a PR soon.

Note: it is theoretically possible to expose any of the fields listed [here in the pg spec](https://github.com/brianc/node-postgres/blob/master/lib/connection.js#L579).

I'll be exposing `code`, `detail`, and `hint`.

benjie:
@zopf Are there security concerns with doing so - revealing too much of your database internals may be risky? I'm wondering if code/detail may be safe but hint maybe only wants to be on development (or opt-in)?

How about we change the `--extended-errors` option from a boolean to a list of fields to be exposed, so people can opt-into it?

[...]

PR #513, opened 2017-07-13
feat(postgraphql) add extendedErrors option

zopf:
In response to issue #485.  Adds the `extendedErrors` option to the `postgraphql` middleware and CLI.  Includes documentation update, but no tests since I could not find a good way to fit an error-producing test into the kitchen-sink integration test suite.

Example:

```
// ---- PostGraphQL setup (for Koa)
[...]
config.options = {
  extendedErrors: ['hint', 'detail', 'errcode']
};
app.use(postgraphql(config.postgresCnx, config.schemaName, config.options));


// ---- SQL function:
CREATE OR REPLACE FUNCTION throw_test_error(throw boolean)
RETURNS boolean LANGUAGE 'plpgsql' STABLE AS $f$
begin
  if throw then
    RAISE EXCEPTION 'test message'
      USING ERRCODE = '12345', HINT = 'test hint', DETAIL = 'test detail';
  else return true;
  end if;
end; $f$;
GRANT EXECUTE ON FUNCTION throw_test_error(boolean) TO PUBLIC;


// ---- test query:
query test {
  throwTestError(throw: true)
}


// ---- test response, with extendedErrors undefined
{
  "errors": [
    {
      "message": "test message",
      "locations": [
        {
          "line": 28,
          "column": 3
        }
      ],
      "path": [
        "throwTestError"
      ]
    }
  ],
  "data": {
    "throwTestError": null
  }
}


// ---- test response, with extendedErrors set to ['hint', 'detail', 'errcode']
{
  "errors": [
    {
      "message": "test message",
      "locations": [
        {
          "line": 28,
          "column": 3
        }
      ],
      "path": [
        "throwTestError"
      ],
      "hint": "test hint",
      "detail": "test detail",
      "errcode": "12345"
    }
  ],
  "data": {
    "throwTestError": null
  }
}

### CODE REVIEW

**Comment by zopf** [comment]
Note: this allows the user to expose the `hint`, `detail` and `code` (as `errcode`) fields on a Postgres error as desired.  Additional fields that could be added in future PRs can be seen here: https://github.com/brianc/node-postgres/blob/master/lib/connection.js#L579

File: src/graphql/utils/extendedFormatError.ts [hunk 1]
```diff
@@ -0,0 +1,43 @@
...
+export type GraphQLErrorLocation = {
+  line: number,
+  column: number,
+}
+
+export type GraphQLErrorExtended = GraphQLError & {
+  hint: string,
+  detail: string,
+  code: string,
+}
```
benjie: Please move this into `src/postgraphql` as it's done as an operation after graphql resolves rather than being part of the graphql resolving itself.

src/graphql/utils/extendedFormatError.ts
```diff
@@ -0,0 +1,43 @@
...
+ * Response Format, Errors section of the GraphQL Specification, plus a few
+ * additional fields relevant to postgres errors, including HINT, DETAIL,
+ * and ERRCODE.
+ */
+export function extendedFormatError(error: GraphQLError, fields: Array<String>): GraphQLFormattedErrorExtended {
+  if (!error) {
+    throw new Error('Received null or undefined error.')
+  }
+  const originalError = error.originalError as GraphQLErrorExtended
+  fields = fields.map(field => field.toLowerCase())
```
benjie: This should be done when the option is passed to postgraphql so it's only executed once in total rather than once for every error

src/postgraphql/postgraphql.ts
```diff
@@ -21,6 +21,7 @@ type PostGraphQLOptions = {
   jwtPgTypeIdentifier?: string,
   watchPg?: boolean,
   showErrorStack?: boolean,
+  extendedErrors?: Array<string>,
```
benjie: Process it to lower case in this file 👍 

(Regarding my comment on
```
fields = fields.map(field => field.toLowerCase())
```
)

zopf: done :)

src/graphql/utils/extendedFormatError.ts
zopf: ah, yes, makes sense.  done

src/graphql/utils/extendedFormatError.ts
zopf: yes that would be more efficient.  done

zopf:
I had indeed tried that same integration test approach before, but encountered errors I did not fully understand, so I backed it out.

For example:

```
● unique-constraints.graphql

    - Snapshot
    + Received

    @@ -1,18 +1,16 @@
     Object {
       "data": Object {
    -    "a": Object {
    -      "email": "sara.smith@email.com",
    -      "name": "Sara Smith",
    -    },
    +    "a": null,
         "b": null,
    -    "c": Object {
    -      "personId1": 1,
    -      "personId2": 2,
    -    },
    -    "d": Object {
    -      "personId1": 4,
    -      "personId2": 4,
    -    },
    +    "c": null,
    +    "d": null,
         "e": null,
       },
    +  "errors": Array [
    +    [GraphQLError: current transaction is aborted, commands ignored until end of transaction block],
    +    [GraphQLError: current transaction is aborted, commands ignored until end of transaction block],
    +    [GraphQLError: current transaction is aborted, commands ignored until end of transaction block],
    +    [GraphQLError: current transaction is aborted, commands ignored until end of transaction block],
    +    [GraphQLError: current transaction is aborted, commands ignored until end of transaction block],
    +  ],
     }
```

I was puzzled because `unique-constraints.graphql`'s function shouldn't be affected at all by my `extended-errors.graphql` file's execution, wouldn't you think?

But here's the problem: we're executing all the graphql queries in parallel, using the same pg client!  Thus, if any of the queries causes an error, it causes the others to fail during that same transaction on that same connection.

I am going to try resolving that by executing the integration test graphql queries in sequence instead.

benjie:
Good idea; `withPgClient` wraps them in a transaction, but you can use savepoints between them, something like:

```js
    return await withPgClient(async pgClient => {
      // Add data to the client instance we are using.
      await pgClient.query(await kitchenSinkData)
      for (const fileName of queryFileNames) { // <--
        await pgClient.query("savepoint pretest"); // <--
        // Read the query from the file system.
        const query = await new Promise((resolve, reject) => {
          readFile(resolvePath(queriesDir, fileName), 'utf8', (error, data) => {
            if (error) reject(error)
            else resolve(data)
          })
        })
        await pgClient.query("rollback to savepoint pretest"); // <--
        ...
```

zopf:
It appears that the savepoint doesn't prevent a raised exception from rendering the transaction unusable.

Here's an isolated SQL test case:
```
begin;
savepoint foo;
select a.throw_test_error(true);
rollback to savepoint foo;
select a.throw_test_error(false);
commit;
```
yields:
```
ERROR:  current transaction is aborted, commands ignored until end of transaction block
********** Error **********

ERROR: current transaction is aborted, commands ignored until end of transaction block
SQL state: 25P02
```

So, it appears the only way to ensure that a query that could raise an exception does not affect other tests is to run it in a separate transaction, which means running it in a separate `withPgClient` block.  The mildly unfortunate implication is that this means we need to set up the kitchen sink data on each client.. but luckily there is not much data, so it only adds a couple seconds to the full test suite's execution time.  Will submit a revised commit in a bit for review.

benjie:
@zopf you didn't rollback to savepoint after the second one before the commit?

```bash
cat <<HERE | psql postgres
begin;
savepoint foo;
do \$\$ begin raise exception 'Exception thrown from PL/PGSQL'; end; \$\$ language plpgsql;
rollback to savepoint foo;
do \$\$ begin raise exception 'Exception thrown from PL/PGSQL'; end; \$\$ language plpgsql;
rollback to savepoint foo;
select 1;
commit;
HERE
```

```
SET
Time: 0.205 ms
BEGIN
Time: 0.059 ms
SAVEPOINT
Time: 0.058 ms
ERROR:  P0001: Exception thrown from PL/PGSQL
CONTEXT:  PL/pgSQL function inline_code_block line 1 at RAISE
LOCATION:  exec_stmt_raise, pl_exec.c:3165
Time: 1.487 ms
ROLLBACK
Time: 0.067 ms
ERROR:  P0001: Exception thrown from PL/PGSQL
CONTEXT:  PL/pgSQL function inline_code_block line 1 at RAISE
LOCATION:  exec_stmt_raise, pl_exec.c:3165
Time: 0.103 ms
ROLLBACK
Time: 0.038 ms
 ?column?
----------
        1
(1 row)

Time: 0.206 ms
COMMIT
Time: 0.048 ms
```

zopf:
Sorry, should have posted my `throw_test_error` function!  The `false` argument means that it just returns true.  I actually don't need to have it there.  Posting full example here for clarity:

```
DROP FUNCTION IF EXISTS a.throw_test_error(boolean);
CREATE FUNCTION a.throw_test_error(throw boolean)
RETURNS boolean LANGUAGE 'plpgsql' STABLE AS $f$
begin
  if throw then
    RAISE EXCEPTION 'test message'
      USING ERRCODE = '12345', HINT = 'test hint', DETAIL = 'test detail';
  else return true;
  end if;
end; $f$;

begin;
savepoint foo;
select a.throw_test_error(true);
rollback to savepoint foo;
select 1;
commit;
```

It turns out that I was executing this on a client that maintained the same connection.  After the commit, when I ran the script again, it would give me the `ERROR:  current transaction is aborted, commands ignored until end of transaction block`.

If I just run it once on a fresh connection, or if I run it via the command line, it appears to work properly.

After some refactoring, it looks like the `savepoint` and `rollback` are working properly with sequential queries with just a single client!  I'm adding a test on the HTTP interface as well, since that's where the error reporting and formatting actually happens...

zopf:
Soooo after figuring out how to get all these tests to run in sequence effectively instead of in parallel, I realized that since this is just the schema passing into `graphql`, the error is never actually getting passed through `formatError`, since that happens only in the HTTP request handler.  Thus, we can't actually test the effect of the `extendedErrors` option in `postgraphqlIntegrationQueries-test.js`!

I wrote a couple simple tests in `createPostGraphQLHttpRequestHandler-test.js` which should be enough to cover the basics here.

benjie:
Oh gosh, I'm sorry I should have realised that we were testing the wrong thing! 😳

benjie:
Just some minor lint issues to fix

zopf:
Ok, lint issues fixed!

src/postgraphql/cli.ts
```diff
@@ -122,6 +123,7 @@ const server = createServer(postgraphql(pgConfig, schemas, {
   pgDefaultRole,
   watchPg,
   showErrorStack,
+  extendedErrors,
```
benjie: You've added the options here, but you've not added them to the list on lines 23-45? Also you may need to coerce to array here by splitting on commas?

zopf: Absolutely right.  Fixed, tested with CLI, and cleaned up docs

benjie:
@zopf If you're happy with my changes I'll get this merged and released in 3.4.0 sometime next week 👍

PR #553, opened 2017-08-15
Introduced pgTransactionSetup option with tests to main postgraphql API

explicitcall:
This adds `pgTransactionSetup` option to the top-level middleware API and then propagated up to `withPostGraphQLContext`, which checks if the option is set at all. If not, it's then set to the postgraphql-provided JWT parsing implementation.

### DIFF
--- a/src/postgraphql/postgraphql.ts
+++ b/src/postgraphql/postgraphql.ts
@@ -5,6 +5,7 @@ import { EventEmitter } from 'events'
 import { createPostGraphQLSchema, watchPostGraphQLSchema } from 'postgraphile-core'
 import createPostGraphQLHttpRequestHandler, { HttpRequestHandler } from './http/createPostGraphQLHttpRequestHandler'
 import exportPostGraphQLSchema from './schema/exportPostGraphQLSchema'
+import { TransactionSetupOpts } from './withPostGraphQLContext'
 
 type PostGraphQLOptions = {
   classicIds?: boolean,
@@ -26,7 +27,7 @@ type PostGraphQLOptions = {
   exportJsonSchemaPath?: string,
   exportGqlSchemaPath?: string,
   bodySizeLimit?: string,
-  pgTransactionSetup?: (opts: any) => Promise<string | undefined>,
+  pgTransactionSetup?: (opts: TransactionSetupOpts) => Promise<string | undefined>,
   pgSettings?: { [key: string]: mixed },
   appendPlugins?: Array<(builder: mixed) => {}>,
   prependPlugins?: Array<(builder: mixed) => {}>,
--- a/src/postgraphql/postgraphql.ts
+++ b/src/postgraphql/postgraphql.ts
@@ -26,6 +26,7 @@ type PostGraphQLOptions = {
   exportJsonSchemaPath?: string,
   exportGqlSchemaPath?: string,
   bodySizeLimit?: string,
+  pgTransactionSetup?: (opts: any) => Promise<string | undefined>,
   pgSettings?: { [key: string]: mixed },
   appendPlugins?: Array<(builder: mixed) => {}>,
   prependPlugins?: Array<(builder: mixed) => {}>,
--- a/src/postgraphql/withPostGraphQLContext.ts
+++ b/src/postgraphql/withPostGraphQLContext.ts
@@ -5,6 +5,16 @@ import { ExecutionResult } from 'graphql'
 import * as sql from 'pg-sql2'
 import { $$pgClient } from '../postgres/inventory/pgClientFromContext'
 
+export interface TransactionSetupOpts {
+  pgClient: Client,
+  jwtToken?: string,
+  jwtSecret?: string,
+  jwtAudiences?: Array<string>,
+  jwtRole: Array<string>,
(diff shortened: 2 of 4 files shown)


explicitcall:
This PR obviously needs more user-visible documentation for the new option, but I just want to check if it would work at all for you @benjie

src/postgraphql/withPostGraphQLContext.ts
```diff
@@ -61,7 +73,11 @@ export default async function withPostGraphQLContext(
 
   // Run the function with a context object that can be passed through
   try {
-    const pgRole = await setupPgClientTransaction({
+    if (!pgTransactionSetup) {
+      pgTransactionSetup = setupPgClientTransaction
+    }
+
+    const pgRole = await pgTransactionSetup({
```
benjie: I do not like overwriting `arguments`; can we change this to something more like:

```js
const setupPgClientTransaction = pgTransactionSetup || defaultSetupPgClientTransaction;
const pgRole = await setupPgClientTransaction({
```

(and rename the existing `setupPgClientTransaction` to `defaultSetupPgClientTransaction`)

src/postgraphql/postgraphql.ts
```diff
@@ -26,6 +27,7 @@ type PostGraphQLOptions = {
   exportJsonSchemaPath?: string,
   exportGqlSchemaPath?: string,
   bodySizeLimit?: string,
+  pgTransactionSetup?: (opts: TransactionSetupOpts) => Promise<string | undefined>,
```
benjie: This needs a much scarier name like `dangerouslySetInnerHTML` has in React; maybe something like: `pgDangerouslyOverrideDefaultTransactionSetup`

We should also throw an error if you specify any of the options that `setupPgClientTransaction` uses if you are using `pgDangerouslyOverrideDefaultTransactionSetup`, e.g. `jwtRole`, `jwtSecret`, `jwtAudiences`, `pgDefaultRole`, `pgSettings` - because you're opting to take this entirely into your own hands you will be responsible for re-implementing these functionalities and supporting the default options will be misleading.

explicitcall:
Many thanks for the feedback!

This approach was based on my need to manually parse a JWT token from the request. Now I'm actually trying to get the role setting from the request session. In both cases standard postgraphql JWT parsing coming with `setupPgClientTransaction` is redundant. Do you think that abandoning this approach and rewriting this with a `Map` modification callback is worth it even for the mentioned scenarios, where library users parse (or completely discard) JWT?

benjie:
If you don't specify a `jwtSecret` then we'll automatically skip getting the JWT, so no redundancy:

https://github.com/postgraphql/postgraphql/blob/233f55c53f8c982cb566b48b63e0fa5bcc7ed1fd/src/postgraphql/http/createPostGraphQLHttpRequestHandler.js#L394

---

Perhaps this is better served by allowing `pgSettings` (which already exists) to be a function. Then

https://github.com/postgraphql/postgraphql/blob/233f55c53f8c982cb566b48b63e0fa5bcc7ed1fd/src/postgraphql/http/createPostGraphQLHttpRequestHandler.js#L403

could become:

```js
await (typeof pgSettings === 'function' ? pgSettings(req) : pgSettings),
```

That would allow you to supplement the auto-generated settings from the options. And if you don't want some of the default settings like `pgDefaultRole` just don't set the options associated with them and handle it entirely in your `pgSettings` function.

---

Unless you're expecting to do something to the transaction that cannot be done via `set_config(key, value, true)` it seems ill-advised to replace the tested logic. There are things that you MIGHT want to do outside `set_config` though; for example you might want to `SET LOCAL SESSION AUTHORIZATION` or `RESET ALL` or similar - so it's definitely worth having the discussion. I'm just concerned that if we give people such a big escape hatch to abuse I'm going to get a lot of issues raised from people not understanding it and doing insecure things. If we can serve your needs by simply making `pgSettings` a function then that's a much less risky approach and we can save this escape hatch for when someone really needs it.

If you decide to go the `pgSettings` approach I'd encourage you to close this PR and open a new one, since the code in here is useful in case we want this in future.

explicitcall:
Well, I've actually managed to achieve this with plain `pgSettings` object this way:

```JavaScript
export async function index(req, res, next) {
  return postgraphql(sails.config.connections.localPostgresqlServer,
    'public', {
      graphiql: false,
      pgDefaultRole: 'customer',
      pgSettings: {
        'user.id': `${req.session.passport.user}`,
      },
    })(req, res, next);
}
```

From what I see, regenerating `postgraphql` middleware on every request shouldn't have a significant impact as that only returns existing functions, would it implicitly regenerate the schema on every request though?

benjie:
Yeah, that's going to be pretty expensive - it'll have to introspect the database for every request!

benjie:
The syntax I'm proposing would be:

```js
export postgraphql(sails.config.connections.localPostgresqlServer,
  'public', {
    graphiql: false,
    pgDefaultRole: 'customer',
    pgSettings: req => ({
      'user.id': `${req.session.passport.user}`,
    }),
  })
```

explicitcall:
Closing this with new implementation submitted as #554

PR #556, opened 2017-08-16
v3 `pgSettings` function

explicitcall:
I hope this gets merged to v4 soon, which I prefer to use rather than v3, due to better SQL queries v4 generates on average 😄


benjie:
It's out in `postgraphql@4.0.0-alpha2.12`

Issue #566, opened 2017-08-26
Bug: pgSettings turn value types into strings


lifeiscontent:
given the following pgSettings:

``` js
pgSettings: req => ({
            'user.user_id': req.ctx.session.user_id || ''
          })
```

I expect the following:

when `req.ctx.session.user_id === null` to set `user.user_id` to
  expected: `""` or `null`
  actual: `"null"`

when `req.ctx.session.user_id === undefined` to set `user.user_id` to
  expected: `""` or `null`
  actual: `"undefined"`

benjie:
Suggestion: strings and numbers should be passed through, null/undefined should be dropped, and everything else should throw an error.

lifeiscontent:
@benjie if I was to make a PR for this, which branch or repo should I work in?

benjie:
Master; I'll then merge into v3 and v4

benjie:
Specifically you want to extract this line and add slightly more logic:

https://github.com/postgraphql/postgraphql/blob/master/src/postgraphql/http/createPostGraphQLHttpRequestHandler.js#L404

PR #589, opened 2017-09-26
feat(pgsettings): Only use settings of type string or number

skvale:
~~For example, instead of an `undefined` pgSetting becoming `"undefined"` turn it into `null` so `set_config` can still accept it~~

Fixes: #566

### DIFF
--- a/src/postgraphql/withPostGraphQLContext.ts
+++ b/src/postgraphql/withPostGraphQLContext.ts
@@ -253,19 +253,20 @@ function getPath(inObject: mixed, path: Array<string>): any {
 }
 
 /**
- * Check if a pgSetting is valid.
- * Null and Undefined settings are not valid.
- * pgSettings objects throw an error.
+ * Check if a pgSetting is a string or a number.
+ * Null and Undefined settings are not valid and will be ignored.
+ * pgSettings of other types throw an error.
  *
  * @private
  */
 function isPgSettingValid(pgSetting: mixed): boolean {
+  const supportedSettingTypes = ['string', 'number']
+  if (supportedSettingTypes.indexOf(typeof pgSetting) >= 0) {
+    return true
+  }
   if (pgSetting === undefined || pgSetting === null) {
     return false
   }
-  if (typeof pgSetting === 'object') {
-    throw new Error(`Invalid pgSetting: ${pgSetting} is an Object.`)
-  }
-  return true
+  throw new Error(`Error converting pgSetting: ${typeof pgSetting} needs to be of type ${supportedSettingTypes.join(' or ')}.`)
 }
 // tslint:enable no-any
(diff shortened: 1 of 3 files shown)


src/postgraphql/withPostGraphQLContext.ts
```diff
@@ -163,7 +163,7 @@ async function setupPgClientTransaction ({
   // this prevents an accidentional overwriting
   if (typeof pgSettings === 'object') {
     for (const key of Object.keys(pgSettings)) {
-      localSettings.set(key, String(pgSettings[key]))
+      localSettings.set(key, !!pgSettings[key] ? String(pgSettings[key]) : null)
```
benjie: This is not a safe way to check if the value should be excluded - they might deliberately be setting it to the empty string `""` or to the number zero `0` and this will instead set `null`. A safer way would be `pgSettings[key] != null ? String(...) : null`; however I'd rather throw if the values are invalid:

> strings and numbers should be passed through, null/undefined should be dropped, and everything else should throw an error.
> -- https://github.com/postgraphql/postgraphql/issues/566#issuecomment-325129925

So I'd like to see some `typeof`'s around here. For anyone wondering: I'm deliberately suggesting we should throw on booleans, and on objects that have a `toString` value.

skvale:
@benjie 
That sounds good.  
 - I wasn't sure how to output deprecation warnings just `console.warn`?
 - Still thinking of a better error/deprecation message "`Error converting pgSetting: ${pgSetting} is an Object.`"
 - Could also throw an error for `Symbol` types in response to 
      > everything else should throw an error

benjie:
> I wasn't sure how to output deprecation warnings just console.warn?

Yup; with `DEPRECATED` prefixed. Feel free to use `chalk` to make it more obvious!

> Still thinking of a better error/deprecation message "Error converting pgSetting: ${pgSetting} is an Object."

Generally you should add what it is (as you have done) and what it *should be* - i.e. append "expected string or number" (no need to include `null` in the list IMO).

> Could also throw an error for Symbol types in response to

Yes, literally everything that doesn't match `(pgSetting == null || ['string', 'number'].includes(typeof pgSetting)` should  throw.

> Not sure why we should throw for booleans, but could add that check too

Because in SQL it's not clear if you'd check for it to exist or be `NULL`, or for it to be `"0"` vs `"1"`; or `"TRUE"` vs `"FALSE"`; or `"t"` vs `"f"`; or `"true"` vs `"false"`

skvale:
I based this PR off of v4 `postgraphql:benjie/graphql-build`

skvale:
and addressed all of the requested changes, I think.

benjie:
I'd still like to see this; but the tests need to pass on CI first 👍

benjie:
Thanks @skvale!

PR #591, opened 2017-09-27
Clarify the expected value of `jwtRole`

bradleyayers:


benjie:
Weird CI issue :(

chadfurman:
@benjie CI doesn't run automatically when a PR is opened, does it?  If so, that's very dangerous.  If not, good for you!  :)

benjie:
Yes it does, but CI doesn't have access to anything dangerous (e.g. it doesn't publish to npm/etc, so doesn't have any secrets in its envvars or similar). I'm not aware of any security issues with doing so, but if we were to auto-publish to npm or similar then I'd certainly lock it down! If you have evidence of it being dangerous configured how it is currently please let me know so I can disable it.

benjie:
So I think the CI issue is that something's updated and UglifyJs can't handle `=>` arrow function syntax. v4 has a yarn.lock preventing this being an issue but v3 doesn't.

chadfurman:
w.r.t. security, a compromised CI server (if it builds _anything_) could backdoor the builds.  That said, if the CI server is read-only and only runs the tests, the wors that could happen is that the CI server is zombified and used to perform malicious acts against others.

PR #595, opened 2017-10-01
feat(postgraphql): add option to enable default mutations only for the specified tables

knqyf263:
Close https://github.com/postgraphql/postgraphql/issues/576
Relate: https://github.com/graphile/graphile-build/pull/86

I'd like to use PostGraphQL as a library.
Is it ok to fix only these code?

If there are strange points such as variable name and fixing points, I will fix it soon.
So please point out anything.


benjie:
Consider https://github.com/postgraphql/postgraphql/issues/687#issuecomment-365868029 before merging

benjie:
Thanks so much for the PR; in the end we went with [Smart Comments](https://www.graphile.org/postgraphile/smart-comments/#omitting) to achieve this. My apologies for not closing this along with the associated [graphile-build PR](https://github.com/graphile/graphile-build/pull/86).

PR #613, opened 2017-11-02
chore(docs)

BWillis98:
Fix typos and escape single quotes in comments.


baohuynh27:
Yes

examples/forum/TUTORIAL.md
```diff
@@ -162,8 +162,8 @@ To add comments, just see the SQL below:
 ```sql
 comment on table forum_example.person is 'A user of the forum.';
 comment on column forum_example.person.id is 'The primary unique identifier for the person.';
-comment on column forum_example.person.first_name is 'The person’s first name.';
-comment on column forum_example.person.last_name is 'The person’s last name.';
+comment on column forum_example.person.first_name is 'The person’’s first name.';
+comment on column forum_example.person.last_name is 'The person’’s last name.';
```
benjie: These are typographic quotes so do not require escaping.

BWillis98: What does typographic mean? Nothing came up on Google, and what do you want in place of the double quotes? When I pasted these in from the tutorial they failed.

benjie:
```

benjie: It works fine with the ["right single quotation mark"](http://www.fileformat.info/info/unicode/char/2019/index.htm) (`&rsquo;` in HTML parlance) we already have. I generally refer to these as "typographic quotes"; here's some further reading:

https://practicaltypography.com/straight-and-curly-quotes.html

Adding two of the typographic quotes next to each other will just result in two quotes in the output. In SQL only straight single quotes and double quotes (`'` and `"`) need escaping - typographic quotes (`‘`, `’`, `“` and `”`) are just normal characters like the letter `a` and have no special meaning.

I wonder if you have a browser extension installed that changes typographic quotes to straight quotes when you copy and paste?

BWillis98: Most likely, because they become typographic and that causes the comments to fail. I'll remove those and resubmit the PR.

PR #651, opened 2017-12-18
feat(postgraphql) Add EventListener object to Context object. (#501)

Belline:
This implements the solution described in #501 , allowing Plugins to reach the req/res objects to set Cookies and other uses as well

### DIFF
--- a/src/postgraphql/http/createPostGraphQLHttpRequestHandler.js
+++ b/src/postgraphql/http/createPostGraphQLHttpRequestHandler.js
@@ -66,7 +66,7 @@ const origGraphiqlHtml = new Promise((resolve, reject) => {
  * @param {GraphQLSchema} graphqlSchema
  */
 export default function createPostGraphQLHttpRequestHandler (options) {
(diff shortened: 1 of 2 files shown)


benjie:
Great PR: small, clean and simple. However, I think you could use:

https://github.com/postgraphql/postgraphql/pull/601

To achieve the same goal without requiring any code changes?

chadfurman:
@benjie we need access to the response object, as well.  #601 only provides the request object

benjie:
You should be able to get `res` from `req.res` but feel free to open a PR adding `res` to #601

chadfurman:
TIL we can get res from req

fictorial:
I don't see how req.res is available since this call seems to happen before the graphql is executed.

I don't see a way to set a cookie or other header from within the current flow.  It appears that postgraphile as a lib doesn't expect to be middleware and thus does not call any next function it is given so that option is out.

Is there a standard way to hook into the flow to be able to set a response header based on the graphql response?

PR #655, opened 2017-12-26
Expand jwt verification options

skvale:
Resolves https://github.com/postgraphql/postgraphql/issues/625

`jsonwebtoken`'s VerifyOptions type:
```typescript
export interface VerifyOptions {
    algorithms?: string[];
    audience?: string | string[];
    clockTolerance?: number;
    issuer?: string | string[];
    ignoreExpiration?: boolean;
    ignoreNotBefore?: boolean;
    jwtId?: string;
    subject?: string;
    /**
     *@deprecated
     *@member {string} - Max age of token
     */
    maxAge?: string;
}
```
I left out a maxAge option in Postgraphql because it is deprecated.

If someone has been passing an empty string for jwtAudiences, this would be a breaking change.
`audience: jwtAudiences || ['postgraphql']`


skvale:
It's very possible I'm wrong, but I was thinking the cli was the only exposed api to worry about being a breaking change.

Adding [audiences](https://github.com/postgraphql/postgraphql/pull/655/files#diff-6c38b9313e10fd19f30ddbdd09c18c73R42) did not change.

jwtAudiences is added to [verifyOptions](https://github.com/postgraphql/postgraphql/pull/655/files#diff-6c38b9313e10fd19f30ddbdd09c18c73R42) and is passed through `createPostGraphQLHttpRequestHandler` to `withPostGraphQLContext` with the other available options.

but yes, I could make those changes if that isn't right.

skvale:
Good call, thanks for the review and direction.
I addressed those things 👍

src/postgraphql/cli.ts
```diff
@@ -38,7 +39,15 @@ program
   .option('-l, --body-size-limit <string>', 'set the maximum size of JSON bodies that can be parsed (default 100kB) The size can be given as a human-readable string, such as \'200kB\' or \'5MB\' (case insensitive).')
   .option('--secret <string>', 'DEPRECATED: Use jwt-secret instead')
   .option('-e, --jwt-secret <string>', 'the secret to be used when creating and verifying JWTs. if none is provided auth will be disabled')
-  .option('-A, --jwt-audiences <string>', 'a comma separated list of audiences your jwt token can contain. If no audience is given the audience defaults to `postgraphql`', (option: string) => option.split(','))
+  .option('-A, --jwt-audiences <string>', 'DEPRECATED Use jwt-verify-audience instead', (option: string) => option.split(','))
+  .option('--jwt-verify-algorithms <string>', 'a comma separated list of the names of the allowed jwt token algorithms', (option: string) => option.split(','))
+  .option('--jwt-verify-audience <string>', 'a comma separated list of audiences your jwt token can contain. If no audience is given the audience defaults to `postgraphql`', (option: string) => option.split(','))
```
benjie: Any reason to not move the `-A` to the `--jwt-verify-audience` entry? I'm assuming if we did so the behaviour for people who use `-A` would be identical, and that way the `-A` option doesn't need to be deprecated?

benjie:
This is good to merge (avoid rebasing now please 👍 ) - just wanted to discuss the `-A` option (above) before the final merge.

src/postgraphql/cli.ts
skvale: I don't have a reason, just didn't think about doing it.  Yeah, it should work.

benjie: Awesome; move it when you get a chance then we can merge👍

Issue #657, opened 2017-12-30
PGPASS Environment Variable Support


antmldr:
I'm submitting a ... 
- [ ] bug report
- [x] feature request
- [ ] question

It'd be useful for postgraphql to be able to utilize the PGPASS environment variable, if present, when building its connection string (see: [cli.ts](https://github.com/postgraphql/postgraphql/blob/master/src/postgraphql/cli.ts#L102))

Is this something appropriate for v3 or v4?
Would it be something the project would be open to a PR for? I'd be happy to implement & test it.

benjie:
PR welcome; do it against v3 only (master) - I’ll merge to v4 later. Ensure it still functions correctly if the envvar is not set (e.g. check `{username: undefined}` is treated the same as if the key were omitted: `{}` (failing that, use Object.assign or similar  to only add each key if the envvar is set)

Thanks!

antmldr:
Awesome, thanks for merging & your guidance here!

PR #658, opened 2017-12-31
feat(postgraphql): support PGUSER and PGPASSWORD environment variables

antmldr:
As per #657, it'd be be useful for postgraphql to be able to utilize the PGUSER and PGPASSWORD environment variables, if present, when building its connection string.

Testing this change, the behavior of setting these environment variables maintains consistency with other environment variables (e.g. PGPORT) which the project currently supports, e.g.:

- Setting PGUSER and PGPASSWORD won't clobber existing parts of the connection string if they are defined,
- If there is a connection string, but it fails to mention a user or password, it will use environment variables, if available. If env vars are not available, it will use defaults,
- In the case where a connection string is not defined, if there are environment variables, it will use them, and,
- In a case where nothing is defined (no env vars, no connection string), the application uses its defaults and can successfully connect to a default postgres server.

Other:
- It appears to deal well with special characters as well - didn't have any issues with passwords or users containing @'s or :'s.
- In #657 I erroneously referred to it as "PGPASS" - as per psql documentation, this commit uses the correct varname, "PGPASSWORD".
- Tested on OSX and Debian


benjie:
Travis CI is having some issues right now. Hopefully things improve soon!

PR #667, opened 2018-01-19
feat(postgraphile) Support .postgraphilerc.js and DATABASE_URL

syndesis:
- Updated name of program in help section
- Added support for a .postgraphilerc.js configuration file
- Included sample.postgraphilerc.js in examples
- Added support for DATABASE_URL with connection string
- Added .idea to gitignore

### DIFF
--- a/src/postgraphql/cli.ts
+++ b/src/postgraphql/cli.ts
@@ -11,6 +11,15 @@ import postgraphql from './postgraphql'
 
 // tslint:disable no-console
 
+let config = {
+  options: {},
+}
+try {
+  config = require(process.cwd() + '/.postgraphilerc.js') // tslint:disable-line no-var-requires
+} catch (error) {
+  // Use command line options
+}
+
 // TODO: Demo Postgres database
 const DEMO_PG_URL = null
 
@@ -62,8 +71,8 @@ program
 program.on('--help', () => console.log(`
   Get Started:
 
-    $ postgraphql --demo
-    $ postgraphql --schema my_schema
+    $ postgraphile --demo
+    $ postgraphile --schema my_schema
 `.slice(1)))
 
 program.parse(process.argv)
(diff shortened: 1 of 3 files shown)


src/postgraphql/cli.ts
```diff
@@ -11,6 +11,15 @@ import postgraphql from './postgraphql'
 
 // tslint:disable no-console
 
+let config = {
+  options: {},
+}
+try {
+  config = require(process.cwd() + '/.postgraphilerc.js') // tslint:disable-line no-var-requires
```
benjie: If we remove the `.js` extension here then both .js and .json will be supported

src/postgraphql/cli.ts
```diff
@@ -113,12 +123,56 @@ const {
...
+  showErrorStack = showErrorStack,
+  extendedErrors = extendedErrors,
+  bodySizeLimit = bodySizeLimit,
+  appendPlugins: appendPluginNames = appendPluginNames,
+  prependPlugins: prependPluginNames = prependPluginNames,
+  // replaceAllPlugins is NOT exposed via the CLI
+  readCache = readCache,
+  writeCache = writeCache,
+// tslint:disable-next-line no-any
+} = program as any)
```
benjie: I think a much easier way of achieving this might be:

benjie: I think a much easier way of achieving this might be:

```js
} = Object.assign({}, config['options'], program) as any
```

src/postgraphql/cli.ts
```diff
@@ -113,12 +123,56 @@ const {
   readCache,
   writeCache,
 // tslint:disable-next-line no-any
-} = program as any
+} = config['options'] as any
```
benjie: This will throw if `config.options` isn't set (e.g. if you source an RC that is just `{}`); however, see recommendation below to fix this and remove a load of code in one fell swoop.

syndesis:
Closing PR in favor of new one with requested changes on different branch.

PR #674, opened 2018-01-30
chore: Solve TypeScript issues on CI

benjie:
Latest versions of lodash typings define:

```
    interface WeakMap<K extends object, V> { }
```

which is in line with the latest TypeScript compiler. We, however, run a much older version of TypeScript, so the above causes a clash.

Also it uses the `object` type which we didn't have.

Also @types/commander now references the commander module directly; we don't support that so I've limited the version of the types file.

Also GraphQL v0.12 has different error messages and different schema comment formatting to v0.11, so the tests have been updated to account for this.

This PR should resolve these issues without causing any changes to the compiled code.


benjie:
Due to a peer dependency issue with codemirror-graphql which I cannot solve without a breaking change (requiring that the graphql version is <0.12) we no longer support Node v4 on CI.

PR #675, opened 2018-01-30
docs(TUTORIAL): Revoke public grant on function execution before function definitions

mattbretl:
I moved `alter default privileges..` to an earlier spot in the tutorial.  The way it's currently written, once you reach the grants section, it's too late to `alter default privileges`.

I also rebuilt the schema.sql script.


benjie:
Thanks Matt! Are you working on some presentations or workshops or similar that reference this?

PR #680, opened 2018-02-02
Remove middle layers from docker image

jakubknejzlik:
This should decrease the docker image, because each RUN command creates new layer in docker image.
Image size reduction 278MB -> 142MB.


benjie:
Hey @angelosarto can you sign off on this?

benjie:
Thanks @jakubknejzlik, @angelosarto! 🙏

PR #690, opened 2018-02-19
fix(docs): Fix pg-pool import in library.md

petetnt:
Another tiny documentation fix: The `Pool` import is the default/unnamed export of the standalone `pg-pool` module as seen in the [docs](https://github.com/brianc/node-pg-pool/blob/master/README.md).

Signed-off-by: petetnt <pete.a.nykanen@gmail.com>

### DIFF
--- a/docs/library.md
+++ b/docs/library.md
@@ -92,12 +92,12 @@ createPostGraphQLSchema('postgres://localhost:5432')
 
 Now that you have your schema, in order to execute a GraphQL query you will need to get a PostGraphQL context object with `withPostGraphQLContext`. The context object will contain a Postgres client which has its own transaction with the correct permission levels for the associated user.
 
-You will also need a Postgres pool from the [`pg-pool`][] module.
+You will also need a Postgres pool from the [`pg`][] module.
 
 `withPostGraphQLContext`, like `createPostGraphQLSchema`, will also return a promise.
 
 ```js
-import Pool from 'pg-pool'
+import { Pool } from 'pg'
 import { graphql } from 'graphql'
 import { withPostGraphQLContext } from 'postgraphql'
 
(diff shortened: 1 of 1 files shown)


benjie:
I think we should just be doing `import { Pool } from 'pg'` actually, and not mentioning `pg-pool` at all.

benjie:
https://node-postgres.com/api/pool

petetnt:
Update in https://github.com/postgraphql/postgraphql/pull/690/commits/b5deac0b8376ba22424d4b4ac3cf32e1548593c7.

I wonder if there are potential possible version mismatch issues? Should `postgraph` expose its own `Pool` dependency, or should the method just accept `Pool` configuration options instead of the whole pool (is not reusing the same pool an issue)?

docs/library.md
```diff
@@ -145,7 +145,7 @@ Arguments include:
 This function sets up a PostGraphQL context, calls (and resolves) the callback function within this context, and then tears the context back down again finally resolving to the result of your function. The callback is expected to return a promise which resolves to a GraphQL execution result. The context you get as an argument to `callback` will be invalid anywhere outside of the `callback` function.
 
 - **`options`**: An object of options that are used to create the context object that gets passed into `callback`.
-  - `pgPool`: A required instance of a Postgres pool from [`pg-pool`][]. A Postgres client will be connected from this pool.
+  - `pgPool`: A required instance of a Postgres pool from [`https://www.npmjs.com/package/pg`][]. A Postgres client will be connected from this pool.
```
benjie: Should this just be `pg`?

benjie:
We do duck-typing on Pool now (as of Saturday 😂 ) so it shouldn't be as much of an issue as before.

PR #723, opened 2018-03-20
added handleErrors option

pyramation:


src/postgraphile/postgraphile.ts
```diff
@@ -45,6 +46,13 @@ type PostGraphileOptions = {
   // `json` (which causes the stack to become an array with elements for each
   // line of the stack).
   showErrorStack?: boolean,
+  // Enables ability to modify errors before sending them down to the client
+  // optionally can send down custom responses
```
benjie: Please add `/* @middlewareOnly */` as it only affects the HTTP middleware (not the createPostGraphileSchema function)

src/postgraphile/http/createPostGraphileHttpRequestHandler.js
```diff
@@ -544,8 +544,14 @@ export default function createPostGraphileHttpRequestHandler(options) {
       if (result && result.errors)
         result.errors = result.errors.map(formatError)
 
-      res.setHeader('Content-Type', 'application/json; charset=utf-8')
-      res.end(JSON.stringify(result))
+      if (options.handleErrors && result.errors) {
+        result.errors = options.handleErrors(result.errors, req, res)
+      }
```
benjie: I'd prefer that `handleErrors` *replaced* `formatError` so that it has access to the GraphQLError directly - once it's gone through `formatError` some content is lost. So something more like:

src/postgraphile/postgraphile.ts
```diff
@@ -45,6 +46,13 @@ type PostGraphileOptions = {
...
+  handleErrors?: ((
+    errors: Array<GraphQLErrorExtended>,
```
benjie: As above, this should be `Array<GraphQLError>` - handling the unmangled errors

benjie: We should add validation that throws an error if `handleErrors` is set at the same time as either `showErrorStack` or `extendedErrors`.

lifeiscontent: @benjie where thouls the validation be created?

benjie: Here's an example of a validation that just outputs a warning:

https://github.com/graphile/postgraphile/blob/d0283d5535d542838af9a69c16889cceb46a8f81/src/postgraphile/postgraphile.ts#L156-L162

Putting it around there would make sense to me. This one should throw though 👍

lifeiscontent: @benjie how's this?

``` ts
  if (options.handleErrors && (options.extendedErrors || options.showErrorStack)) {
    throw Error('cannot use handleErrors with other error options')
  }
```

benjie: throw new Error('cannot use handleErrors with other error options')

lifeiscontent:
@benjie updated!

src/postgraphile/http/__tests__/createPostGraphileHttpRequestHandler-test.js
```diff
@@ -466,6 +466,54 @@ for (const [name, createServerFromHandler] of Array.from(serverCreators)) {
...
+        })
+    })
+
+    test('will allow user to send custom responses when handleErrors is set and sends a response', async () => {
+      pgPool.connect.mockClear()
+      pgClient.query.mockClear()
+      pgClient.release.mockClear()
+      const server = createServer({
+        handleErrors: (errors, req, res) => {
+          res.statusCode = 401
```
benjie: This should `return errors` - tests can act as demo code.

lifeiscontent:
@benjie updated.

Issue #634, opened 2017-11-20
Feature Request: Batch query support


zopf:
I'm submitting a ... 
- [ ] bug report
- [x] feature request
- [ ] question 

Some other GraphQL server implementations support a batch query protocol, whereby multiple GraphQL requests can be assembled into an array, the array is sent to the server, each of the requests is separately executed on the server, and their results are returned in one array to the client.

Example usage can be found here for Apollo: https://blog.graph.cool/improving-performance-with-apollo-query-batching-66455ea9d8bc

And an example implementation in Apollo server can be found here: https://github.com/apollographql/apollo-server/blob/3fecfcd3218c2f7f2c5ed59a351d71022944bc35/packages/apollo-server-core/src/runHttpQuery.ts#L65

This feature request proposes that PostGraphQL support a similar scheme, in order to allow clients to optionally make GraphQL requests to the PostGraphQL server in a batch mode, enabling multiple operations to be performed with a single HTTP request.

An example use case would be a client that needs to perform ten separate mutations (say, updating ten separate DB objects with new data).  With batching, instead of sending ten separate HTTP requests to the PostGraphQL server, the client could send one larger HTTP request that contains an array of those ten mutations, and receive and array of ten results.

benjie:
I definitely want to support this and it should be fairly easy to add. PRs welcome.

(In the mean time you can work around this by performing multiple mutations in the same GraphQL query:

```graphql
mutation MultiCreate {
  post1: createPost(...) { ...Payload }
  post2: createPost(...) { ...Payload }
  post3: createPost(...) { ...Payload }
  post4: createPost(...) { ...Payload }
  post5: createPost(...) { ...Payload }
}
fragment Payload on CreatePostPayload {
  ...
}
```
)

benjie:
![](https://github.trello.services/images/mini-trello-icon.png) [Query batching](https://trello.com/c/kxA9Is4o/9-query-batching)

w9:
@benjie Does this workaround guarantees atomicity (my guess is not)? And if not how do I simulate the atomicity of

```
INSERT INTO table (c1, c2) VALUES (1,2),(3,4),(5,6);
```

?

w9:
To answer myself---I found the following solution works:

```
create type add_many_people_input as (
  name text,
  age numeric
);

create function add_many_people(
  inputs add_many_people_input[]
)
  returns void
language plpgsql security definer as $$
begin
  insert into person (name, age)
    select * from unnest(inputs);
[...]

PR #734, opened 2018-04-10
feat(postgraphile): adding query batching #634

Trupal00p:
I've modified the HTTP handler to accept an array of queries for query batching as discussed in #634. 

Quick summary of changes: I extracted out the query running logic into its own function then map it over the incoming queries if they are in an array.

There are a couple things in this change to take note of and change if there's a better approach.  So looking for input on these:

1. This changes how errors are returned for individual queries.  Query errors will always return with a 200 status code and with the ususal "errors" entry in the returned object. This is so that if a single query fails the rest do not.  I've also added a "status" entry with the old status code in the returned error data object , not sure if this is even helpful.

2.  I've moved some of the logging steps around not sure if this will have a negative impact on performance.

3.  I feel like this could be refactored further, but wanted to get something working out there.

### DIFF
--- a/src/postgraphile/http/__tests__/createPostGraphileHttpRequestHandler-test.js
+++ b/src/postgraphile/http/__tests__/createPostGraphileHttpRequestHandler-test.js
@@ -234,7 +234,7 @@ for (const [name, createServerFromHandler] of Array.from(serverCreators)) {
       await request(server)
         .post('/graphql')
         .send({ query: '{' })
-        .expect(400)
+        .expect(200)
         .expect('Content-Type', /json/)
         .expect({
           errors: [
@@ -251,7 +251,7 @@ for (const [name, createServerFromHandler] of Array.from(serverCreators)) {
       await request(server)
         .post('/graphql')
         .send({ query: '{notFound}' })
-        .expect(400)
+        .expect(200)
         .expect('Content-Type', /json/)
         .expect({
           errors: [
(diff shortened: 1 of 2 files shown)


benjie:
Running `npm run lint -- --fix` should fix most of the linting errors for you. Will attempt to find time to review in the next few days; if you've not heard anything by Monday feel free to give me a nudge.

benjie:
Errors are now mostly `ReferenceError: handleErrors is not defined`

Trupal00p:
Horray tests are passing now :) thank for the pointers and for working on this awesome project.

I think I completely misunderstood how the graphql error handling was supposed to work.  This should be a bit cleaner.  

For now, I've removed the status entry that I mention in point 1 above.  Might make sense to add as an extesion to the error.

Trupal00p:
A bit more info to support the approach regarding no 1 above.  The Apollo server implementation seems to return 200 for most batched requests as well.  Saving 4XX responses for when there is something wrong with the whole request.

Thier batching implementation can be seen here: [https://github.com/apollographql/apollo-server/blob/9e8984c2ee977f75e4d4bdebb295fe5826e5518e/packages/apollo-server-core/src/runHttpQuery.ts](https://github.com/apollographql/apollo-server/blob/9e8984c2ee977f75e4d4bdebb295fe5826e5518e/packages/apollo-server-core/src/runHttpQuery.ts)

benjie:
Thanks for the PR; I'm afraid I had to go a different way due to other changes I was making to core (plus I wanted query batching to be opt-in via CLI) but I appreciate the nudge!

Batched queries support is being added via https://github.com/graphile/postgraphile/pull/749

Trupal00p:
No worries, glad it's getting in there somehow!

Issue #623, opened 2017-11-10
Time out for post graphql server


mamtalakhani:
Can you plz let me know where to set the timeout for postgraphql server.

benjie:
What timeout are you looking for?

mamtalakhani:
We have a query which runs for a few minutes in Postgres. When the same query is written in GraphiQL,the server responds with 504 Gateway timeout exactly after 1.5 min. We would like to change the 1.5 min timeout variable.

benjie:
I'm not aware of this limit (that isn't to say it doesn't exist). The error "Gateway Timeout" implies that you have something between the PostGraphQL server and the client (aka a "Gateway") and that that is causing the connection to time out. On Heroku this period is 30 seconds.

benjie:
This was added by the above commit 🙏

PR #741, opened 2018-04-23
feat(server): add server timeout configuration #623

kronken:
I added the ability to set timeout for socket connections.


src/postgraphile/cli.ts
```diff
@@ -422,6 +424,8 @@ if (noServer) {
     const rawMiddleware = postgraphile(pgConfig, schemas, postgraphileOptions)
     const middleware = pluginHook('cli:server:middleware', rawMiddleware, { options: postgraphileOptions })
     const server = createServer(middleware)
+    server.timeout = serverTimeout
```
benjie: I'm not super comfortable using a property access to `timeout` when there's a setter method `setTimeout` present; however I'm struggling to tell from the docs if they do the same thing or not:

screenshot 2018-05-03 17 33 43

Can you point to someone of authority stating that these are either equivalent, or that they're different and why you'd use one over the other?

benjie: Hey @kronken, any idea regarding this? Are they different?

benjie: Judging by the Node.js source they're equivalent:

https://github.com/nodejs/node/blob/79e63c5fc727254e3d1299c7d062b6caca6e9a86/lib/_http_server.js#L310-L315

PR #763, opened 2018-05-20
feat(postgraphql): add support for pre-request function #760

### LINKED ISSUE #760: pre-request stored procedures
I'm submitting a ... 

* [ ] bug report
* [x] feature request
* [ ] question 

PostGraphile version:

4.0.0-beta.8

I'm coming from PostgREST and finding Postgraphile and GraphQL really nice to work with. Thanks heaps for this project!

One of the features I'm missing from PostgREST is the "pre-request" option which takes a stored procedure that is executed at the start of every request. It is called after JWT processing and `set role` are completed but before any data is fetched. This feature permits the developer to easily hook into the PostgREST call flow.

Reference: https://postgrest.org/en/v4.4/auth.html#custom-validation
Discussions: https://github.com/PostgREST/postgrest/pull/632 https://github.com/PostgREST/postgrest/issues/631

Can we have a similar function in Postgraphile?

There are several authentication use cases where this will be useful, checking if a user is active, blacklisted JWT, etc.

Another good use case is to securely set the schema `search_path` for multi-tenant applications using Postgres schema layering.

Example schemas in the DB:
- `client1` -- contains tables + data for this client
- `client2` -- contains tables + data for this client
- `base` -- contains all stored procedures and empty tables 
- `public` -- contains pgcrypto, etc functions

Default `search_path` is `base,public`

If JWT contains a claim `clientname: client1`, the pre-request procedure can query the JWT claim and `set_config('search_path', 'client1,base,public', false) `. All data is now fetched from client1 schema.

Thanks!

mlipscombe:
Resolves #760 

Adds support for a --pre-request option that calls the named stored procedure after setting the JWT claims and the role.


src/postgraphile/cli.ts
```diff
@@ -82,6 +82,7 @@ program
   .option('-p, --port <number>', 'the port to be used. Defaults to 5000', parseFloat)
   .option('-m, --max-pool-size <number>', 'the maximum number of clients to keep in the Postgres pool. defaults to 10', parseFloat)
   .option('-r, --default-role <string>', 'the default Postgres role to use when a request is made. supercedes the role used to connect to the database')
+  .option('--pre-request <string>', 'the Postgres identifier for a function that will be called before each request')
```
benjie: Let's rename it to `--transaction-validation-function` as that's effectively its purpose.

typings/pg-sql2.d.ts
```diff
@@ -8,4 +8,5 @@ declare module 'pg-sql2' {
   export function query(strings: TemplateStringsArray, ...values: Array<SQLNode | SQLQuery>): SQLQuery
   export function value(val: mixed): SQLNode
   export function join(items: Array<SQLNode | SQLQuery>, separator?: string): SQLQuery
+  export function raw(text: string): SQLNode
```
benjie: I'd rather avoid `raw` entering our codebase.

src/postgraphile/withPostGraphileContext.ts
```diff
@@ -240,6 +247,16 @@ async function setupPgClientTransaction({
     await pgClient.query(query)
   }
 
+  if (typeof preRequest === 'string') {
+    const query = sql.compile(sql.query`select ${sql.raw(`${preRequest}()`)}`)
```
benjie: Just use:

```js
const query = `select ${preRequest}()`
```

(That way we can avoid using `sql.raw`, plus it's faster since it doesn't need to compile from an AST.)

benjie:
Having thought about this a bit more; this should be done via a plugin rather than as core functionality; we can evaluate the hooks it needs as part of #777

PR #789, opened 2018-06-27
fix(koa): don't compress assets via send

benjie:
Koa sometimes seems to hang when loading GraphiQL; I've not been able to deliberately reproduce it. This seems to be solving it so far... Really we ought to replace `sendFile` with something Koa-friendly.

### DIFF
--- a/src/postgraphile/http/createPostGraphileHttpRequestHandler.js
+++ b/src/postgraphile/http/createPostGraphileHttpRequestHandler.js
@@ -426,6 +426,9 @@ export default function createPostGraphileHttpRequestHandler(options) {
 
         // Sends the asset at this path. Defaults to a `statusCode` of 200.
         res.statusCode = 200
+        if (req._koaCtx) {
+          req._koaCtx.compress = false
+        }
         await new Promise((resolve, reject) => {
           const stream = sendFile(req, assetPathRelative, {
             index: false,


sjmcdowall:
I guess I would ask is why is GraphiQL or anything in the chain doing the compression and not allowing any registered compress() type middleware do it?

benjie:
Nothing's doing compression (except the Koa middleware of course); the issue is that `send` writes the data straight to the `res` response object (including setting the relevant headers/etc) which works just fine with Express because it's based on Connect based on http.createServer and the response cycle ends there. However Koa takes more control and doesn't like it when middleware writes directly to the response object - when we return control to Koa it wants the data to not already have been written. The Koa compress middleware has even stronger expectations than Koa itself, and these expectations are broken by most things that write directly to `res`, as PostGraphile does currently.

hengnee:
I think the same thing is happening for Fastify as it seems to skip the Fastify Routes and fastify.compress plugin don't seem to be able to compress the output from PostGraphile.

benjie:
Please submit a PR with a failing test for your setup.

hengnee:
> Please submit a PR with a failing test for your setup.

Hi, I'm so sorry but may I provide a link for the failing test which has comments regarding the issue instead?

https://github.com/hengnee/rectify_fastify_postgraphile

benjie:
If you file a PR it’ll be addressed faster; but if you can’t, please file and issue with the above otherwise your request will be lost.

Issue #520, opened 2017-07-19
How to import postgraphql in a TypeScript project?


zopf:
I am fairly new to TypeScript, so forgive me if I'm making a newbie mistake.

I am attempting to import the `postgraphql` module in my `index.ts` file, like so:

```
import * as postgraphql from 'postgraphql';
```

TypeScript is telling me that `Module ''postgraphql'' resolves to a non-module entity and cannot be imported using this construct.`

For my other imports, I installed their typings using e.g. `npm install --save-dev @types/underscore`, but it does not appear there is a `@types/postgraphql` module.  These typings come from https://github.com/DefinitelyTyped/DefinitelyTyped.

How do I go about setting up the proper typings for the `postgraphql` module so that I can use it in a TypeScript project?

ar4mirez:
Did you tried just doing:
```javascript
import postgraphql from 'postgraphql'
```
You would do `import * as postgraphql from 'postgraphql';` if postgresql exports multiple keys.

Hope this helps. 🤞

zopf:
I did indeed try that, but then it's an even simpler error: `Cannot find module 'postgraphql'.`

That is the same type of error that appears when I have not installed the typings for a module using e.g. `npm install --save-dev @types/underscore`.

benjie:
Try

```
import postgraphql = require('postgraphql')
```

You can also add typings yourself; e.g. here's some I had to make for v4:

https://github.com/postgraphql/postgraphql/blob/1417bc37b62bae165564eaf8e844c32b312a4c4e/typings/postgraphql-build.d.ts

We're probably going to be moving from TypeScript to Flow at some point, FYI.

zopf:
Still get the same `Cannot find module 'postgraphql'.` error with the `require` statement, plus another error because I have my `tsconfig.module` set to `es2015`;)

I'm making my own typings now.

Based upon this spec ( http://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html ) it seems like we should probably either have a `types` field in the `package.json`.  I'll make a PR...

zopf:
Ok hm I'm not going to make a PR just yet because, well, I don't know enough about postgraphql's build process and structure to put the definitions in the right place.  My guess is that it should go in `index.d.ts` at the root, and the build script / npm deploy script should be updated to include it in the npm package, but please let me know where it should actually go.

Here is the type definition that I'm testing out in my project now, which seems to compile properly:

```
declare module 'postgraphql' {

[...]

PR #794, opened 2018-07-07
feat(types): ship automatic TypeScript typings

cdaringe:
# problem statment

- TS typings are out of date
- TS typings don't ship by default

partially address https://github.com/graphile/postgraphile/issues/520

# solution

- have this TS project auto-build its typing artifcacts
    - achieve this theoretically via a 3 LOC code change in the tsconfig.json
    - address the cascading set of small issues associated with ^


package.json
```diff
@@ -34,6 +34,9 @@
     "prepack": "./scripts/build"
   },
   "dependencies": {
+    "@types/pg": "^7.4.10",
+    "@types/graphql": "^0.8.2",
+    "@types/jsonwebtoken": "<7.2.1",
```
cdaringe: because our interfaces expose types that use these types, the Typescript documentation states to bundle them as dependencies.  i was mixed feelings on this!

package.json
```diff
@@ -75,8 +76,8 @@
     "source-map-support": "^0.4.6",
     "supertest": "^2.0.1",
     "ts-node": "^2.0.0",
-    "tslint": "^4.2.0",
-    "typescript": "2.1.5"
+    "tslint": "^5.10.0",
```
cdaringe: so we can disable new rules.  will discuss later

scripts/lint
```diff
@@ -4,4 +4,4 @@ npm_bin=$(npm bin)
 
 set -e
 
-$npm_bin/tslint $(find src -name "*.[tj]s" -a \! -wholename '*/node_modules/*') $@
+$npm_bin/tslint --format verbose $(find src -name "*.[tj]s" -a \! -wholename '*/node_modules/*') $@
```
cdaringe: adds rule names to output, essentially.

src/interfaces.ts
```diff
@@ -0,0 +1,213 @@
+import { GraphQLErrorExtended } from './postgraphile/extendedFormatError'
+import { PluginHookFn } from './postgraphile/pluginHook'
+import { IncomingMessage, ServerResponse } from 'http'
+import { GraphQLSchema } from 'graphql'
+import { Pool } from 'pg'
+import jwt = require('jsonwebtoken')
+import { GraphQLError } from 'graphql/error'
+
+export namespace Postgraphile {
```
cdaringe: types that were used in various places i moved into a name space.  for instance, `PostGraphileOptions` was almost fully duped in 2 places.  i defined it once, imported it back two places, and extended it as needed

src/postgraphile/cli.ts
```diff
@@ -11,10 +11,13 @@ import { Pool } from 'pg'
...
 import { makePluginHook, PostGraphilePlugin } from './pluginHook'
 import debugFactory = require('debug')
+import { Postgraphile } from '../interfaces'
+import mixed = Postgraphile.mixed
 
 const debugCli = debugFactory('postgraphile:cli')
 
-// tslint:disable no-console
+// tslint:disable no-console no-shadowed-variable
+// no-shadowed-variable is erroroneously reporting
```
cdaringe: im betting the older tslint didnt do this :/  it was claiming that `plugins` in a typedef was shadowed.  infact, it was just annotating the type.  bonkers AST parse.  as you can see, i didn't modify anything to do w/ a variable called `plugins` in this file.

src/postgraphile/http/createPostGraphileHttpRequestHandler.d.ts
```diff
@@ -1,101 +0,0 @@
-import { IncomingMessage, ServerResponse } from 'http'
```
cdaringe: all of this content was trashed or moved into `interfaces.d.ts`, as referenced above

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -1,30 +1,49 @@
 /* eslint-disable */// Because we use tslint
+/* tslint:disable:no-any no-var-requires */
```
cdaringe: and here's the real only significant offender of this PR.  to get the build to auto-populate typings, `allowJs` must be set to false.  this was the lone `.js` in the lib.

this file could use have used some deep, deep tidy before i put my fingers on it.  do to the size and complexity of it, rather than put proper types on everything in the true TS spirit, I declared _many_ of the internals as `any`.  this was to prevent any major refactor or create _even more change_ than we already want for this PR.

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -1,30 +1,49 @@
...
+
+/**
+ * Creates a GraphQL request handler that can support many different `http` frameworks, including:
+ *
+ * - Native Node.js `http`.
+ * - `connect`.
+ * - `express`.
+ * - `koa` (2.0).
+ */
+export interface ICreateRequestHandler extends Postgraphile.PostGraphileOptions {
```
cdaringe: boom! that huge red-file (`createPostGraphileHttpRequestHandler.d.ts`), gets compacted down to these 9 LOC. mo betta, much betta.

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -183,19 +193,20 @@ export default function createPostGraphileHttpRequestHandler(options) {
 
     // If the user wants to see the error’s stack, let’s add it to the
     // formatted error.
+    error.stack = error.stack || '[empty stack]'
     if (options.showErrorStack)
       formattedError.stack =
-        options.showErrorStack === 'json'
+        options.showErrorStack
```
cdaringe: 👀 needs discussion.  `showErrorStack` is a bool everywhere else, and the lib wouldn't compile with the above conditional.  is this really a supported feature?  it looks like in both places, the stack makes it through.

im not sure of the intent.

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -221,22 +232,22 @@ export default function createPostGraphileHttpRequestHandler(options) {
...
+        parent(req, res, (error: Error) => {
           if (error) {
             return next(error)
           }
           fn(req, res, next)
         })
       }
     },
-    (req, res, next) => next(),
+    (req: any, res: any, next: any) => next(),
```
cdaringe: honestly, i dont know how this is working at all right now (before i touched it, or after).   i can't find `next` on 245  in scope?  our initial value is a function which calls `next`, which i dont see defined anywhere.

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -430,7 +438,7 @@ export default function createPostGraphileHttpRequestHandler(options) {
           req._koaCtx.compress = false
         }
         await new Promise((resolve, reject) => {
-          const stream = sendFile(req, assetPathRelative, {
+          sendFile(req, assetPathRelative, {
```
cdaringe: ya ya ya.  the linter made me do all of this pedantic junk.  im sorry!

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -742,7 +749,7 @@ export default function createPostGraphileHttpRequestHandler(options) {
 
       res.setHeader('Content-Type', 'application/json; charset=utf-8')
 
-      res.end(JSON.stringify(returnArray ? results : results[0]))
+      res.end(JSON.stringify(returnArray ? results : results![0]))
```
cdaringe: this is a new TS feature i learned recently.  assert that <preceding_var> is defined!

src/postgraphile/pluginHook.ts
```diff
@@ -1,9 +1,10 @@
 /* eslint-disable */// Because we use tslint
```
cdaringe: only change below is changing where types are imported from

src/postgraphile/postgraphile.ts
```diff
@@ -1,188 +1,18 @@
 import { Pool, PoolConfig } from 'pg'
```
cdaringe: only changes below are changing where types are imported from, and migrating these types to a common interfaces file

tsconfig.json
```diff
@@ -14,10 +16,9 @@
     "suppressImplicitAnyIndexErrors": true,
     "strictNullChecks": true,
     "noFallthroughCasesInSwitch": true,
-    "noUnusedParameters": true,
+    "noUnusedParameters": false,
```
cdaringe: because so often we have `(req, res, next)` where `res` isn't used

tsconfig.json
```diff
@@ -1,6 +1,8 @@
 {
   "compilerOptions": {
-    "allowJs": true,
+    "declaration": true,
```
cdaringe: 🎉

typings/mixed.d.ts
```diff
@@ -1,18 +0,0 @@
-/**
```
cdaringe: 👋 bye bye

typings/object.d.ts
```diff
@@ -1,4 +0,0 @@
-/**
```
cdaringe: 👋 adddddiiioos!

cdaringe:
a net -1K LOC!  i say that's a win :)

src/interfaces.ts
benjie: Capital g please 🤗

src/interfaces.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  /**
+   * A request handler for one of many different `http` frameworks.
+   */
+  export interface HttpRequestHandler {
+    (req: IncomingMessage, res: ServerResponse, next?: (error?: mixed) => void): void
+    (ctx: { req: IncomingMessage, res: ServerResponse }, next: () => void): Promise<void>
+    getGraphQLSchema: () => Promise<GraphQLSchema>
+    formatError: (e: GraphQLError) => GraphQLError
+    pgPool: Pool
```
benjie: I think this has more stuff on it now, I should check.

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
benjie: I know how that feels. It’s one of the function arguments: req, res, NEXT 🙃

Basically this is a no-op middleware we use so we can efficiently combine the other middlewares into one middleware. It helps the JIT.

src/interfaces.ts
benjie: https://github.com/graphile/postgraphile/blob/master/src/postgraphile/http/createPostGraphileHttpRequestHandler.js#L813-L816

src/postgraphile/http/__tests__/createPostGraphileHttpRequestHandler-test.js
```diff
@@ -1,855 +0,0 @@
...
-      expect(additionalGraphQLContextFromRequest).toHaveBeenCalledTimes(1)
-      expect(
-        additionalGraphQLContextFromRequest.mock.calls[0][0],
-      ).toBeInstanceOf(http.IncomingMessage)
-      expect(
-        additionalGraphQLContextFromRequest.mock.calls[0][1],
-      ).toBeInstanceOf(http.ServerResponse)
-    })
-  })
-}
```
benjie: I don't think you've replaced this test file after deleting it?

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -36,12 +55,12 @@ const sendFile = require('send')
 const LRU = require('lru-cache')
 const crypto = require('crypto')
 
-const calculateQueryHash = queryString => crypto.createHash('sha1').update(queryString).digest('base64')
+const calculateQueryHash = (query: string) => crypto.createHash('sha1').update(query).digest('base64')
```
benjie: I prefer `queryString: string` since "query string" is what I would call it aloud. "query" has a different meaning to me.

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -94,15 +111,14 @@ const origGraphiqlHtml = new Promise((resolve, reject) => {
  * We need to be able to share the withPostGraphileContext logic between HTTP
  * and websockets
  */
-const withPostGraphileContextFromReqResGenerator = options => {
+function withPostGraphileContextFromReqResGenerator(options: ICreateRequestHandler): any {
```
benjie: Rather than `any`, consider using something like:

```ts
function withPostGraphileContextFromReqResGenerator<T>(options: ICreateRequestHandler): (req: IncomingMessage, res: ServerResponse, moreOptions: any, fn: (pgClient: PoolClient) => T) => T
```

You can use `any` instead of `T` if you like. Effectively whatever is returned by the callback function `fn` will be returned by the function returned from withPostGraphileContextFromReqResGenerator

src/interfaces.ts
```diff
@@ -0,0 +1,213 @@
...
+    // [experimental] Plugin hook function, enables functionality within
+    // PostGraphile to be expanded with plugins. Generate with
+    // `makePluginHook(plugins)` passing a list of plugin objects.
+    /* @middlewareOnly */
+    pluginHook?: PluginHookFn,
+    // Should we use relay pagination, or simple collections?
+    // "omit" (default) - relay connections only,
+    // "only" - simple collections only (no Relay connections),
+    // "both" - both
+    simpleCollections?: 'omit' | 'both' | 'only',
```
benjie: We should consider adding `[propName: string]: any` here because PostGraphile plugins can add additional options to this object (e.g. the supporter plugin adds `simpleSubscriptions` as an option)

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
benjie: This was before my time, but I believe there used to be an option where you could have the stack return an array rather than a string. I don't *think* this is currently possible, but a review of the code would be needed to assert this is the case.

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -268,11 +279,11 @@ export default function createPostGraphileHttpRequestHandler(options) {
...
   const queryCache = LRU({
     max: queryCacheMaxSize,
-    length: (n, key) => n.length + SHA1_BASE64_LENGTH,
+    length: (n: any) => n.length + SHA1_BASE64_LENGTH,
   })
 
-  let lastGqlSchema
-  const parseQuery = (gqlSchema, queryString) => {
+  let lastGqlSchema: any
+  const parseQuery = (gqlSchema: any, queryString: string) => {
```
benjie: gqlSchema is an instance of GraphQLSchema from `graphql`

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -519,10 +526,10 @@ export default function createPostGraphileHttpRequestHandler(options) {
     // The `result` will be used at the very end in our `finally` block.
     // Statements inside the `try` will assign to `result` when they get
     // a result. We also keep track of `params`.
-    let paramsList
-    let results
+    let paramsList: any
+    let results: any
```
benjie: results will be an optional array of GraphQL results (something like `{data?: any, errors?: Array<GraphQLError>}` ?)

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
benjie: This is nice! This has been annoying me for a while 👍

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -850,6 +857,13 @@ function addCORSHeaders(res) {
   )
 }
 
+function createBadAuthorizationHeaderError(): void {
+  throw httpError(
```
benjie: Don't throw here, we throw down below (and the name is just `create`)

src/postgraphile/http/setupServerSentEvents.ts
```diff
@@ -20,8 +21,8 @@ export default function setupServerSentEvents(req, res, options) {
     req._koaCtx.compress = false
   }
 
-  const sse = str => {
-    if (isKoa) {
+  const sse = (str: string) => {
+    if (isKoa && stream) {
```
benjie: Ugh. This is one of the things I hate about typed JS. The logic here would be confusing to the reader as they'd be "if it's koa but there is no stream then do res.write? Why would there not be a stream...?" and then read the code to find it's impossible to not have a stream if it's Koa.

Can you use `stream!.write(str)` below instead so that it's easier to read and doesn't change the compiled code?

tsconfig.json
benjie: Can we configure it like ESLint using these rules rather than disabling it outright? I'm not sure how flexible it is in this regard.

1. we don't care if parameters start with underscores (e.g. `_res` would be ignored)
2. we don't care if one of the parameters *after* this one is used (e.g. `(req, res, next) => next()` shouldn't raise)

tsconfig.json
```diff
@@ -14,10 +16,9 @@
     "suppressImplicitAnyIndexErrors": true,
     "strictNullChecks": true,
     "noFallthroughCasesInSwitch": true,
-    "noUnusedParameters": true,
+    "noUnusedParameters": false,
     "noUnusedLocals": true
   },
-  "include": ["src", "typings"],
```
benjie: Why have you removed this? I store a lot of extra stuff (tests for user issues, WIP code, local test scripts, etc) that I wouldn't want to be affected by tsconfig

typings/pg-sql2.d.ts
```diff
@@ -6,6 +6,6 @@ declare module 'pg-sql2' {
 
   export function compile(query: SQLQuery): QueryConfig
   export function query(strings: TemplateStringsArray, ...values: Array<SQLNode | SQLQuery>): SQLQuery
-  export function value(val: mixed): SQLNode
+  export function value(val: any): SQLNode
```
benjie: Why was this change required? Specifically this should only receive JSON-like values, I think...

src/postgraphile/http/__tests__/createPostGraphileHttpRequestHandler-test.js
benjie: (This is possibly where 855 of the removed 1K LOC came from 😳 )

src/interfaces.ts
cdaringe: a5b736f

src/interfaces.ts
cdaringe: af448cc

src/interfaces.ts
cdaringe: 20a50fe 

this one was a bit of a drag, inducing a little more typing refactor than I had wanted.  ill see if i can annotate it.  i bundled your remark from https://github.com/graphile/postgraphile/pull/794/files#r200834248 in this commit too.

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
cdaringe: ya.  it logically doesn't make sense that `showErrorStack` would affect the return type of the stack anyway.  i'm prone to change this whole logic entirely to:

`formattedError.stack = option.showErrorStack ? (error.stack || '') : ''`

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
cdaringe: ah, next isnt really used.  we may as well have literally used a no-op on 245.  237-241 provides a `next` implementation for us. doh!  thx :)

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
cdaringe: aa99779

the way the control structures flow here make the compiler (validly, IMHO) think that results can be used before assigned. for this reason i eager assigned results to `[]`.

we should scrub the whole handler generation process at some point and port it into smaller blocks/files/fns/etc

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
cdaringe: whoops! great catch. thx. fb7430b

src/postgraphile/http/setupServerSentEvents.ts
cdaringe: fe907e5

tsconfig.json
cdaringe: ah, perfect. i wasn't aware that that was a thing.  looks like ts honors it as well. 2396b0e

tsconfig.json
cdaringe: mmm.  i think i see what you mean.  flipping tsconfig to `allowJs: false` prevents `.js` files from being copied into the build, so those files would be omitted anyway if we auto gen the declarations.  if that content needs to ship, we can do it in a post processing step.  does that work?  are you certain those need to be shipped?

typings/pg-sql2.d.ts
cdaringe: i confess.  laziness as i wasn't not sure how to import a locally defined interface :/ `../src/interfaces`

src/postgraphile/http/__tests__/createPostGraphileHttpRequestHandler-test.js
cdaringe: oh brother.  🤦‍♂️ 

0ab968a

cdaringe:
alright, i need to jet for the day, but hopefully this last commit brings CI to green.  almost all talking points were addressed with the exception of:

- extra js artifacts included in the build
- the `showStack` bool vs string option

benjie:
Great! We're going to need some more tests to check these changes are safe, for example we don't have enough tests for our Koa support yet and I've spent quite a bit of time recently improving the Koa support so I'd be loathe to have a regression here.

cdaringe:
i'm a koa user, so koa support is important to me :) good news--koa tests are turn on and 💚!

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -752,25 +752,23 @@ export default function createPostGraphileHttpRequestHandler(options: ICreateReq
...
-      // Execute our request handler.
-      requestHandler(req, res, next).then(
-        // If the request was fulfilled, noop.
-        () => {
-          /* noop */
-        },
-        // If the request errored out, call `next` with the error.
-        error => next(error),
-      )
+      await koaMiddleware(req as any, res, requestHandler)
```
benjie: Can we just `return koaMiddleware(...)` and avoid the `async/await`?

benjie:
> i'm a koa user, so koa support is important to me :) good news--koa tests are turned on and 💚!

Can we make sure that it works with all the different Koa middlewares as well, particularly `koa-compress`, we don't have tests for this but I've been manually testing using https://github.com/graphile/examples (`wip` branch). If you wanted to write some automated tests for this that'd be awesome!

cdaringe:
We could add some assertions, sure.  Is that a value add test though?  Part of me says we're testing Koa by that point not postgraphile.  The postgraphile integration point is a call to app.use, a middleware just like the any others.  I'm willing to add some, but I'd be prone to keep it rather light.

sjmcdowall:
You would think, yes? But no -- the inter twinings of Postgraphile and Koa go deep.  Just 3 weeks ago (or so) this exact scenario broke Postgraphile in a Koa deploy, the use of koa-compress -- due to some twerking of the CTX stuff within Postgraphile .. Obviously fixed now -- but -- hence I think the request for regression tests?

cdaringe:
one test added in a693a77

src/postgraphile/http/__tests__/createPostGraphileHttpRequestHandler-test.js
```diff
@@ -850,5 +853,23 @@ for (const [name, createServerFromHandler] of Array.from(serverCreators)) {
...
+            onPreCreate: app => {
+              app.use(compress({
+                threshold: 0
+              }))
+            }
+          }
+        )
+        await request(server)
+          .get('/graphiql')
+          .expect('Content-Encoding', /gzip/)
```
benjie: It's specifically the GraphiQL assets that failed to serve - the JS and CSS - because they're served via `send` which has issues with Koa.

cdaringe: gotcha.  i'm not sure what your expectations are, but 565cf87 adds a download/`send` assertion _with_ the compress plugin enabled.

i also linked this build into my project will compress and a variety of other middlewares, and it's taking well.

benjie: Yeah; I'm not sure exactly how it was failing - Koa was hanging or something, so the connection never closed. Checking `sendFile` is called is probably not sufficient to test that; but I'm not sure how best to go about doing it other than spinning up a literal server and testing it end-to-end. I think there's still definitely some quirks going on with Koa (particularly around GraphiQL) but I've not managed to make them reproducible yet.

cdaringe: Cool.

Well, koa tests are now enabled, a send file test is now inserted, and I have tested this end-to-end with koa.

I think we're g2g.

cdaringe:
Merged latest master to this feature branch.  Friday's changes introduced a few conflicts.  Resolved.  Any outstanding blockers?

cdaringe:
hey @benjie, i know you & i both have day jobs, so no pressure, but it would be cool if we could talk about getting this in before the next set of _other_ features.  it'd be great to not have to chase master.  i think this is ready-to-roll

benjie:
@cdaringe Totally; I will review this before making any major changes.

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
benjie: So it turns out that this is actually a documented feature: 

> `showErrorStack`: Enables adding a stack field to the error response. Can be either the boolean `true` (which results in a single stack string) or the string `'json'` (which causes the stack to become an array with elements for each line of the stack). Recommended in development, not recommended in production.
> -- https://www.graphile.org/postgraphile/usage-library/#api-postgraphilepgconfig-schemaname-options

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
benjie: Yeah this reduce is super confusing; the `next` is actually called, it's just the inner most "no-op" middleware; each middleware wraps the previous one and we have this be the initial one.

benjie:
> achieve this theoretically via a 3 LOC code change in the tsconfig.json

😅 

This at least justifies my not having done it until now; 1.2k LOC changed 😳

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
cdaringe: in the original ternary, either way, a stack was assigned.  this is a bug with the original impl.  let's file a bug and address it independently.

cdaringe: i filed https://github.com/graphile/postgraphile/issues/804 for addressing this

cdaringe:
Cool!  Ya, sorry it blew up. `allowJs` being false really took this from an easy win to beast.

benjie:
😅 Okay; I've reviewed it fully and made a load of changes and I'm pretty much ready to merge it. I need to give it a last check over (and fix any CI errors) but I'm already running super late for my evening plans so I'm going to head out.

I've reverted a few things to their previous behaviour.

benjie:
I just went through the compiled bundle and inspected every d.ts file; the external requires that we need to bundle types for are:

```
    "@types/graphql": "^0.8.2",
    "@types/jsonwebtoken": "<7.2.1",
    "@types/koa": "2.0.44",
    "@types/pg": "^7.4.10",
```

And `postgraphile-core` which doesn't currently have types. We have a declaration in `typings/postgraphile-core.d.ts` but I don't think that's being referenced by the build... How do we solve this @cdaringe?

cdaringe:
> but I don't think that's being referenced by the build

it's being referenced by the build, but not being _included in it_.  based on your remark in #520, i think you already figured it out [:)](https://github.com/sheerun/prettier-standard)

thanks for taking to the time to get back over all of this!

benjie:
Thank you for doing the work! Being TypeScript through and through should make a big difference to a lot of users, and greater confidence in PRs.

PR #806, opened 2018-07-21
chore(lint): apply `prettier` against the codebase

benjie:
Prettier has been configured to apply minimal changes to the codebase; I'll follow this up with a PR that enables semicolons (🎉)

If you're checking this diff, you probably want to do so using `git diff -w --word-diff-regex=[^[:space:]]` so that it ignores *all* whitespace changes.


benjie:
This PR was manually reviewed to ensure the changes were deliberate.

PR #810, opened 2018-07-22
chore(lint): use more standard TSLint config

benjie:
- don't use namespace
- don't use interfaces starting with `I`
- remove a number of formatting rules now served by prettier


benjie:
@cdaringe I'm trying to move from our entirely custom tslint config to use the recommended config as much as possible. This PR is a step towards that; but it simplifies some of the code you did in #794. Are you hesitant about any of these changes (specifically, does using the namespace give any advantages? Importing directly seems a lot more convenient and simply requiring from the interfaces file seems to be a namespace anyway.

sjmcdowall:
I am sort of wondering if you want to try to use 
    "tslint-consistent-codestyle"
in the tslint.json ?  Before the prettier so it doesn't interfere.. that may even have more standard coding stuff "we" want to do (and I use it in all my projects) .. <shrug> 

As for namespace stuff -- since TS 2.6 or so that's pretty deprecated mostly and now is seen as not best practices using them unless really needed ... So I like the direction this is going ... <shrug>

benjie:
> One question -- why the "no I" interfaces? Just curious -- I use IXXXX for all my interfaces but maybe because I just followed that over from my Java days :)

Primarily because none of our other interfaces use this naming convention, and secondly because that seems to be [what TypeScript recommends in their wiki](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#names).

Using `tslint-consistent-codestyle` is going to be too much of a change right now, but it could certainly be a long term goal.

sjmcdowall:
Well to be fair -- those are the "rules" for contributing to TypeScript itself .. but fair enough .. if that's the same rules we want to adopt I'm not a big on it either way ...

benjie:
Yeah; as I say primarily it's due to consistency with our other interfaces 👍

cdaringe:
No sweat from me :)

PR #829, opened 2018-08-14
feat(jwt): include JWT claims in withPostGraphileContext callback

hyperparsa:
for issue: https://github.com/graphile/graphile-build/issues/266

withPostGraphileContext will have callback in form of:
`
{
  pgClient,
  pgRole,
  jwtClaims
}
`

### DIFF
--- a/src/postgraphile/__tests__/withPostGraphileContext-test.js
+++ b/src/postgraphile/__tests__/withPostGraphileContext-test.js
@@ -222,8 +222,8 @@ test('will include JWT claims as jwtClaims in context callback', async () => {
     },
     context => context,
   );
-  expect(jwtClaims).toEqual({ aud: 'postgraphile', a: 1, b: 2, c: 3 })
-})
+  expect(jwtClaims).toEqual({ aud: 'postgraphile', a: 1, b: 2, c: 3 });
+});
 
 test('jwtClaims should be an empty object if there is no JWT token', async () => {
   const pgClient = { query: jest.fn(), release: jest.fn() };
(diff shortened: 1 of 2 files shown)


src/postgraphile/__tests__/withPostGraphileContext-test.js
```diff
@@ -209,6 +209,34 @@ test('will add extra claims as available', async () => {
...
+  const pgPool = { connect: jest.fn(() => pgClient) };
+  const { jwtClaims } = await withPostGraphileContext(
+    {
+      pgPool,
+    },
+    context => context,
+  );
+  expect(jwtClaims).toEqual({});
+});
+
```
benjie: If there’s no JWT then jwtClaims should be undefined (or null). It defaults to empty object only to make the logic of that function simpler; you’ll need to add a check e.g. `jwtClaims: hasJwt?jwtClaims:null` Into the return value of getSettingsForPgClientTransaction

hyperparsa: Alrighty!

benjie:
👌  Perfect, thanks for doing this! I'll probably add pgLocalSettings in future also; but that can wait for now 👍

hyperparsa:
😀

sjmcdowall:
As someone who is working with this stuff (JWT, etc) -- what does this PR mean in terms of changes I can expect or need to know about? :)  I don't think much since we are doing all the JWT stuff in Middleware before -- but just need to pass the claims in and I think that doesn't change .. ??

benjie:
It just makes `jwtClaims` available on the GraphQL `context` (third argument to resolve); mostly useful to people using `graphile-utils` or similar to add plugins to the schema. No code changes necessary (though if you were previously using `additionalGraphQLContextFromRequest` to provide this functionality you can perhaps simplify/remove that code).

PR #831, opened 2018-08-15
chore(tests): enforce current behaviour when settings are provided by both JWT and pgSettings

benjie:
The rules are this:

If it's the `role` setting then the first of these will win:

1. Role from pgSettings
2. Role from JWT
3. Role from `pgDefaultRole`

Otherwise, the first of these will win:

1. JWT
2. pgSettings

I'll probably change this in v5.

### DIFF
--- a/src/postgraphile/__tests__/withPostGraphileContext-test.js
+++ b/src/postgraphile/__tests__/withPostGraphileContext-test.js
@@ -573,6 +573,60 @@ test('will set a role provided in the JWT', async () => {
   ]);
 });
 
+test('if same settings are set by pgSettings and JWT, JWT will "win", except for the "role" because of legacy', async () => {
+  const pgClient = { query: jest.fn(), release: jest.fn() };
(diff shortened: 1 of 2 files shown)


sjmcdowall:
I agree with the comment that in V5 the rules should be consistent -- IMHO 

pgSettings
JWT 

in that order .. BUT -- do a logical union of the two .. so if JWT has claims that aren't filled in by pgSettings they are also included .. I am sure there is some fancy set theory symbolism I could use to indicate what I mean but I think you grok it :)

benjie:
Yeah; you mean `{...jwtSettings, ...pgSettings}` 😛 

I agree that `pgSettings` should win. I'm planning to pass the JWT to the pgSettings function in v5 so it can vet the previous settings and add its own if it wants. That's a significant change though. The reason JWT previously won was pgSettings used to be a constant rather than a function call, dynamic pgSettings is more recent.

PR #853, opened 2018-09-07
Document skipPlugins and expose CLI option

### LINKED ISSUE #850: Document skipPlugins
postgraphile-core accepts skipPlugins; we should expose this via PostGraphileOptions.

mattbretl:
Fixes #850


benjie:
Nice! Run `yarn prettier:fix` to fix the lint errors then it should be good to merge 👍

PR #910, opened 2018-11-16
Provide jwtClaims in resolver context when set via pgSettings

RomanGotsiy:
If JWT token is parsed/validated via `pgSettings`, jwtClaims are not passed to the custom resolver context.

```js
makeExtendSchemaPlugin(
  build => ({
    typeDefs: gql`
      extend type Thing {
        isOwner: Boolean! @requires(columns: ["owner"])
      }
    `,
    resolvers: {
      Thing: {
        isOwner: async (row, _args, context) => {
          return row.owner === context.jwtClaims.user_id; // <-- // context.jwtClaims is null here
        }
      }
    }
  })
)
```

This PR fixes it.

### DIFF
--- a/src/postgraphile/withPostGraphileContext.ts
+++ b/src/postgraphile/withPostGraphileContext.ts
@@ -309,6 +309,9 @@ async function getSettingsForPgClientTransaction({
       if (pgSettings.hasOwnProperty(key) && isPgSettingValid(pgSettings[key])) {
         if (key === 'role') {
           role = String(pgSettings[key]);
+        } else if (key.startsWith('jwt.claims.')) {
+          // strip 'jwt.claims.' part (11 symbols)
+          jwtClaims[key.substring(11)] = String(pgSettings[key]);
         } else {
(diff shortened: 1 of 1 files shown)


benjie:
Hi @RomanGotsiy, thanks for your pull request! I'm afraid that this is not the direction I'd like to take this feature. The `jwtClaims` are only intended to be added to context by default when their parsing is done internally by PostGraphile (because without it, there's no way of putting them on context). If, however, you do your own JWT parsing via `pgSettings` and you want the claims available on context, you should do so via the `additionalGraphQLContextFromRequest` option. This allows you to decide whether or not you want `jwtClaims` to be present on context. I hope this makes sense, feel free to discuss this further.

RomanGotsiy:
Ah. Missed that function.
This way I have to parse JWT token twice.

But this is not an issue for my use case as I am using this primarily for testing.

Thanks for the quick reply!

benjie:
You can cache it to `req` in `pgSettings` and reference it in `additionalGraphQLContextFromRequest` to save computing it twice.

PR #924, opened 2018-11-23
feat(http): add tests for fastify server

### LINKED ISSUE #869: Does PostGraphile support Fastify?
https://www.fastify.io/docs/latest/Middlewares/

Would be good to add it to the list of servers that we can be mounted as a middleware.

benjie:
Fastify passes our core test suite 💯 `fastify.use(postgraphile(...))`

It doesn't seem to work with mounting a middleware on a subpath right now (i.e. `fastify.use('/path/to', postgraphile(...))`), presumably it's not doing `originalUrl` / `url` manipulation?

Fixes #869

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "connect": "^3.5.0",
     "css-loader": "1.0.0",
     "express": "^4.14.0",
+    "fastify": "^1.13.0",
     "html-webpack-inline-source-plugin": "^0.0.10",
     "html-webpack-plugin": "^3.2.0",
     "jest": "^18.1.0",
--- a/src/postgraphile/http/__tests__/createPostGraphileHttpRequestHandler-test.js
+++ b/src/postgraphile/http/__tests__/createPostGraphileHttpRequestHandler-test.js
@@ -109,10 +109,10 @@ const serverCreators = new Map([
     'fastify',
     async (handler, _options, subpath) => {
       let server;
-      function serverFactory(handler, opts) {
+      function serverFactory(fastlyHandler, opts) {
         if (server) throw new Error('Factory called twice');
         server = http.createServer((req, res) => {
-          handler(req, res);
+          fastlyHandler(req, res);
         });
         return server;
       }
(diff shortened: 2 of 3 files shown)


ken0x0a:
I think
```
fastify.use('/path/to', postgraphile(..., ..., { graphqlRoute: '/path/to', ... }))
```

is also works fine, as followings works fine

```
function postgraphileService(fastify ) {
  app.use(postgraphile(..., ..., { graphqlRoute: '/path/to', ... }))
}
fastify.register(postgraphileService, { prefix: '/path/to' })
```

benjie:
Maybe they’ve added originalUrl support?

If you’re relying on this I strongly suggest you open a PR upgrading fastify and enabling the tests I disabled; without them there’s a higher chance of regression.

ken0x0a:
@benjie Thank you for a reply! I know you are very busy right now.

Actually, I'd rather trying to do **[even-lower-level-access](https://www.graphile.org/postgraphile/usage-schema/#even-lower-level-access)**, to get object

which `operationName` as key, and
`sql` as value,

to make API even faster,
without `parsing query`, `validate args`, `JSON.stringify` & `JSON.parse`.
As I'm using persist query, those are too much work for every request.

Postgraphile is awesome!
I feel, Postgraphile become even faster after upgraded to 4.4.1.
Thank you for your tremendous works & support!

I'll try PR if I couldn't do sth mentioned above or not satisfied with that.

ken0x0a:
@benjie 
Would you give me hint to solve following error? Thank you!

```
   Cannot find module '../../assets/favicon.ico' from 'createPostGraphileHttpRequestHandler.ts'

    However, Jest was able to find:
    	'http/createPostGraphileHttpRequestHandler.ts'

    You might want to include a file extension in your import, or update your 'moduleFileExtensions', which is currently ['ts', 'js', 'ico', 'html', 'json'].

    See https://jestjs.io/docs/en/configuration#modulefileextensions-array-string

    However, Jest was able to find:
    	'./postgraphile.ts'

    You might want to include a file extension in your import, or update your 'moduleFileExtensions', which is currently ['ts', 'js', 'ico', 'html', 'json'].

    See https://jestjs.io/docs/en/configuration#modulefileextensions-array-string

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:229:17)
      at Object.<anonymous> (src/postgraphile/http/file:/Users/k/GitHub/postgraphile/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts:44:1)
```

PR #958, opened 2019-01-11
fix(server): throw when encountering `undefined` positional arguments

### LINKED ISSUE #951: Warn user when they pass `undefined` schema in library mode
I'm submitting a ... 

* [x] bug report
* [ ] feature request
* [ ] question 

**PostGraphile version:** `v4.1.0`

**Steps to reproduce:**

```ts
import postgraphile from 'postgraphile';
import { createServer } from 'http';

const {
  PORT,
  POSTGRES_DB,
  POSTGRES_PASSWORD,
  POSTGRES_USER,
  POSTGRES_HOST,
  POSTGRES_SCHEMA,
} = process.env;

const POSTGRES_CONNECTION_STRING = `postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}/${POSTGRES_DB}`;

const app = createServer(
  postgraphile(POSTGRES_CONNECTION_STRING, POSTGRES_SCHEMA, {
    graphiql: true,
    enhanceGraphiql: true,
  }),
);

const port = PORT || 5000;

app.listen(port, (err?: Error) => {
  if (err != null) console.error(err);
  console.log(`Listening on http://localhost:${port}`);
});
```

**Current behavior:**

When using as a library (as demonstrated above), if the second argument to `postgraphile` is `undefined`, the server _appears_ to start successfully and responds to requests at `/graphql`, but does not serve the UI at `/graphiql`.

In this case, it took me a while to realize I was not setting the `POSTGRES_SCHEMA` environment variable and that it was causing the problem.

**Expected behavior:**

Invalid configuration should immediately throw and prevent the server from starting all together.

marshall007:
Fixes #951 

**Note:** I had to manually run `yarn add graphql postgraphile-core` in order to get the tests running on my local machine. For whatever reason these dependencies were not installed with just `yarn`.

### DIFF
--- a/src/postgraphile/postgraphile.ts
+++ b/src/postgraphile/postgraphile.ts
@@ -121,20 +121,20 @@ export default function postgraphile(
     schema = schemaOrOptions;
     incomingOptions = maybeOptions || {};
   }
-  // If the second argument is the incomingOptions so set `schema` to the
-  // default and `incomingOptions` to the second argument.
+  // If the second argument is null or an object then use default `schema`
+  // and set incomingOptions to second or third argument (or default).
   else if (typeof schemaOrOptions === 'object') {
     schema = 'public';
     incomingOptions = schemaOrOptions || maybeOptions || {};
   }
-  // Otherwise the second argument is undefined, use defaults for both `schema` and
-  // `incomingOptions`.
+  // Otherwise if the second argument is present it's invalid: throw an error.
+  else if (arguments.length > 1) {
+    throw new Error(
+      'The second argument to postgraphile was invalid... did you mean to set a schema?',
+    );
+  }
+  // No schema or options specified, use defaults.
   else {
-    if (arguments.length > 1) {
-      throw new Error(
-        'The second argument to postgraphile was `undefined`... did you mean to set a schema?',
-      );
-    }
     schema = 'public';
     incomingOptions = {};
   }
--- a/src/postgraphile/postgraphile.ts
+++ b/src/postgraphile/postgraphile.ts
@@ -114,41 +114,36 @@ export default function postgraphile(
   // must process them with `pluginHook` before we can rely on them.
   let incomingOptions: PostGraphileOptions;
 
-  // If the second argument is undefined, use defaults for both `schema` and
-  // `incomingOptions`.
-  if (typeof schemaOrOptions === 'undefined') {
-    schema = 'public';
-    incomingOptions = {};
-  }
   // If the second argument is a string or array, it is the schemas so set the
   // `schema` value and try to use the third argument (or a default) for
   // `incomingOptions`.
-  else if (typeof schemaOrOptions === 'string' || Array.isArray(schemaOrOptions)) {
+  if (typeof schemaOrOptions === 'string' || Array.isArray(schemaOrOptions)) {
     schema = schemaOrOptions;
     incomingOptions = maybeOptions || {};
   }
-  // Otherwise the second argument is the incomingOptions so set `schema` to the
+  // If the second argument is the incomingOptions so set `schema` to the
   // default and `incomingOptions` to the second argument.
-  else {
+  else if (typeof schemaOrOptions !== 'undefined') {
     schema = 'public';
     incomingOptions = schemaOrOptions;
+  }
+  // Otherwise the second argument is undefined, use defaults for both `schema` and
+  // `incomingOptions`.
+  else {
+    if (arguments.length > 1) {
+      throw new Error('The second argument to postgraphile was `undefined`... did you mean to set a schema?');
+    }
+    schema = 'public';
+    incomingOptions = {};
+  }
+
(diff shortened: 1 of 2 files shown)


src/postgraphile/postgraphile.ts
```diff
@@ -114,41 +114,36 @@ export default function postgraphile(
...
+  // `incomingOptions`.
+  else {
+    if (arguments.length > 1) {
+      throw new Error('The second argument to postgraphile was `undefined`... did you mean to set a schema?');
+    }
+    schema = 'public';
+    incomingOptions = {};
+  }
+
+  if (typeof poolOrConfig === 'undefined' && arguments.length > 1) {
```
benjie:
```suggestion
  if (typeof poolOrConfig === 'undefined' && arguments.length >= 1) {
```

Should this be `>=` - i.e. if they passed undefined?

benjie:
I'm going to classify this as a bugfix rather than a breaking change because it may help users find previously undetected errors in their code. Should they want to skip an arg they should instead pass `null`.

benjie:
Notes to self:

- [ ] Check you can pass `null` for connection string and have it extracted from env
- [ ] Check that `null` for schema behaves reasonably
- [ ] Check that `null` for options behaves reasonably

src/postgraphile/postgraphile.ts
```diff
@@ -114,41 +114,40 @@ export default function postgraphile(
...
-    schema = 'public';
-    incomingOptions = {};
-  }
   // If the second argument is a string or array, it is the schemas so set the
   // `schema` value and try to use the third argument (or a default) for
   // `incomingOptions`.
-  else if (typeof schemaOrOptions === 'string' || Array.isArray(schemaOrOptions)) {
+  if (typeof schemaOrOptions === 'string' || Array.isArray(schemaOrOptions)) {
     schema = schemaOrOptions;
     incomingOptions = maybeOptions || {};
```
benjie: If we call `postgraphile(pool, null, {dynamicJson: true})` then I think the options object will be dropped - we don't want this. I think if we change the above to:

```js
if (!schemaOrOptions || typeof schemaOrOptions === 'string' || Array.isArray(schemaOrOptions)) {
  schema = schemaOrOptions || 'public';
  incomingOptions = maybeOptions || {};
```

then this situation will be handled? What do you think?

src/postgraphile/postgraphile.ts
```diff
@@ -114,41 +114,40 @@ export default function postgraphile(
...
   }
-  // Otherwise the second argument is the incomingOptions so set `schema` to the
+  // If the second argument is the incomingOptions so set `schema` to the
   // default and `incomingOptions` to the second argument.
+  else if (typeof schemaOrOptions !== 'undefined') {
+    schema = 'public';
+    incomingOptions = schemaOrOptions || {};
```
benjie:
```suggestion
    if (typeof schemaOrOptions !== 'object') {
      throw new Error("Expected schema string/array or options object, instead received " + typeof schemaOrOptions);
    }
    incomingOptions = schemaOrOptions || {};
```

This'll handle things like:

`postgraphile(pool, false)` / `postgraphile(pool, true)` / `postgraphile(pool, 0)` / `postgraphile(pool, true, {...})` / `postgraphile(pool, 0, {...})` I think.

benjie:
Maybe it'd be cleaner to restructure the code using `arguments.length` directly:

```js
function postgraphile(a, b, c) {
  let pool = a;
  let schema;
  let options;
  if (arguments.length >= 1 && typeof a === 'undefined') {
    throw new Error("Pool argument passed to postgraphile was undefined; to use the defaults please pass `null` (or omit all arguments).");
  }
  if (arguments.length === 3) {
    if (!b) {
      throw new Error("Options specified, but schema omitted, pass your schema name or 'public'");
    }
    if (!c) {
      throw new Error("Third argument to postgraphile was falsy - please either omit it or pass an options object");
    }
    [schema, options] = [b, c];
  } else if (arguments.length === 2) {
    if (!b) {
      throw new Error("Second argument to postgraphile was falsy");
    }
    if (typeof b === 'string' || Array.isArray(b)) {
      schema = b
    } else if (typeof b === 'object') {
      options = b;
    } else {
      throw new Error("Invalid options passed to PostGraphile");
    }
  }
  // ...
}
```

I think I'll remove this flexibility in v5 and just go with 3 standard arguments like in the docs.

marshall007:
@benjie I think I may have come up with a simpler stop-gap solution, but let me know if that doesn't work for you. As for breaking changes in v5, I'd be in favor of simplifying further to two arguments: 

```ts
function postgraphile(poolConfig?: Pool | PoolConfig, options?: PostGraphileOptions) {
  // ...
}
``` 

... with `options.schema` as a nested property instead.

benjie:
That makes a lot of sense 👍  for schema as an option.

src/postgraphile/postgraphile.ts
benjie: Handled 👍

PR #970, opened 2019-01-20
feat(hooks): add hooks for HTTP results

### LINKED ISSUE #962: Add a hook for custom logging of successful requests and errors
I'm submitting a ...

* [ ] bug report
* [x] feature request
* [ ] question

PostGraphile version (library): 4.2

First of all, thanks for an amazing project! We're using postgraphile in production and loving it.

One issue we have is that we want to log whether a request was successfully processed or if it failed. For this purpose, we'd love to have hooks/callback functions that we can supply to Postgraphile and that'll then be called when sending a response.

For logging/debugging purposes, it would be great to have the status code and message being sent (incl. data). The simplest thing I can think of is to call the callback function with the response object in [line 786 of createPostGraphileHttpRequestHandler.ts](https://github.com/graphile/postgraphile/blob/master/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts#L786) and the response value (`returnArray ? results : results[0]!`). This should hopefully require minimal maintenance as the response object is unlikely to change.

Some users might want the additional info in the built-in query log [here](https://github.com/graphile/postgraphile/blob/master/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts#L732-L737), but for our purposes at least, that's not needed. It might be too expensive to prettify the query etc. so we'd prefer not to have the callback/hook get this information.

Would this implementation make sense? If so, should I just add an additional key to the options object with a callback function?

### LINKED ISSUE #969: Add a hook for augmenting GraphiQL HTML (e.g. dark mode)

benjie:
Fixes #962
Fixes #969

### DIFF
--- a/src/postgraphile/cli.ts
+++ b/src/postgraphile/cli.ts
@@ -24,6 +24,11 @@ import { mixed } from '../interfaces';
 import * as manifest from '../../package.json';
 import sponsors = require('../../sponsors.json');
 
+// tslint:disable-next-line no-any
+function isString(str: any): str is string {
+  return typeof str === 'string';
+}
+
 const sponsor = sponsors[Math.floor(sponsors.length * Math.random())];
 
 const debugCli = debugFactory('postgraphile:cli');
@@ -703,7 +708,7 @@ if (noServer) {
               pgPort !== 5432 ? `:${pgConfig.port || 5432}` : ''
             }${pgDatabase ? `/${pgDatabase}` : ''}`;
 
-        const information = pluginHook(
+        const information: Array<string> = pluginHook(
           'cli:greeting',
           [
             `GraphQL API:         ${chalk.underline.bold.blue(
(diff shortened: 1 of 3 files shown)


benjie:
Example of a GraphiQL dark mode hook:

```js
const fs = require('fs');

module.exports = {
  ['postgraphile:graphiql:html'](html) {
    const darkModeHtml = fs.readFileSync(`${__dirname}/darkmode.html`, 'utf8');
    // Have to replace it in the body (rather than head) so the JS runs first.
    return html.replace('</body>', darkModeHtml + '</body>');
  },
};
```

where the `darkmode.html` file contains the contents of https://gist.githubusercontent.com/bkeating/cfe0d5e72bd3f9f1f77e1a2ff2309972/raw/31f8d436878f98e56f363fd52bb173f336a69b8e/graphiql-dark-mode.css

PR #986, opened 2019-02-05
feat(server): websockets support

benjie:
Enables people to experiment with their own subscription implementations without needing to use an alternative server.

graphiql-subscriptions

- [x] Don't stop subscription for watch-mode / introspection
- [x] Don't allow non-subscriptions over websockets (by default)
- [x] Handle errors from subscriptions
- [x] Check headers work over queries and subscriptions

### DIFF
--- a/src/postgraphile/http/subscriptions.ts
+++ b/src/postgraphile/http/subscriptions.ts
@@ -3,7 +3,6 @@ import { HttpRequestHandler, mixed } from '../../interfaces';
 import { subscribe, ExecutionResult } from 'graphql';
 import { RequestHandler, Request, Response } from 'express';
 import * as WebSocket from 'ws';
-import { parse } from 'url';
 import { SubscriptionServer, ConnectionContext } from 'subscriptions-transport-ws';
 import parseUrl = require('parseurl');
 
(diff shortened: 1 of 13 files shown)


src/postgraphile/http/subscriptions.ts
```diff
@@ -0,0 +1,210 @@
...
+          }),
+        )
+        .then(null, reject);
+    });
+  };
+
+  const wss = new WebSocket.Server({ noServer: true });
+
+  websocketServer.on('upgrade', (req, socket, head) => {
+    const url = req.originalUrl || req.url;
```
benjie: ⚠️  This logic doesn't match our HTTP handler logic

src/postgraphile/http/subscriptions.ts
```diff
@@ -0,0 +1,210 @@
...
+    withPostGraphileContextFromReqRes,
+    handleErrors,
+  } = postgraphileMiddleware;
+  const graphqlRoute = options.graphqlRoute || '/graphql';
+
+  const schema = await getGraphQLSchema();
+
+  const contextKeepalivePromisesByOpId = {};
+
+  const contextKey = (ws: WebSocket, opId: string) => ws['postgraphileId'] + '|' + opId;
```
benjie: ⚠️ Forgot to add `postgraphileId`

PR #1000, opened 2019-02-23
chore(comments): fix queryDocumentAst comment

benjie:


src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -681,6 +681,8 @@ export default function createPostGraphileHttpRequestHandler(
             // send the errors to the client with a `400` code.
             if (validationErrors.length > 0) {
               result = { errors: validationErrors, statusCode: 400 };
+            } else if (!queryDocumentAst) {
+              throw new Error('Could not process query');
```
benjie: This should never fire?

PR #1007, opened 2019-02-27
chore: refactor a bunch of stuff for subscriptions

benjie:


src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -274,7 +274,6 @@ export default function createPostGraphileHttpRequestHandler(
 
   const staticValidationRules = pluginHook('postgraphile:validationRules:static', specifiedRules, {
     options,
-    httpError,
```
benjie: This was never used, and it's not compatible with subscriptions.

PR #1015, opened 2019-03-07
fix(postgraphile): add buffer support as jwt secret

dijam:
Currently it is only accepted to provide type `string` for JWT secret when JWT can accept both `buffer` and `string`. With this PR, we allow type buffer to be passed as jwtSecret.

### DIFF
--- a/src/postgraphile/withPostGraphileContext.ts
+++ b/src/postgraphile/withPostGraphileContext.ts
@@ -283,9 +283,15 @@ async function getSettingsForPgClientTransaction({
     // Try to run `jwt.verify`. If it fails, capture the error and re-throw it
     // as a 403 error because the token is not trustworthy.
     try {
-      // If a JWT token was defined, but a secret was not provided to the server
-      // throw a 403 error.
-      if (typeof jwtSecret !== 'string') throw new Error('Not allowed to provide a JWT token.');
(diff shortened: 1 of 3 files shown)


src/postgraphile/withPostGraphileContext.ts
```diff
@@ -283,9 +283,10 @@ async function getSettingsForPgClientTransaction({
     // Try to run `jwt.verify`. If it fails, capture the error and re-throw it
     // as a 403 error because the token is not trustworthy.
     try {
-      // If a JWT token was defined, but a secret was not provided to the server
-      // throw a 403 error.
-      if (typeof jwtSecret !== 'string') throw new Error('Not allowed to provide a JWT token.');
+      // If a JWT token was defined, but a secret was not provided to the server or
+      // secret had unsupported type, throw a 403 error.
+      if (!Buffer.isBuffer(jwtSecret) && typeof jwtSecret !== 'string')
+        throw new Error('JWT secret must be a "string" or "buffer".');
```
benjie: Since this is a user visible error message, lets change this back to what it was, but add a server-side console message.

```suggestion
{
        // tslint:disable-next-line no-console
        console.error("ERROR: `jwtToken` was provided, but `jwtSecret` was not set to a string or buffer - rejecting request.")
        throw new Error('Not allowed to provide a JWT token.');
      }
```

PR #1026, opened 2019-03-27
Improvements to Forum Example Authentication and Windows Support for Build Schema Script

calben:
This is a pretty small set of changes.

1. Simplified the `authenticate` function in the schema so that it is simple sql code that requires less explanation.
1. Added a `change_password` functionality.
1. Added Windows support to the build_schema (just needed to add `trim` thanks to windows newline nonsense)

### DIFF
--- a/examples/forum/TUTORIAL.md
+++ b/examples/forum/TUTORIAL.md
@@ -28,6 +28,7 @@ In this tutorial we will walk through the Postgres schema design for a forum app
   - [JSON Web Tokens](#json-web-tokens)
   - [Logging In](#logging-in)
   - [Using the Authorized User](#using-the-authorized-user)
+  - [Updating Passwords](#updating-passwords)
   - [Grants](#grants)
   - [Row Level Security](#row-level-security)
 - [Conclusion](#conclusion)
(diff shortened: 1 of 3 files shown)


examples/forum/TUTORIAL.md
```diff
@@ -694,6 +680,33 @@ comment on function forum_example.current_person() is 'Gets the person who was i
...
+  if exists (select 1 from forum_example_private.person_account where person_account.person_id = current_person.id and person_account.password_hash = crypt($1, person_account.password_hash)) 
+  then
+    update forum_example_private.person_account set password_hash = crypt($2, gen_salt('bf')) where person_account.person_id = current_person.id; 
+    return true;
+  else 
+    return false;
+  end if;
+end;
+$$ language plpgsql strict security definer;
+```
```
benjie: This could also be accomplished with SQL; something like:

```sql
create function forum_example.change_password(current_password text, new_password text) 
returns boolean as $$
  with updated as (
    update forum_example_private.person_account
      set password_hash = crypt($2, gen_salt('bf'))
      where person_account.person_id = (forum_example.current_person()).id
      and person_account.password_hash = crypt($1, person_account.password_hash)
      returning *;
  )
  select exists (select 1 from updated);
end;
$$ language plpgsql strict security definer;
```

(untested)

PR #1081, opened 2019-05-21
New feature graphiql-authorization-event-origin

deinspanjer:
allows a window to update graphiql with an authorization header value to be used on every request

See postgraphiql/public/example_authorization_wrapper/index.html for a quick demo

Happy to turn it into a proper PR if there is interest.

### DIFF
--- a/postgraphiql/src/components/PostGraphiQL.js
+++ b/postgraphiql/src/components/PostGraphiQL.js
@@ -163,9 +163,11 @@ class PostGraphiQL extends React.PureComponent {
      * a message event listener to allow windows with that origin to update our Authorization header.
      */
     if (POSTGRAPHILE_CONFIG.graphiqlAuthorizationEventOrigin) {
-      console.log('Setting up listener for messages from ', POSTGRAPHILE_CONFIG.graphiqlAuthorizationEventOrigin);
+      // We want to support a space separated list of origins but avoid a false positive if someone is tricky.
+      const paddedOriginsList = ` ${POSTGRAPHILE_CONFIG.graphiqlAuthorizationEventOrigin} `;
+      console.log('Setting up listener for messages from', paddedOriginsList);
       window.addEventListener("message", (event) => {
-          if (event.origin !== POSTGRAPHILE_CONFIG.graphiqlAuthorizationEventOrigin) {
+          if (!paddedOriginsList.includes(event.origin)) {
             console.error('window.postMessage received from an unauthorized origin: ', event.origin);
             return;
           }
@@ -174,7 +176,7 @@ class PostGraphiQL extends React.PureComponent {
             const val = JSON.stringify({ Authorization: event.data.Authorization });
             // destructuring to be paranoid and grab only the Authorization header value
             this.updateHeadersState(val);
-            console.log('Authorization header updated by ', POSTGRAPHILE_CONFIG.graphiqlAuthorizationEventOrigin);
+            console.log('Authorization header updated by ', event.origin);
           }
         }, false);
     } else {
--- a/postgraphiql/src/components/PostGraphiQL.js
+++ b/postgraphiql/src/components/PostGraphiQL.js
@@ -19,6 +19,7 @@ const {
     streamUrl: 'http://localhost:5000/graphql/stream',
     enhanceGraphiql: true,
     subscriptions: true,
+    graphiqlAuthorizationEventOrigin: null
   },
 } = window;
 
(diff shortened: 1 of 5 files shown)


postgraphiql/src/components/PostGraphiQL.js
```diff
@@ -94,14 +95,19 @@ class PostGraphiQL extends React.PureComponent {
     // stuffs.
     schema: null,
     showHeaderEditor: false,
+
+    // deinspanjer: I think we'd like to have a way to wait for an initial value
+    // if we are using graphiqlAuthorizationEventOrigin but I'm not sure how to do that.
     headersText: '{\n"Authorization": null\n}\n',
```
benjie: I think reading it from the query hash would be a good solution (and would solve other people's problems without requiring `graphiqlAuthorizationEventOrigin` also). e.g. `http://localhost:5000/graphiql#token=skldjfojweoij0goerng0iergn`

```suggestion
    headersText: headersFromLocationHash ? JSON.stringify(headersFromLocationHash) : '{\n"Authorization": null\n}\n',
```

Where

```js
const qs = require('qs');
const hashObject = qs.parse(window.location.hash || "");
const token = hashObject.token;
const headersFromLocationHash = token ? {Authorization: `Bearer ${token}`} : null;
```

postgraphiql/src/components/PostGraphiQL.js
```diff
@@ -152,6 +158,32 @@ class PostGraphiQL extends React.PureComponent {
...
+    /**
+     * If we were configured with a graphiqlAuthorizationEventOrigin then we will set up
+     * a message event listener to allow windows with that origin to update our Authorization header.
+     */
+    if (POSTGRAPHILE_CONFIG.graphiqlAuthorizationEventOrigin) {
+      // We want to support a space separated list of origins but avoid a false positive if someone is tricky.
+      const paddedOriginsList = ` ${POSTGRAPHILE_CONFIG.graphiqlAuthorizationEventOrigin} `;
+      console.log('Setting up listener for messages from', paddedOriginsList);
+      window.addEventListener("message", (event) => {
+          if (!paddedOriginsList.includes(event.origin)) {
```
benjie: This is insecure, I think you mean:

```suggestion
          if (!paddedOriginsList.includes(` ${event.origin} `) || !event.origin) {
```

Though a better approach may be to replace `paddedOriginsList` with

```js
const originsWhitelist = POSTGRAPHILE_CONFIG.graphiqlAuthorizationEventOrigin.split(/\s+/);
```

postgraphiql/src/components/PostGraphiQL.js
```diff
@@ -152,6 +158,32 @@ class PostGraphiQL extends React.PureComponent {
...
+            console.error('window.postMessage received from an unauthorized origin: ', event.origin);
+            return;
+          }
+
+          if (event.data && event.data.Authorization) {
+            const val = JSON.stringify({ Authorization: event.data.Authorization });
```
benjie: Since headers are case insensitive, they're typically handed around in lower case (as in `req.headers` in Node.js). I recommend sticking to this lower-case convention, or using a case-insensitive accessor.

postgraphiql/src/components/PostGraphiQL.js
```diff
@@ -209,6 +241,21 @@ class PostGraphiQL extends React.PureComponent {
...
+         headersText: newHeaderText,
+         headersTextValid: isValidJSON(newHeaderText),
+       },
+       () => {
+         if (this.state.headersTextValid && this.subscriptionsClient) {
+           // Reconnect to websocket with new headers
+           this.subscriptionsClient.close(false, true);
+         }
+       }
+     )
```
benjie: This'd be a lot neater if we refactored everything for React hooks. (Do not action this comment!)

postgraphiql/src/components/PostGraphiQL.js
deinspanjer: Great, in my private build I have this hash parsing.  I want to change it to be `Bearer=` and I can submit a separate PR for that.

postgraphiql/src/components/PostGraphiQL.js
deinspanjer: Hmm.. Okay, I was going along with the casing that was used in the stringified `headersText` value.  Should that also be changed to lowercase?

deinspanjer:
I would definitely like to help enable a more flexible way to interact via plugins.  The primary reason I didn't go that route was because I was going for minimal touch and I didn't have any experience with your plugin system (or React in general honestly ;).

If I could get a little guidance from you or others in the postgraphile community, I'd be happy to work on refactoring this feature into such a generic callback hook

postgraphiql/src/components/PostGraphiQL.js
deinspanjer: Ah great catch, yep I forgot to pad the testing origin een though I specifically padded the list so it would match when padded!

I was cheating a little here using .includes since I remembered how it worked.  If we have an actual array with split, what is the modern effective way to scan the list for a match?  I could do it in a forEach or something but it feels like there is probably a more modern JS way.

benjie: You can use `Array.prototype.includes` rather than `String.prototype.includes` 😉 It's available in all current versions of Node (and PostGraphile only supports 8.6+ anyway). Ref: https://node.green/#ES2016-features-Array-prototype-includes

postgraphiql/src/components/PostGraphiQL.js
benjie: That's effectively user text, which should be treated case-insensitively. A user could enter `"AuThOrIzAtIoN": "Bearer ..."` and it should still work. The keys on the `event.data` payload should be lower case/camelCase; e.g.

```suggestion
          if (event.data && event.data.authorization) {
            const val = JSON.stringify({ Authorization: event.data.authorization });
```

But should we also merge in the other previous headers or overwrite them?

PR #1089, opened 2019-05-28
feat(jwt): add asymmetric JWT signing and verifying support

speller:
- Added ability to pass a public key to the jsonwebtoken.verify function.
- Added support for passing options to the jsonwebtoken.sign function (rest of the functionality will be added to the `graphile-engine` package slightly later).

Suggested changes were not tested enough for the backward compatibility, no tests were added for the new functionality. My current priority is to make asymmetric encryption working and that's all. Please add all the required tests by yourself. Also, please fix my English grammar if it's not perfect.

Private and public keys are required to be in the PEM format. I'm using the following script to run CLI PostGrahile version under Ubuntu:

```
export PG_JWT_PRIVATE_KEY="`cat mykey.pem`"
export PG_JWT_PUBLIC_KEY="`cat mykey-pub.pem`"

node cli.js -c postgres://*************?ssl=true \
	-s test_public -j -t test_public.jwt_token --enhance-graphiql \
	--jwt-secret "${PG_JWT_PRIVATE_KEY}" \
	--jwt-public-key "${PG_JWT_PUBLIC_KEY}" \
	--default-role test_anonymous \
	--jwt-sign-algorithm RS512 --jwt-sign-expires-in "30m" \
	--jwt-verify-algorithms RS512
```

I'm also using the following commands to generate correct keys:

```
openssl genrsa -out mykey.pem 2048
openssl rsa -in mykey.pem -pubout -out mykey-pub.pem
```


betaboon:
@speller i guess you misstyped in title and meant "asymmetric"
i am looking forward to this PR :)

speller:
> @speller i guess you misstyped in title and meant "asymmetric"
> i am looking forward to this PR :)

You're right, I've mistyped it :) Fixed. The correct term is used in the commit text, fortunately )

src/postgraphile/cli.ts
```diff
@@ -276,9 +276,54 @@ pluginHook('cli:flags:add:webserver', addFlag);
...
+    '--jwt-sign-jwt-id <string>',
+    'jsonwebtoken jwtid option',
+  )
+  .option(
+    '--jwt-sign-no-timestamp <string>',
+    'jsonwebtoken noTimestamp option',
+  )
+  .option(
+    '--jwt-sign-encoding <string>',
+    'jsonwebtoken encoding option',
```
benjie: Whilst I appreciate the hard work you've put into this, once added we will have 24 JWT CLI flags, and 42 non-JWT CLI flags - more than a third of our CLI flags will be for JWTs. I'm not super comfortable with this; we already have too many JWT flags IMO (and too many CLI flags in general).

(My main concern here is that I think the more flags we add, the more likely people are to skip over the docs, and thus the more time I will be spending answering questions rather than advancing the project.)

I think for CLI users it makes more sense to supply these options via a `.postgraphilerc.js` file; what do you think? If so, this would be more of a documentation issue than a code issue. Library users will already be supplying these via `jwtSignOptions`, I think?

src/postgraphile/cli.ts
```diff
@@ -592,7 +661,9 @@ const postgraphileOptions = pluginHook(
     enhanceGraphiql: enhanceGraphiql ? true : undefined,
     jwtPgTypeIdentifier: jwtPgTypeIdentifier || deprecatedJwtPgTypeIdentifier,
     jwtSecret: jwtSecret || deprecatedJwtSecret,
+    jwtPublicKey,
```
benjie: Do we need a new field for this, or can you just send the public key via `jwtSecret`? (Is this just so people can find it easier?) I feel like passing a public key to `jwtSecret` is fine; it's effectively saying "the JWT secret is identified by this public key". What do you think?

Tolli81:
Pitching in as a CLI user, not currently using JWTs but planning to :)

I agree with the point of huge number of jwt flags, it is in fact a bit
overwhelming. Using postgraphilerc.js sounds really good, and that way it's
perhaps easier to provide examples for different use cases.

However mixing secrets and public keys in the same option does not sound as
good, at least not while called one of those. The name could perhaps be
generalized, but I vote for the unambiguous use of two options :)

-Tolli

Þann fim., 6. jún. 2019, 07:53 Benjie Gillam skrifaði <
notifications@github.com>:

src/postgraphile/cli.ts
speller: @benjie I see your point and agree. Will update the edits to use file configuration.

src/postgraphile/cli.ts
speller: The issue is that if Postgraphile issues and verified JWTs by itself, it has to have a pair of priv/pub keys configured. Only when it is used only for issuing JWTs or only for verifying JWTs the single `secret` parameter can be used.

speller: One of the options is to deprecate the `jwtSecret` option and propagate users to use `jwtPublicKey` and `jwtPrivateKey` instead. This will remove any misunderstandings in the future. What do you think about it?

speller: Maybe `jwtSignKey` and `jwtVerifyKey`?

betaboon: i would vote against having a dual-use-kinda parameter (eg `jwtSecret` means secret for symetric, but public-key for asymmetric)

Even tho it is more code i would vote for something like the following, because IMHO stuff should be explicitly clear, especially when related to security/crypto:

- separate signing and verifying-options (as we might have use-cases where postgraphile just does one of the two)
- have `signingAlgorithm` and `verifyingAlgorithm` options
- have `signingSecret`, `signingPublicKey` and `signingPrivateKey` but make the code check (eg `signingSecret` is only applicable+required for symetric algorithms)

speller: @betaboon I think, `jwtSignKey` and `jwtVerifyKey` are the most appropriate names. Because a signing key can be a secret or a private key. But it is used for signing. A verify key can be the same secret or a public key. So, I don't see any issues. Anyway, all of the options: `signingSecret`, `signingPublicKey` and `signingPrivateKey` would be passed to the same parameter to the internal sign function. Separation of these values will lead to bringing a lot of logic about crypto algorithms used by the third-party library. I think it is not a good way in terms of future support to keep the logic in actual state.

src/postgraphile/cli.ts
```diff
@@ -278,7 +278,7 @@ pluginHook('cli:flags:add:webserver', addFlag);
 program
   .option(
     '-e, --jwt-secret <string>',
-    'the secret to be used when creating and verifying JWTs. if none is provided auth will be disabled',
+    'the secret or private key to be used when creating and verifying JWTs. if none is provided auth will be disabled. in case of private key, the value should be in the PEM format. if PostGraphile will be used to verify tokens, public key should be set to the --jwt-public-key parameter',
```
benjie:
```suggestion
    'the secret or private key to be used when creating and/or verifying JWTs. If none is provided, JWT auth will be disabled. If using a private key, the value should be in the PEM format and the public key should be provided via `--jwt-public-key`',
```

src/postgraphile/cli.ts
```diff
@@ -425,6 +426,16 @@ const {
...
+  jwtSignAlgorithm,
+  jwtSignKeyId,
+  jwtSignExpiresIn,
+  jwtSignNotBefore,
+  jwtSignAudience,
+  jwtSignSubject,
+  jwtSignIssuer,
+  jwtSignJwtId,
+  jwtSignNoTimestamp,
+  jwtSignEncoding,
```
benjie: We currently allow passing through `jwtVerifyOptions` in library mode (see [library usage](https://www.graphile.org/postgraphile/usage-library/)); instead of accepting all these options can we just accept one called `jwtSignOptions` instead?

src/postgraphile/cli.ts
benjie: I think the pairing of jwtSecret (private key, or secret string) and jwtPublicKey works well.

src/postgraphile/withPostGraphileContext.ts
```diff
@@ -104,7 +105,7 @@ const withDefaultPostGraphileContext: WithPostGraphileContextFn = async (
 
   const { role: pgRole, localSettings, jwtClaims } = getSettingsForPgClientTransaction({
     jwtToken,
-    jwtSecret,
+    jwtSecretOrPublicKey: jwtPublicKey || jwtSecret,
```
benjie: Lets just pass through values directly right up until they're used.

```suggestion
    jwtSecret,
    jwtPublicKey,
```

src/postgraphile/withPostGraphileContext.ts
```diff
@@ -266,15 +267,15 @@ export default withPostGraphileContext;
 // errors, however, as this will stop the request execution.
 function getSettingsForPgClientTransaction({
   jwtToken,
-  jwtSecret,
+  jwtSecretOrPublicKey,
```
benjie:
```suggestion
  jwtSecret,
  jwtPublicKey,
```

src/postgraphile/withPostGraphileContext.ts
```diff
@@ -266,15 +267,15 @@ export default withPostGraphileContext;
...
   jwtAudiences,
   jwtRole,
   jwtVerifyOptions,
   pgDefaultRole,
   pgSettings,
 }: {
   jwtToken?: string;
-  jwtSecret?: string;
+  jwtSecretOrPublicKey?: string;
```
benjie:
```suggestion
  jwtSecret?: string | Buffer;
  jwtPublicKey?: string | Buffer;
```

src/postgraphile/withPostGraphileContext.ts
```diff
@@ -297,7 +298,7 @@ function getSettingsForPgClientTransaction({
     try {
```
benjie:
```suggestion
    try {
      const jwtVerificationSecret = jwtPublicKey || jwtSecret;
```

src/postgraphile/withPostGraphileContext.ts
```diff
@@ -297,7 +298,7 @@ function getSettingsForPgClientTransaction({
...
       // If a JWT token was defined, but a secret was not provided to the server or
       // secret had unsupported type, throw a 403 error.
-      if (!Buffer.isBuffer(jwtSecret) && typeof jwtSecret !== 'string') {
+      if (!Buffer.isBuffer(jwtSecretOrPublicKey) && typeof jwtSecretOrPublicKey !== 'string') {
```
benjie:
```suggestion
      if (!Buffer.isBuffer(jwtVerificationSecret) && typeof jwtVerificationSecret !== 'string') {
```

src/postgraphile/withPostGraphileContext.ts
```diff
@@ -310,7 +311,7 @@ function getSettingsForPgClientTransaction({
           `Provide either 'jwtAudiences' or 'jwtVerifyOptions.audience' but not both`,
         );
 
-      const claims = jwt.verify(jwtToken, jwtSecret, {
+      const claims = jwt.verify(jwtToken, jwtSecretOrPublicKey, {
```
benjie:
```suggestion
      const claims = jwt.verify(jwtToken, jwtVerificationSecret, {
```

benjie:
(Sorry for the delay, July was epically busy for me.)

src/postgraphile/cli.ts
speller: Should we deprecate the use of `jwtVerify***` options then, to provide the equal configuration possibilities for both signing and verifying? It seems to be surprising for users to have `jwtVerify*` variables and have no `jwtSign*` ones. What do you think?

benjie: Yeah; I'll probably remove them in v5 so it makes sense to deprecate them. I think we'll probably accept a JSON5 object on the CLI in v5; so we can just do `--jwt-sign '{issuer:"postgraphile",noTimestamp:true}'`

benjie: (Give me a shout when you're ready for me to re-review.)

speller:
@benjie I've changed the logic if options and added support for `jwtSignOptions` and `jwtVerifyOptions` in the `.postgraphilerc.js` file. The new `jwtVerifyOptions` configuration parameter has a higher priority than current `jwtVerify*` options. If `jwtVerifyOptions` configuration is set, all the current `jwtVerify*` configurations will be ignored. This is made to make both configurations to be the same in the future, when `jwtVerify*` CLI and RC configurations will be deprecated in favor of the new-style config.

src/postgraphile/cli.ts
```diff
@@ -565,16 +568,19 @@ function trimNulls(obj: object): object {
...
-  algorithms: jwtVerifyAlgorithms,
-  audience: jwtVerifyAudience,
-  clockTolerance: jwtVerifyClockTolerance,
-  jwtId: jwtVerifyId,
-  ignoreExpiration: jwtVerifyIgnoreExpiration,
-  ignoreNotBefore: jwtVerifyIgnoreNotBefore,
-  issuer: jwtVerifyIssuer,
-  subject: jwtVerifySubject,
-});
+const jwtVerifyOptionsActual: jwt.VerifyOptions =
```
benjie: I think we should warn people that their other settings are being ignored if they're using `jwtVerifyOptions`. Since this is security-critical, I think throwing an error makes sense:

```suggestion
if (
  jwtVerifyOptions &&
  (jwtVerifyAlgorithms ||
    jwtVerifyAudience ||
    jwtVerifyClockTolerance ||
    jwtVerifyId ||
    jwtVerifyIgnoreExpiration ||
    jwtVerifyIgnoreNotBefore ||
    jwtVerifyIssuer ||
    jwtVerifySubject)
) {
  throw new Error(
    'You may not mix `jwtVerifyOptions` with the legacy `jwtVerify*` settings; please only provide `jwtVerifyOptions`.'
  );
}
const jwtVerifyOptionsActual: jwt.VerifyOptions =
```

benjie: Rather than using `jwtVerifyOptionsActual` here, could you instead rename the old `jwtVerifyOptions` to, e.g., `rawJwtVerifyOptions` - I find it more obvious if the final state is named sensibly.

benjie: Sorry, I think I miscommunicated.

To make it clearer what I meant, I would like you to change line 430:

https://github.com/graphile/postgraphile/blob/7feea7cc2975a4fd9626e9f3fef7f47f6323569e/src/postgraphile/cli.ts#L430

to use a different name for the destructured property:

```
jwtVerifyOptions: rawJwtVerifyOptions,
```

---

Line 605 

https://github.com/graphile/postgraphile/blob/7feea7cc2975a4fd9626e9f3fef7f47f6323569e/src/postgraphile/cli.ts#L605

would get the reverse treatment:

```
jwtVerifyOptions,
```

---

And replace `jwtVerifyOptionsActual` here with `jwtVerifyOptions`.

---

For me it's more important that the final variable name is correct rather than the initial destructured one (so doing it this way would be consistent with similar patterns in the rest of the codebase).

benjie:
CI is unhappy with the code formatting; please run `yarn prettier:fix`. Thanks! :pray:

src/postgraphile/cli.ts
```diff
@@ -278,7 +278,7 @@ pluginHook('cli:flags:add:webserver', addFlag);
 program
   .option(
     '-e, --jwt-secret <string>',
-    'the secret to be used when creating and verifying JWTs. if none is provided auth will be disabled',
+    'the secret or private key to be used when creating and/or verifying JWTs. If none is provided, JWT auth will be disabled. If using a private key, the value should be in the PEM format and the public key should be provided via `--jwt-public-key`',
```
benjie: The `--jwt-public-key` flag doesn't seem to have been added.

src/postgraphile/withPostGraphileContext.ts
```diff
@@ -295,9 +299,10 @@ function getSettingsForPgClientTransaction({
...
     // as a 403 error because the token is not trustworthy.
     try {
+      const jwtVerificationSecret = jwtPublicKey || jwtSecret;
       // If a JWT token was defined, but a secret was not provided to the server or
       // secret had unsupported type, throw a 403 error.
-      if (!Buffer.isBuffer(jwtSecret) && typeof jwtSecret !== 'string') {
+      if (!Buffer.isBuffer(jwtVerificationSecret) && typeof jwtVerificationSecret !== 'string') {
         // tslint:disable-next-line no-console
         console.error(
           'ERROR: `jwtToken` was provided, but `jwtSecret` was not set to a string or buffer - rejecting request.',
```
benjie: This error needs fixing to differentiate between `jwtSecret` and `jwtPublicKey`
```suggestion
          `ERROR: '${jwtPublicKey ? 'jwtPublicKey' : 'jwtSecret'}' was not set to a string or buffer - rejecting JWT-authenticated request.`,
```

(This will need to be ran through prettier.)

src/postgraphile/cli.ts
speller: I was sure we've decided in our previous conversation to avoid adding new CLI parameters and use config file instead.

benjie: In which case it needs removing from the option description :)

speller: Sorry, seems like I'm confused with the Github's PR interface... Will update sources

speller: Updated. But CI build is failing, I don't understand why...

benjie: It could certainly be more obvious! (PR to improve that would be welcome.)

The issue is this:


```
$ yarn prettier --list-different
$ prettier 'src/**/*.[tj]s' 'postgraphiql/src/**/*.[tj]s' --list-different
src/postgraphile/withPostGraphileContext.ts
```

This means that `src/postgraphile/withPostGraphileContext.ts` has not been formatted with prettier. You should be able to fix it by running `yarn prettier:fix`. My editor runs prettier for me automatically; I don't know how to format code by hand any more :wink:

speller: I just applied your suggested change in the GitHub UI and it failed the prettier check. That's why I was confused as I remember that I didn't change this file on my own.

speller: So, now prettier is happy, but something went wrong in the `package.json`...

benjie: This is not the fault of your PR, it's now passing CI. However looks like a vulnerability has been discovered in lodash in the mean time.

The easiest way to fix this is to delete these four lines from the `yarn.lock`:

https://github.com/graphile/postgraphile/blob/d094541ab63d9f5cab9ac6ef36d37b8ca09f054d/yarn.lock#L5006-L5009

And then run `yarn` again. This should install a newer version of lodash that fixes the vulnerability and appeases Snyk. We don't use `defaultsDeep` so we're technically not vulnerable. Don't worry if you don't want to take this on, I can address it on master next week and then merge it back into this branch.

benjie:
Are you able to add any tests for this, @speller? I think you should be able to mimic what's done for the existing `jwtSecret` tests, but with public/private key secrets.

speller:
@benjie Is PR okay now for merging?

Issue #191, opened 2016-10-27
How to use postgraphql with multiple schemas with same table names


langpavel:
## Example:

``` sql
CREATE SCHEMA s1;
CREATE SCHEMA s2;
CREATE TABLE s1.test_case(ID serial, value text);
CREATE TABLE s2.test_case(ID serial, value text);
```

And then start the server on the schemas s1,s2:
`postgraphql --connection pg://langpavel:xxx@localhost:5434/namecollision --schema s1,s2`

```
Error: Collection of name 'test_cases' already exists in the inventory.
    at Inventory.addCollection (/home/langpavel/.nvm/versions/node/v6.9.1/lib/node_modules/postgraphql/build/interface/Inventory.js:28:19)
    at Object.addPgCatalogToInventory (/home/langpavel/.nvm/versions/node/v6.9.1/lib/node_modules/postgraphql/build/postgres/inventory/addPgCatalogToInventory.js:24:23)
    at Object.<anonymous> (/home/langpavel/.nvm/versions/node/v6.9.1/lib/node_modules/postgraphql/build/postgraphql/schema/createPostGraphQLSchema.js:53:20)
    at next (native)
    at fulfilled (/home/langpavel/.nvm/versions/node/v6.9.1/lib/node_modules/postgraphql/build/postgraphql/schema/createPostGraphQLSchema.js:4:58)
    at process._tickCallback (internal/process/next_tick.js:103:7)
```
## In `psql`, behavior is driven by `search_path`.

For example:

``` sql
INSERT INTO s1.test_case (value) VALUES ('Schema s1');
INSERT INTO s2.test_case (value) VALUES ('Schema s2');

SELECT * FROM s1.test_case; -- Schema s1
SELECT * FROM s2.test_case; -- Schema s2

-- SELECT * FROM test_case;
-- ERROR:  42P01: relation "test_case" does not exist

SET search_path=s1, s2;
SELECT * FROM test_case; -- Schema s1

SET search_path=s2, s1;
SELECT * FROM test_case; -- Schema s2
```
## Suggested implementation

Will be more convenient if `postgraphql --schema` argument behaves like `search_path`.

There should be more options, I like only the first one:
- `--hide-schema-collisions` - accept only first name on search_path 
- `--alias-schema-collisions` - excepth first name, other object will be prefixed with schema.
  
  For example:
  - `s1.test_case` becomes `test_case` because `s1` is first in `--schema` argument, but
[...]

PR #1090, opened 2019-05-29
feat(postgraphile): ignore schema and rely on search_path

michaelbridge:
See https://github.com/graphile/graphile-engine/pull/467 re: issue #191.

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -109,6 +109,11 @@ export interface PostGraphileOptions<
   // from the generated GraphQL schema as general applications don't need them
   // to be exposed to the end user. You can use this flag to include them in
   // the generated schema (not recommended).
+  ignoreSchemas?: boolean;
+  // Not recommended. By default, database objects within generated queries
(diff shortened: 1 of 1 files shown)


src/interfaces.ts
```diff
@@ -109,6 +109,11 @@ export interface PostGraphileOptions<
   // from the generated GraphQL schema as general applications don't need them
   // to be exposed to the end user. You can use this flag to include them in
   // the generated schema (not recommended).
+  ignoreSchemas?: boolean;
+  // Not recommended. By default, database objects within generated queries
+  // consist of the PostgreSQL schema name and table name separated by a dot.
+  // Set 'true' to omit the schema name, such that relevant table is determined
+  // based on the current search_path.  Useful in multi-tenant applications.
```
benjie: The comment should be above the field rather than below (it's because we auto-generate the docs from the source). We'll revisit this once we've decided on the right direction for the underlying Graphile Engine PR 👍

benjie:
Closing as https://github.com/graphile/graphile-engine/pull/467 is closed.

PR #1148, opened 2019-08-30
fix: make sure x-graphql-event-stream url is domain relative

### LINKED ISSUE #1143: x-graphql-event-stream is invalid when graphqlRoute is "/"
when `graphqlRoute` is "/", `x-graphql-event-stream` become `//stream` and that's understood as a protocol relative uri by browsers, so they try to connect to https://stream/

one easy option would be to remove trailling "/" from `graphqlRoute` when building the `x-graphql-event-stream` header

I'd like it if `graphqlRoute: ""` was not mapped to `graphqlRoute: "/graphql"` but only when `null` or `undefined` but this would break backward compatibility

mathroc:
closes #1143

### DIFF
--- a/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
+++ b/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
@@ -208,7 +208,7 @@ export default function createPostGraphileHttpRequestHandler(
   // Gets the route names for our GraphQL endpoint, and our GraphiQL endpoint.
   const graphqlRoute = options.graphqlRoute || '/graphql';
   const graphiqlRoute = graphiql ? options.graphiqlRoute || '/graphiql' : null;
-  const streamRoute = `${graphqlRoute}/stream`;
+  const streamRoute = `${graphqlRoute.replace(/\/*$/, '')}/stream`;
 
   // Throw an error of the GraphQL and GraphiQL routes are the same.
   if (graphqlRoute === graphiqlRoute)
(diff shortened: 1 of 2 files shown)


src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -387,7 +387,9 @@ export default function createPostGraphileHttpRequestHandler(
           /<\/head>/,
           `  <script>window.POSTGRAPHILE_CONFIG=${safeJSONStringify({
             graphqlUrl: `${externalUrlBase}${graphqlRoute}`,
-            streamUrl: watchPg ? `${externalUrlBase}${graphqlRoute}/stream` : null,
+            streamUrl: watchPg
+              ? `${externalUrlBase}${graphqlRoute}`.replace(/\/*$/, '') + `/stream`
```
benjie: I’m not at my computer; but would this be equivalent? If so, I’d much prefer it for DRY, simplicity and performance reasons 👍

```suggestion
              ? `${externalUrlBase}${streamRoute}`
```

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -562,7 +564,10 @@ export default function createPostGraphileHttpRequestHandler(
     if (watchPg) {
       // Inform GraphiQL and other clients that they can subscribe to events
       // (such as the schema being updated) at the following URL
-      res.setHeader('X-GraphQL-Event-Stream', `${externalUrlBase}${graphqlRoute}/stream`);
+      res.setHeader(
+        'X-GraphQL-Event-Stream',
+        `${externalUrlBase}${graphqlRoute}`.replace(/\/*$/, '') + `/stream`,
```
benjie: As above

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
mathroc: yes, I didn't notice before that `externalUrlBase` can't end with à `/`

mathroc:
should be fine now

PR #1157, opened 2019-09-23
chore: move from TSLint to ESLint

benjie:


benjie:
This does weaken linting somewhat; but we can add more power to it over time. For now, using ESLint is important to standardise our tooling.

PR #1167, opened 2019-10-16
feat(jwt): support lazy public key loading for verification

ab-pm:
I did need to use the callback version of `jwt.verify` and added proper types as well as prevented throwing on functions. @speller's commit came just right :-)


src/postgraphile/withPostGraphileContext.ts
```diff
@@ -277,8 +277,8 @@ function getSettingsForPgClientTransaction({
   pgSettings,
 }: {
   jwtToken?: string;
-  jwtSecret?: string | Buffer;
-  jwtPublicKey?: string | Buffer;
+  jwtSecret?: jwt.Secret;
```
benjie: Should we be doing the same for `jwtSecret`?

ab-pm: If you mean `| GetPublicKeyOrSecret`, that doesn't work as [`jwt.sign`](https://github.com/auth0/node-jsonwebtoken#jwtsignpayload-secretorprivatekey-options-callback) does not accept a callback but requires the secret to be available synchronously. That's why I was really happy to see that @speller had just separated the types…

benjie: :+1: Cool; always good to have justification. :)

ab-pm:
Thanks for forcing me to write the tests - turns out had not worked by simply relaxing the precondition :-/

ab-pm:
This is weird, `eslint(prettier/prettier)` complains about the trailing commas in my vscode, but `yarn prettier` does actually expect them… Let's make the CI happy.

benjie:
Hmmm; prettier and ESLint should both be pulling from the prettier config file here: https://github.com/graphile/postgraphile/blob/master/prettier.config.js

Trailing commas are expected.

ab-pm:
Is it possible to re-run the CI now that https://github.com/graphile/graphile-engine/pull/546 is merged?

benjie:
Alas we have to ship it first :disappointed: 

Maybe I should merge this into the monorepo after all.

benjie:
Merged in the latest engine.

src/postgraphile/withPostGraphileContext.ts
```diff
@@ -103,7 +103,7 @@ const withDefaultPostGraphileContext: WithPostGraphileContextFn = async (
   // Warning: this is only set if pgForceTransaction is falsy
   const operationType = operation != null ? operation.operation : null;
 
-  const { role: pgRole, localSettings, jwtClaims } = getSettingsForPgClientTransaction({
+  const { role: pgRole, localSettings, jwtClaims } = await getSettingsForPgClientTransaction({
```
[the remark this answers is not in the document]
ab-pm: My thoughts exactly :-)

PR #1174, opened 2019-11-06
feat(postgraphiql): save headers to localStorage

phryneas:
It was pretty annoying to me that PostGraphiQL "forgets" the headers at every page refresh (we have schema polling disabled).
This will store the headers in localStorage as soon as they're valid JSON and retrieve them on component creation.

### DIFF
--- a/postgraphiql/src/components/PostGraphiQL.js
+++ b/postgraphiql/src/components/PostGraphiQL.js
@@ -71,6 +71,8 @@ const websocketUrl = POSTGRAPHILE_CONFIG.graphqlUrl.match(/^https?:/)
       l.port !== 80 && l.port !== 443 ? ':' + l.port : ''
     }${POSTGRAPHILE_CONFIG.graphqlUrl}`;
 
+const STORAGE_KEY_HEADERS_TEXT = 'PostGraphiQL_headersText';
+
 /**
  * The standard GraphiQL interface wrapped with some PostGraphile extensions.
  * Including a JWT setter and live schema udpate capabilities.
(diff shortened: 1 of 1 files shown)


benjie:
I understand the frustration, but this is another convenience vs security issue; storing JWTs and other private header data to LocalStorage opens an attack path that the users of GraphiQL might not expect. I'm very uncomfortable making storing it the default behaviour. Maybe we can make it opt-in either via a CLI/library option, or with a UI option like a checkbox "save these headers :information_source:" above the headers input? The info button should explain why it's not enabled by default.

phryneas:
I guess this checkbox would calm your concerns? :)

postgraphiql/src/components/PostGraphiQL.js
```diff
@@ -93,6 +99,21 @@ class PostGraphiQL extends React.PureComponent {
...
+      this._storage.get(STORAGE_KEYS.SAVE_HEADERS_TEXT) || 'false',
+    );
+    if (saveHeadersText) {
+      this.state.saveHeadersText = saveHeadersText;
+      const headersText = this._storage.get(STORAGE_KEYS.HEADERS_TEXT);
+      if (headersText) {
+        this.state.headersText = headersText;
+      }
+    }
+  }
```
benjie: Please move this code to the `state =` property initialiser above, and simplify it, e.g.

```js
    saveHeadersText: this._storage.get(STORAGE_KEYS.SAVE_HEADERS_TEXT) === 'true' ? true : false,
    headersText: this._storage.get(STORAGE_KEYS.HEADERS_TEXT) || "",
```

(Note headersText doesn't need to respect saveHeadersText here - if it's in local storage then we should display it anyway. The code later should ensure that it's not stored unless it's checked.)

benjie:
(Also please merge the latest master into this branch; we do git squashing so you don't need to worry about keeping the history linear.)

postgraphiql/src/components/PostGraphiQL.js
phryneas: Okay. You sure you want an empty string instead of the default `'{\n"Authorization": null\n}\n'`up until now?

benjie: Use the existing default :) The above comment is just an approximation.

phryneas:
done :)

PR #1177, opened 2019-11-07
feat(tags): upgrade engine & `postgraphile.tags.json5` support

benjie:
Building on https://github.com/graphile/graphile-engine/pull/541, this adds support for a `postgraphile.tags.json5` file that can contain all your smart tags and is automatically imported (and watched, if `--watch` is specified) in CLI mode. In library mode you will need to provide your own `makePgSmartTagsPlugin` or `makeJSONPgSmartTagsPlugin` instead - we didn't want to use the `fs` library in any more places so that we could remain compatible with serverless/Lambda.

Here's an example config file. You can use simple JSON format, but by using JSON5 you can add comments and optionally omit quotes in many places.

```json5
{
  version: 1,
  config: {
    /*
     * There can be entries here for:
     *
     * - `class`: for tables, composite types, views and materialized views
     * - `attribute`: for columns/attributes (of any 'class' type)
     * - `constraint`: for table constraints
     * - `procedure`: for functions/procedures
     */
    class: {
      /*
       * The next level describes the named type. We've just used the table
       * name `post` but it could be `"my_schema.post"` if you have multiple
       * tables with the same name and you don't want this rule to apply to
       * all of them.
       */
      post: {
        // This will override the description sourced from PostgreSQL
        description: "A post within our forum.",

        // You can omit this if you don't want to add any tags
        tags: {},

        /*
         * We've added a shortcut to class-types so you can tag/describe
         * columns at the same time of the class.
         */
        columns: {
          // `body` is the column name
          body: {
            description: "The body of the post",
            tags: {
              // Editing a post is not allowed
              omit: "update"
            }
          }
        }
      }
    }
  }
}
```


ab-pm:
> You can use simple JSON format, but by using JSON5 you can add comments and optionally omit quotes in many places.

How about just supporting YAML? :-)

benjie:
Hell no.

YAML is a maintenance burden (the number of times dependabot raises fixes against projects using yaml, _sigh_), it also allows lots of potentially problematic things like circular references, and it's a really hard format for users to "get right first time" because of how loose it is - mistakes can be subtle and hard to grok. JSON5 gives us the convenience and familiarity of a JS-like syntax with JSON-like promises of straightforward objects, and no-one needs to learn a new weird syntax.

benjie:
Here's a discussion that shaped my choice of JSON5, in case you want to read more on the subject: https://github.com/microsoft/rushstack/issues/1088

PR #1191, opened 2019-11-20
fix(graphiql): fix operation detection on multi-op documents

hsanoyan:
in isSubscription not all operations always run, so
if there is provided name of operation, check only that operation type.

### DIFF
--- a/postgraphiql/src/components/PostGraphiQL.js
+++ b/postgraphiql/src/components/PostGraphiQL.js
@@ -41,11 +41,21 @@ const defaultQuery = `\
 #
 `;
 
-const isSubscription = ({ query }) =>
-  parse(query).definitions.some(
+const isSubscription = ({ query, operationName }) => {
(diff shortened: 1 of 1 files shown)


postgraphiql/src/components/PostGraphiQL.js
```diff
@@ -41,11 +41,21 @@ const defaultQuery = `\
...
+    if (definition) {
+      return definition.operation === 'subscription';
+    }
+  }
+
+  return definitions.some(
     definition =>
       definition.kind === 'OperationDefinition' && definition.operation === 'subscription',
   );
+};
```
benjie: My mistake, I forgot we had access to `graphql` here; we can just use the helper:

```js
import { getOperationAST, parse } from 'graphql';

const isSubscription = ({ query, operationName }) => {
  const operation = getOperationAST(parse(query), operationName);
  return operation.operation === 'subscription';
};
```

PR #1200, opened 2019-11-27
Split JWT verification functionality to the separate module

speller:
Split JWT verification functionality to the separate module making it available for use in plugins

### DIFF
--- a/src/postgraphile/jwtVerify.ts
+++ b/src/postgraphile/jwtVerify.ts
@@ -1,5 +1,6 @@
 import jwt = require('jsonwebtoken');
 import { mixed } from '../interfaces';
+import { JsonWebTokenError } from 'jsonwebtoken';
 
 const undefinedIfEmpty = (
   o?: Array<string | RegExp> | string | RegExp,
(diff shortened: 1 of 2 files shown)


ab-pm:
This doesn't quite work out. You should only either call the `callback` *or*  `return` a result, but not both.

speller:
@ab-pm You're right, I missed the async run logic. Fixed it now. On my instance, it runs well with callbacks in the standard logic of verifying the Authentication header, and synchronously with my plugin.

speller:
Just need to fix tests

ab-pm:
I think it would make more sense to split this in two parts, `validateJwtVerifyOptions` and `processClaims` (what you currently named `rest`), and not to call `verify` itself directly. Or maybe you even drop the `processClaims` and just put that back in `withPostgraphileContext`, as I doubt you need to fetch the `role` in the plugin.

benjie:
This is quite hard to review currently; could you keep the code in the same file initially whilst rewriting it, we can move it to a separate file in a later PR. Try and keep the diff as minimal as possible (feel free to ignore whitespace changes though).

This function is central to may people's security in PostGraphile and is also performance sensitive so I'm keen to ensure it's reviewed in detail before being merged and it's quite hard to tell currently what's actually been changed and what is the same.

speller:
As Postgraphile has no ability to work with asynchronous input parameters resolution when building sql query and thus can not handle asynchronous JWT verification, this PR doesn't have any sense. Closing.

Issue #1220, opened 2020-01-25
Fastify plugins integration


madflow:
I'm submitting a ... 

* [x] bug report
* [x] feature request

PostGraphile version: latest

Hi!

This was triggered by: https://github.com/graphile/postgraphile/pull/789#issuecomment-573980706

Fastify uses a [plugin system](https://www.fastify.io/docs/latest/Plugins/) for registering plugins like [fastify-compress](https://github.com/fastify/fastify-compress) or [fastify-helmet](https://github.com/fastify/fastify-helmet).

Also there is a [Express middleware compatible](https://www.fastify.io/docs/latest/Middleware/) `use` integration.

I mount Postgraphile as a middleware.

Some Fastify plugins seem to work with this setup (`fastify-helmet`) and some do not: for example `fastify-compress`.

However - it is possible to achieve compression by use the (Express!) `compression` middleware.

This was my testcase:

```JS
// yarn add postgraphile fastify fastify-helmet fastify-compress compression
// DATABASE_URL=... node server.js

const { postgraphile } = require("postgraphile");
const fastifyHelmet = require("fastify-helmet");
const expressCompression = require('compression');
const fastifyCompression = require('fastify-compress');

const Fastify = require('fastify');

const app = Fastify({
  logger: true,
});

// This is working
app.register(fastifyHelmet);

// app.use(expressCompression({threshold: 0})); // working

// app.register(fastifyCompression, {threshold: 0}); // not working

app.use(
  postgraphile(process.env.DATABASE_URL, 'public', {
    graphiql: true,
    enhanceGraphiql: true
  })
);

app.listen(process.env.PORT || 3000, err => {
  if (err) {
    app.log.error(err);
    process.exit(1);
  }
});
```

Hopefully I did everything correct here. 

I would aplaude tighter coupling of Fastify (and plugin system) and Postgraphile. 

From what I gather is that the Koa compression issues has been fixed my some Koa specific "hacks". Maybe this would suffice for Fastify, too? Or - should just Express middleware used with the Fastify compat layer?

I can provide a PR with a failing test for the `fastify-compress` issue if desired.

madflow:
Just a fyi: We switched back to `Express` :S . So this issue and the PR #1221 is not critical/important to "me" anymore. Maybe it is important for others - so I am not closing this issue or the PR if not desired.

alemagio:
Hi
don't know if someone might still want it but I recently released a plugin to integrate postgraphile with fastify.
[...]

PR #1221, opened 2020-01-25
Tests for Fastify plugins

madflow:
This PR 

* Adds a test to check if Fastify plugins do their job when used as http server
* Bumps Fastify to the latest version and installs fastify-helmet and fastify-compress as dev dependencies

* The fastify-compress test is failing. It does pass when the `compress` middleware is used instead of the plugin
* See #1220

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,6 +101,8 @@
     "eslint-plugin-prettier": "^3.1.1",
     "express": "^4.14.0",
     "fastify": "^2.2.0",
+    "fastify-compress": "^2.0.0",
+    "fastify-helmet": "^3.0.2",
     "html-webpack-inline-source-plugin": "^0.0.10",
     "html-webpack-plugin": "^3.2.0",
     "jest": "^24.7.1",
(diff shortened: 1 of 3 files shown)


benjie:
In koa we had to hack around it like this: https://github.com/graphile/postgraphile/blob/d9d5c59d106a54612fb3738003b7e631ab6d062e/src/postgraphile/http/koaMiddleware.ts#L19-L46

Maybe a similar hack would work for fastify? It might be that we should use `res.send()` rather than `res.end()` here, and then call `res.end()` separately?

https://github.com/graphile/postgraphile/blob/d9d5c59d106a54612fb3738003b7e631ab6d062e/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts#L895

madflow:
> Maybe a similar hack would work for fastify? It might be that we should use res.send() rather than res.end() here, and then call res.end() separately?

I do not know unfortunatly. Regarding the  ```res.send()```: `ServerResponse` does not have a `send` method. Is there be more refactoring implied or am I missing something?

benjie:
Oh yeah, `res.send` is an Express API. We'd need to figure out what fastify compress is expecting and then use that.

PR #1233, opened 2020-02-19
Added websocket transport support for query & mutation.

klemens-morgenstern:
`subscriptions-transport-ws` supports websocket transport. I just modified the `execute` function which did the trick for me and seems to work with all the auth etc., too.

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -67,7 +67,7 @@ export interface PostGraphileOptions<
   // Enable GraphQL websocket transport support for subscriptions (you still need a subscriptions plugin currently)
   subscriptions?: boolean;
   // ENable GrqphQL websocket transport support for query & Mutation
-  websocketTransport?: boolean;
+  websocketOperations?: 'all' | 'subscriptions' | 'none';
   // [EXPERIMENTAL] Enables live-query support via GraphQL subscriptions (sends updated payload any time nested collections/records change)
   live?: boolean;
   // [EXPERIMENTAL] If you're using websockets (subscriptions || live) then you
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -66,6 +66,8 @@ export interface PostGraphileOptions<
   ownerConnectionString?: string;
   // Enable GraphQL websocket transport support for subscriptions (you still need a subscriptions plugin currently)
   subscriptions?: boolean;
+  // ENable GrqphQL websocket transport support for query & Mutation
+  websocketTransport?: boolean;
   // [EXPERIMENTAL] Enables live-query support via GraphQL subscriptions (sends updated payload any time nested collections/records change)
   live?: boolean;
   // [EXPERIMENTAL] If you're using websockets (subscriptions || live) then you
--- a/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
+++ b/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
@@ -413,7 +413,8 @@ export default function createPostGraphileHttpRequestHandler(
         )
       : null;
 
-    if (subscriptions) {
+    if (subscriptions || options.websocketOperations == 'all'
+                      || options.websocketOperations == 'subscriptions') {
       const server = req && req.connection && req.connection['server'];
       if (!server) {
         // tslint:disable-next-line no-console
--- a/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
+++ b/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
@@ -413,8 +413,11 @@ export default function createPostGraphileHttpRequestHandler(
         )
       : null;
 
-    if (subscriptions || options.websocketOperations == 'all'
-                      || options.websocketOperations == 'subscriptions') {
+    if (
+      subscriptions ||
+      options.websocketOperations == 'all' ||
+      options.websocketOperations == 'subscriptions'
+    ) {
       const server = req && req.connection && req.connection['server'];
       if (!server) {
         // tslint:disable-next-line no-console
--- a/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
+++ b/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
@@ -415,7 +415,7 @@ export default function createPostGraphileHttpRequestHandler(
 
     if (
       subscriptions &&
-      (options.websocketOperations === 'all' || options.websocketOperations === 'subscriptions')
+      options.websocketOperations !== 'none'
     ) {
       const server = req && req.connection && req.connection['server'];
       if (!server) {
--- a/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
+++ b/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
@@ -414,9 +414,10 @@ export default function createPostGraphileHttpRequestHandler(
       : null;
 
     if (
-      subscriptions ||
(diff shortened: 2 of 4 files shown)


benjie:
Auth is a little different for websockets so I had to disable queries/mutations so that it wasn’t a breaking change for people using pgSettings/etc in interesting ways. I wasn’t planning to add support until v5, but we could add this behind a flag if you like?

klemens-morgenstern:
Ah, that makes way more sense; it looked a bit weird that it was just an exception. I guess my `pgSettings` is super boring. 

Yes, a flag or plugin works for me. 

Thanks for all the great work!

src/interfaces.ts
```diff
@@ -66,6 +66,8 @@ export interface PostGraphileOptions<
   ownerConnectionString?: string;
   // Enable GraphQL websocket transport support for subscriptions (you still need a subscriptions plugin currently)
   subscriptions?: boolean;
+  // ENable GrqphQL websocket transport support for query & Mutation
```
benjie: This looks like steganography!
```suggestion
  // Enable GraphQL websocket transport support for query & mutation
```

src/interfaces.ts
```diff
@@ -66,6 +66,8 @@ export interface PostGraphileOptions<
...
+  websocketTransport?: boolean;
```
benjie: i’m not quite happy with this naming but my suggestions aren’t better. Let me ponder.

enableQueriesAndMutationsOverWebsockets
allOperationsOverWebsockets
fullWebsocketSupport
websocketExecute
websocketOperations
graphqlOverWebsocket
wsAllOps

klemens-morgenstern:
Ah yes, the good ol' naming. Let me throw a few in.

 - wsQueryMutation
 - generalWebsocket
 - generalWebsocketTransport
 - gqlOverWebsocket

Alternatively, instead of a flag, the actual options could be like this (`subscription` being kept for compatibility but technically deprecated) 

```ts
  websocketTransport?: boolean|"all"|"only-subscriptions";
```

with `websocketTransport: true` being equal to `websocketTransport: "all", while `subscriptions: true` is equal to `websocketTransport: "only-subscriptions"`.

benjie:
This is better. Great suggestion. Lets drop the boolean and go with:

```
websocketOperations?: "all" | "subscriptions" | "none"
```

Internally it should default to "subscriptions". In addition to the changes in this PR we should also change it here:

https://github.com/graphile/postgraphile/blob/dd2b6b2b364efd015e46cd1faab4d37d2fd9fbad/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts#L416

The reason is that the `subscriptions` flag currently opts you into a number of breaking changes that most people won't notice (e.g. it requires select grants on all primary keys), so this'd give a way to use `subscriptions: true` without enabling subscriptions... :eyes:

In future we can expand it to support query/mutation over websockets without allowing subscriptions themselves. We'll probably drop the `subscriptions` option from v5 as it'll be enabled by default.

klemens-morgenstern:
Any further things I can improve?

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -413,7 +413,11 @@ export default function createPostGraphileHttpRequestHandler(
         )
       : null;
 
-    if (subscriptions) {
+    if (
+      subscriptions ||
+      options.websocketOperations == 'all' ||
+      options.websocketOperations == 'subscriptions'
```
benjie: The subscriptions flag also needs to be set, I think - other things depend on it.

```suggestion
      subscriptions && (
        options.websocketOperations === 'all' ||
        options.websocketOperations === 'subscriptions'
      )
```

src/postgraphile/http/subscriptions.ts
```diff
@@ -188,9 +189,12 @@ export async function enhanceHttpServerWithSubscriptions<
     {
       schema,
       validationRules: staticValidationRules,
-      execute: () => {
-        throw new Error('Only subscriptions are allowed over websocket transport');
-      },
+      execute:
+        options.websocketOperations == 'all'
```
benjie:
```suggestion
        options.websocketOperations === 'all'
```

klemens-morgenstern:
@benjie Just updated the build to pass the remaining format issue. Any thing else i can do to help you with this?

benjie:
Leave it with me :+1:

russelgal:
Looks like there is a mutation error
```{
  "errors": [
    {
      "errcode": "3B001",
      "extensions": {
        "exception": {
          "errcode": "3B001"
        }
      },
      "message": "savepoint \"graphql_mutation\" does not exist",
      "locations": [
        {
          "line": 1,
          "column": 42
        }
      ],
      "path": [
        "updateFormById"
      ]
    }
  ],
  "data": {
    "updateFormById": null
  }
}

File: src/postgraphile/http/createPostGraphileHttpRequestHandler.ts [hunk 5]
```diff
@@ -413,7 +413,10 @@ export default function createPostGraphileHttpRequestHandler(
         )
       : null;
 
-    if (subscriptions) {
+    if (
+      subscriptions &&
+      (options.websocketOperations === 'all' || options.websocketOperations === 'subscriptions')
```
benjie, line 418: Need to handle the current case without breaking. Really we want to make `websocketOperations` have a default; but this will suffice.

```suggestion
      (!options.websocketOperations || options.websocketOperations === 'all' || options.websocketOperations === 'subscriptions')
```

benjie: Actually we can handle this in a much simpler way:

```suggestion
      options.websocketOperations !== 'none'
```

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -413,7 +413,10 @@ export default function createPostGraphileHttpRequestHandler(
         )
       : null;
```
benjie, line 415: We need to add some validation of this option somewhere. This is almost certainly not the best place...

```suggestion

    if (options.websocketOperations && ['all', 'subscriptions', 'none'].indexOf(options.websocketOperations) === -1) {
      throw new Error("Invalid websocketOperations argument");
    }
```

klemens-morgenstern:
They all look fine to me. Regarding the security issue - tell me if I get this right: you can't even invalidate the websocket session once established. That is if the attacker hits the `pgSettings` and then keeps the websocket alive, his session won't be invalidated even if you were to reset everything establishing the connection. 
If that is the case, this seems like a flaw in the websocket transport and would need to be addressed before mergint that.

benjie:
That's one issue (which can be solved by using the session implementation in Graphile Starter which uses database-stored sessions which can be deleted and thus the RLS policies that pull from them would start failing); another is that websockets don't obey the same-origin policy so it's an easy to exploit attack vector if people don't set up their websocket connections well enough (this is still an issue with subscriptions, but it becomes significantly more dangerous when queries and mutations are allowed). There's also performance, stability and scalability concerns; to hear those in more depth I recommend you listen to the section of the last GraphQL Working Group starting here:

https://www.youtube.com/watch?v=FYF15RA9H3k&feature=youtu.be&t=3972

benjie:
This was incorporated into #1338; now we're using a safer graphql websocket transport I'm happier opening  this up :+1: 

Thanks for your work on this @klemens-morgenstern and sorry it took so long to resolve.

Issue #1073, opened 2019-05-16
Provide a callback function to handle serious errors


benjie:
Rather than logging/exiting, allow the application to handle the fail. This needs to give the application a way of retrying (e.g. a retry function they can call after a delay) should they wish to, which also allows them to override the backoff algorithm.

pyramation:
any workarounds for this? maybe an error handler that we can pass into the config? if the handler is specified, the app won't just crash or keep trying, the app can sort out depending on the use case (similar to https://github.com/graphile/postgraphile/issues/1042)

pyramation:
why not just throw this? https://github.com/graphile/postgraphile/blob/master/packages/postgraphile/src/postgraphile/postgraphile.ts#L141

pyramation:
also, can you explain what use cases are for the `attempts`? https://github.com/graphile/postgraphile/blob/master/packages/postgraphile/src/postgraphile/postgraphile.ts#L137

shouldn't this be handled at an infrastructure level?

pyramation:
Here's a quick attempt at an idea: https://github.com/graphile/postgraphile/pull/1302/files

caveat: To avoid the typescript yelling for returning in the while loop, I just (for proof of concept) had the function return a GraphQLSchema, but I know that's incorrect.

pyramation:
should I be looking in `postgraphile-core`?

benjie:
The error is generated asynchronously so it couldn’t be caught in that try/catch - the code has already moved on.

pyramation:
so for v5, do you plan to do this retry logic? 

I would suggest moving it to the infrastructure level, and I'd even help prepare an example for you to see how to decouple it from a devops perspective :) let me know and I can start sending some example repos

pyramation:
also, would you be open to merging a PR or helping me do this PR for v4? My code in my example does what I need from an ops perspective. I just couldn't get the Typescript to be perfect since the function clearly doesn't return a schema. Potentially it may be better to make an option that doesn't call the `exit()`?

Folks using k8s, docker, or anyone running more than one instance of graphile would run into this issue, and the PR as an example I made does the trick, although not perfectly

pyramation:
updated the PR: https://github.com/graphile/postgraphile/pull/1302

I cannot see why the test has failed.. the app it's connected to won't show me anything. Happy to fix

benjie:
Discussion in #1302

benjie:
[...]

PR #1302, opened 2020-06-20
Feat/handle serious error

pyramation:


pyramation:
@benjie I can't see whatever the security/snyk is to fix the PR, please DM or let me know so I can fix :)

benjie:
Snyk doesn't like running against the v4 and v5 branches; I keep telling it to delete the manifest but it keeps coming back.

Screenshot_20200630_122345

I just ignore it now :grimacing:

benjie:
What is the desired behaviour you're going for? At the moment, this seems to result in a response of `{"errors":[{}]}` due to a bug in the handling of non-GraphQLError errors here:

https://github.com/graphile/postgraphile/blob/66c2b4a3713cacd0abb891cb813d45d07b7f6ba3/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts#L862

You've asked in #1073 why PostGraphile has retry logic\* rather than letting the infrastructure level handle it, but by default PostGraphile _does_ let the infrastructure level handle it ─ by exiting the process, the infrastructure can perform a (delayed) restart. I guess you're thinking more of a "health-check" style infrastructure approach, where the infra pings the GraphQL API with something harmless such as the `{__typename}` query and if it receives errors it restarts? Or do you have something else in mind.

\* _We have retry logic because it's a common request from people running PostGraphile with Docker, and because PostGraphile is often not the only middleware you're using, and because a temporary outage of the PostgreSQL server should not prevent the rest of your server from running._

I'd love to understand more about your setup and exactly what your desired behaviour is so that I can analyse this PR better. Currently you've taken a function that was designed to never reject and made it possibly reject indirectly (via the throw example), but I'm not sure what your intentions are beyond that.

pyramation:
I think you're right that may be I confused some infra vs health-check style restarts. Let me explain my use case better :)

So I'm running more than one instance of postgraphile. 

And essentially in the same spirit of
> because a temporary outage of the PostgreSQL server should not prevent the rest of your server from running

I don't want one bad error of one of the instances (of which there could be many) to bring down my entire server.

I've created an LRU cache of postgraphile middlewares, and depending on the request parameters, return the appropriate one. This means I could have 15 instances of postgraphile. If one has schemas with a name conflict (which is what I think I was running in to) it crashes the entire stack. Now, I'm doing somethings that are dynamic letting end users create objects, so let's assume I cannot control if there is a name conflict for this case.

So if this was a user who created something that now crashes the server, calling `process.exit`, correctly as you pointed out, will force my infrastructure to restart. This is not great because it could act as a mini attack, since the amount of time it takes infrastructure to boot (I'm using npx in a container) is non-trivial compared to throwing and error and moving on. For me the difference is like 4 seconds vs 4 ms. That's potentially harmful given it's triggered upon a URL request that an end user can produce.

For me, the throwing the error allows it so if one bad end user creates a set of schemas that happen to produce a bad error, it doesn't harm the other clients.

benjie:
Putting myself in your shoes... I think it might be better addressed by enabling `retryOnInitFail` and adding a schema wait timeout within PostGraphile's server; so if the schema's not ready within, say, 5 seconds the request should raise a 503 or similar saying the service is temporarily unavailable. This would be a good addition to `retryOnInitFail` anyway, and maybe should be enabled by default with a large timeout of, say, 120 seconds (configurable). Sending a signal to the app to inform it the schema is not available would also be a nice addition, so that the app can choose to shutdown the middleware.

This is a non-trivial addition but also not a huge amount of work; would you like to have a go at accomplishing it this way? Would it solve your needs?

pyramation:
> Sending a signal to the app to inform it the schema is not available would also be a nice addition

I totally agree! This is exactly what I need.

So to understand better your proposal, how exactly would we allow `retryInitOnFail` to send a signal? Is there a way to do it from the error when it tries to create the schema?

Let's assume that for a particular case the schema can never be ready, because it has a naming collision, for example. And assuming you have other services running, the goal is to move on from this error in milliseconds -- in the edits I made in this PR, it totally works: https://github.com/graphile/postgraphile/pull/1302/files because the app moves on within milliseconds, not seconds. 

If there was a better way to achieve that, I would totally help build it!

benjie:
We don’t currently know if errors are recoverable or not; even a naming collision can be fixed with a smart comment applied by a later script.

pyramation:
How do you feel about the changes I made? https://github.com/graphile/postgraphile/pull/1302/files

It covers the use case of moving on within milliseconds without killing the process.

benjie:
I don't think the changes you made currently solve the issue in a cohesive manner, since the code that consumes `createGqlSchema` does not expect it to reject (it was designed to never reject), so any requests waiting for that schema will hang in perpetuity (I think, or at least the error isn't likely to be handled well). The changes needed to allow for this are probably a little deeper.

benjie:
Please try https://github.com/graphile/postgraphile/pull/1342 and see if it meets your needs.

Issue #1042, opened 2019-04-11
Process exits on failed initial schema build even in "library mode"


AlexGustafsson:
I'm submitting a ... 

* [x] bug report (of sorts)
* [ ] feature request
* [ ] question 

PostGraphile version: 4.3.3

Steps to reproduce:

```
const express = require('express');
const {postgraphile} = require('postgraphile');

const app = express();

// The IP should NOT have Postgresql listening
const postgraphileMiddleware = postgraphile('127.0.0.1', ...);

app.use(postgraphileMiddleware);
```

Current behavior:

The process exits, regardless of error handling, due to not being able to connect to the host.

```
A fatal error occurred when building the initial schema, so the process will now exit. Error details:

Error: connect ECONNREFUSED 127.0.0.1:5432
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1083:14)
```

This is not especially graceful:

https://github.com/graphile/postgraphile/blob/3ebea166b2099b1d731e50b1f225d7c2f25d8abc/src/postgraphile/postgraphile.ts#L192-L205

Expected behavior:

The process should not be exited. At least not when used as a library. Errors should be escalated and be the chore of the application itself.

benjie:
Indeed; this only occurs at startup as a protection against misconfiguration, but the `process.exit(1)` is indeed too harsh and is particularly problematic for people using the software via Docker/Kubernetes. This is on my near-term TODO list.

Closing as duplicate of #1041

benjie:
The new "retry on fail" behaviour introduced in #1044 (v4.4.0-beta.10) is being disabled by default in #1062 (which will be released with v4.4.0-rc.1); instead you must opt into the feature:

**To _not_ exit on init fail you need to set either `--retry-on-init-fail` on the CLI or `retryOnInitFail: true` in the library.**

Sorry for this inconvenience, but I wanted to ensure that 4.4 is not a breaking change for existing users who may be relying on the exit-on-fail behaviour of v4.0-v4.3.

rc1 should (hopefully) be out in the next few hours.

jmcudk:
I feel like this fix didn't really deal with the initial issue, in my specific application I want to be able to handle errors and exit gracefully rather than just having postgraphile breaking everything and spewing an extravagant amount of logs, my personal preference would be that instead of logging and calling exit, throwing a specific error that the application can catch and deal with.

benjie:
[...]

PR #1315, opened 2020-07-29
feat: add option for throwOnInitFail to handle fails on schema build

charlex:
Adds option named `throwOnInitFail` for #1042 to offer an alternative to exiting the process during library usage.

My usecase: I want to wrap `postgraphile` around a `try {} catch (e) {}`:

 `try { postgraphile() } catch(e) {app.get('/', (req, res) => res.send('There was an error starting Postgraphile'))}`

Note on why this is "Draft": I started this before I noticed that #1302 was open. This PR probably has the same issue of uncaught error handling (I think @benjie's last comment might be referring to this). I figured I'd open this nonetheless to offer a potential different option.

--

I didn't implement this yet, but this commented out TS snippet kind of hints at a third way, a more explicit way, to handle this option:

```ts
type ValueOf<T> = T[keyof T];

export enum OnInitFail {
   ThrowError = 'throw-error',
   ExitProcess = 'exit-process',
   Retry = 'retry',
}

export type OnInitFailOption = ValueOf<OnInitFail>; // === 'throw-error' | 'exit-process' | 'retry'
```

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -10,6 +10,16 @@ import { EventEmitter } from 'events';
 type PromiseOrDirect<T> = T | Promise<T>;
 type DirectOrCallback<Request, T> = T | ((req: Request) => PromiseOrDirect<T>);
 
+// type ValueOf<T> = T[keyof T];
+
+// export enum OnInitFail {
+//   ThrowError = 'throw-error',
+//   ExitProcess = 'exit-process',
+//   Retry = 'retry',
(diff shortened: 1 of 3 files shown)


benjie:
Please try https://github.com/graphile/postgraphile/pull/1342 and see if it meets your needs.

charlex:
@benjie Yeah, I like that I can put in my own handling function now.

Issue #11, opened 2016-04-28
Approaches to JSON


fils:
So I am new to GraphQL and not a DBA (though I did sleep at a Holiday inn Express)....

I am curious if postgraphql can deal with me putting JSON into Postgres.  I found this reference:  https://www.compose.io/articles/is-postgresql-your-next-json-database/  

Is this something that can be approached by postgraphql in some way?  Or do I need to map my JSON (really JSON-LD using the schema.org vocabulary) into a native shema?   So for example the JSON-LD found at the the page http://schema.org/Person  Could it be mapped into a schema and served?

calebmer:
Yep, it can deal with JSON in PostgreSQL but the caveat to using JSON with PostgreSQL is it isn't typed. You can say that a column is of the JSON type, but you can't specify what types the properties are. This doesn't play nice with GraphQL as a general rule is GraphQL can only query the fields of objects. Currently, because of this PostGraphQL returns the JSON type as a string (however I can think of a workaround to return an object, but it wouldn't be typed).

What I'd recommend (and this is what I do), for every type you use in http://schema.org, create a table and then reference those tables with the standard PostgreSQL mechanisms. http://schema.org schemas are "flat" just like PostgreSQL tables making this the optimal approach as the tables will be typed and PostGraphQL can easily pick up relations.

(Side note: I actually model a lot of my tables with http://schema.org properties 😊)

fils:
I see your point re: typed
I also agree I would rather map to a schema to get the benefits.  So for cases like the scheme.org Person and sections like:

```
"performer": [
    {
      "@type": "MusicGroup",
      "name": "Chicago Symphony Orchestra",
      "sameAs": [
        "http://cso.org/",
        "http://en.wikipedia.org/wiki/Chicago_Symphony_Orchestra"
      ]
    },
    {
      "@type": "Person",
      "image": "/examples/jvanzweden_s.jpg",
      "name": "Jaap van Zweden",
      "sameAs": "http://www.jaapvanzweden.com/"
    }
  ],
```

do the nested elements need to be separate tables?  Or can these be generated in a single table (it's been a long time since I played with making making relational database entities.  

Also, how is the @ symbol dealt with?  My JSON-LD needs it, as you know, but I assume it's not welcome in the schema.  Is there some "label" attribute I set up to associate schema names with json labels during serialization.   (I fear this is a RTF question)  

I'm already trying to resolve out how to alter my data pipeline to target Postgres now.  This is looking very sweet.

[...]

PR #1338, opened 2020-08-26
feat(subscriptions): support `graphql-ws` and all operations over websockets

enisdenjo:
### Story 
Support [`graphql-ws`](https://github.com/enisdenjo/graphql-ws) clients alongside [`subscriptions-transport-ws`](https://github.com/apollographql/subscriptions-transport-ws).

A single **breaking change** of dropping the hook `postgraphile:ws:onOperation` in favour of the new `postgraphile:ws:onSubscription` is introduced by opting-in.

Additionally, support all 3 GraphQL operations by setting the `websocketOperations` option to `all`. Defaults to `subscriptions` which restricts operations to subscriptions only.

### TODOs:
- [x] WebSocket transport support for `query` & `mutation`
- [x] Add support for live queries
- [x] Add support for silent auto-reconnect ([Commit c6f7872](https://github.com/enisdenjo/graphql-ws/commit/c6f7872126300befcc47e8e46e82342c2924f453)) 
- [x] Expose client listeners for connected, connecting and disconnected ([Commit 627775b](https://github.com/enisdenjo/graphql-ws/commit/627775b8e1aca8f359607020ff2c3bcc37b50787))
- [x] Implement keep-alive ([PR #11](https://github.com/enisdenjo/graphql-ws/pull/11))


postgraphiql/src/components/PostGraphiQL.js
```diff
@@ -350,9 +352,9 @@ class PostGraphiQL extends React.PureComponent {
             });
           };
 
-          const subscription = this.subscriptionsClient.request(graphQLParams).subscribe(observer);
+          const unsubscribe = this.subscriptionsClient.subscribe(graphQLParams, observer);
```
benjie, line 355: I'd preserve the existing API here; even if `unsubscribe` is the only method this exposes currently, there may be the need to add additional helpers in future, and adding them as properties of an unsubscribe function is kinda ugly.

src/postgraphile/http/subscriptions.ts
```diff
@@ -222,47 +215,39 @@ export async function enhanceHttpServerWithSubscriptions<
...
-
           return response;
         };
-        // onOperation is only called once per params object, so there's no race condition here
-        // eslint-disable-next-line require-atomic-updates
-        params.formatResponse = formatResponse;
+
         const hookedParams = pluginHook
-          ? pluginHook('postgraphile:ws:onOperation', params, {
+          ? pluginHook('postgraphile:ws:onSubscribe', args, {
```
benjie, line 244: Please don't rename the hook or the options unless there's definitely a breaking change here and the old one needs to be removed (in which case we need to have the old one throw when someone else hooks it) as this is a public API.

src/postgraphile/http/subscriptions.ts
```diff
@@ -271,46 +256,35 @@ export async function enhanceHttpServerWithSubscriptions<
...
-       *
-       * The subscriptions-transport-ws client times out by default 30s after last keepalive:
-       *   https://github.com/apollographql/subscriptions-transport-ws/blob/52758bfba6190169a28078ecbafd2e457a2ff7a8/src/defaults.ts#L1
-       *
-       * GraphQL Playground times out after 20s:
-       *   https://github.com/prisma/graphql-playground/blob/fa91e1b6d0488e6b5563d8b472682fe728ee0431/packages/graphql-playground-react/src/state/sessions/fetchingSagas.ts#L81
-       *
-       * Pick a number under these ceilings.
-       */
-      keepAlive: 15000,
```
benjie, line 311: Where's the keepAlive configured now?

postgraphiql/src/components/PostGraphiQL.js
enisdenjo: What API are you talking about exactly? The `subscribe` method intentionally returns **just** a cleanup function because it is intended to be a "subscriber" of an [Observable](https://github.com/tc39/proposal-observable#observable).

```ts
const observable = Observable.create(
  function subscriberFunction(observer: SubscriptionObserver) {
    // Subscriber func returns a cleanup function which will cancel the event stream/dispose of the subscription
    return subscriptionsClient.subscribe(operation, observer);
  }
);
```

src/postgraphile/http/subscriptions.ts
enisdenjo: I changed the name intentionally because the legacy `onOperation` func is different from the current `onSubscribe` func. Using the new lib is a breaking change anyway, so why not clean up?

src/postgraphile/http/subscriptions.ts
enisdenjo: I have envisioned using the [WebSocket Ping and Pong](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers#Pings_and_Pongs_The_Heartbeat_of_WebSockets) for keeping the connection alive. Not only that, but my idea is to abstract away the need of managing the `keepAlive` completely and let a baked-in implementation handle that for the user.

postgraphiql/src/components/PostGraphiQL.js
benjie: I thought Observable's returned a `Subscription` object that has both an `unsubscribe()` method and other details (e.g. the `closed` getter):

```
interface Subscription {

    // Cancels the subscription
    unsubscribe() : void;

    // A boolean value indicating whether the subscription is closed
    get closed() : Boolean;
}
```

You seem to be returning a function directly? I'm just guessing.

src/postgraphile/http/subscriptions.ts
benjie: Okay. Be aware we definitely cannot merge this into v4 then; you might be better off doing it as a PR against the v5 branch (which is restructured as a monorepo).

src/postgraphile/http/subscriptions.ts
benjie: Different clients/networks have different needs here; so I think it definitely needs to be configurable. Using the built in websocket ping/pong makes sense to me, but I'd have to test it to make sure it works for Heroku/etc.

postgraphiql/src/components/PostGraphiQL.js
enisdenjo: Yes, the subscriber returns a cleanup function directly. Its the standard; RxJS, Relay, Apollo, CoreJS and even the [Observable proposal](https://github.com/tc39/proposal-observable) have the following signature:
```ts
function listen(element, eventName) {
    return new Observable(observer => {
        // Create an event handler which sends data to the sink
        let handler = event => observer.next(event);

        // Attach the event handler
        element.addEventListener(eventName, handler, true);

        // Return a cleanup function which will cancel the event stream
        return () => {
            // Detach the event handler from the element
            element.removeEventListener(eventName, handler, true);
        };
    });
}
```

*But*, what you might be confused with is expecting the `subscribe` method **to return an Observable** (which has the `unsubscribe` method). This is **not the case**, it is actually the subscriber (the implementing/constructor part of an Observable). You essentially make an Observable **using** the `subscribe`. This way the lib is completely agnostic to any observable pattern / streaming implementation the user (or a GQL client) might want to use.

Here is an example on how to make an Observable:
```ts
import { Observable } from 'relay-runtime';
// or
import { Observable } from '@apollo/client';
// or
import { Observable } from 'rxjs';
// or
import Observable from 'zen-observable';

function request(operation) {
  return Observable.create((observer) => {
    // subscriber implementation
    return subscriptionsClient.subscribe(operation, observer);
  });
}

const subscription = request({ ... } as Operation);
subscription.unsubscribe(); // ✅
```

src/postgraphile/http/subscriptions.ts
enisdenjo: Its rebased on v4 because I am using this fork in production. I also see this breaking change better fitted for V5. I'll make a PR there too 👍 .

On the side note, how stable is V5? Would you recommend using it in production? Asking this because if I move the PR there too, would it make sense to test it out in production too?

src/postgraphile/http/subscriptions.ts
enisdenjo: Fair point, I'll make the `keepAlive` configurable. [PR #10 at the lib repo](https://github.com/enisdenjo/graphql-transport-ws/issues/10).

src/postgraphile/http/subscriptions.ts
benjie: Don't try and use V5 yet ;)

pyramation:
this looks awesome! following :)

postgraphiql/src/components/PostGraphiQL.js
```diff
@@ -5,7 +5,7 @@ import GraphiQLExplorer from 'graphiql-explorer';
 import StorageAPI from 'graphiql/dist/utility/StorageAPI';
 import './postgraphiql.css';
 import { buildClientSchema, introspectionQuery, isType, GraphQLObjectType } from 'graphql';
-import { createClient } from '@enisdenjo/graphql-transport-ws';
+import { createClient } from 'graphql-transport-ws';
```
benjie, line 8: :raised_hands:

asafyish:
What's the advantage of using graphql-transport-ws instead of subscriptions-transport-ws ?
subscriptions-transport-ws has over a million weekly downloads, while graphql-transport-ws has just over 100.

enisdenjo:
Hey @asafyish, good question! In short - the `subscriptions-transport-ws` is stale, not actively maintained, "Work in progress" for 4 years and has quite an amount of bugs and security implications. (Check out a rough overview of them [here](https://github.com/enisdenjo/graphql-transport-ws/issues/3))

On the other hand, [`graphql-transport-ws`](https://github.com/enisdenjo/graphql-transport-ws) is a complete fresh, zero-dependency, plug-n-play, rewrite of the transport using modern JS. Not only is it a rewrite of the lib itself, but it also offers a redesigned [GraphQL over WebSocket Protocol](https://github.com/enisdenjo/graphql-transport-ws/blob/master/PROTOCOL.md) which aims to be standardised with the help of the foundation's [GraphQL over HTTP work group](https://github.com/graphql/graphql-over-http).

src/postgraphile/http/subscriptions.ts
enisdenjo: Done in [PR #11 at the lib repo](https://github.com/enisdenjo/graphql-transport-ws/pull/11). Will resolve this conversation once I have released a new version and updated this fork.

enisdenjo:
Hey @benjie,

As of [graphql-transport-ws@v1.3.0](https://github.com/enisdenjo/graphql-transport-ws/releases/tag/v1.3.0) the lib is feature full and ready for the world! I've been using it in production for a whole week now - except for the occasional closure caused by network problems, I ran in **no problems at all**.

Will make a PR for v5 later on. However, besides resolving conflicts, rebasing, dropping the version bumps and deciding on how to handle breaking changes; this PR is ready too!

benjie:
Awesome :raised_hands: We're definitely going to have to work to minimize breaking changes; but fortunately the server plugin API has this note at the very top:

> Stability: experimental, may change in semver minor versions.
> -- https://www.graphile.org/postgraphile/plugins/

So basically I'm happy to break the server plugin API if it's absolutely necessary, so long as that breakage is **super obvious**. So we need to make it so people calling the older hooks causes a panic and the system refuses to start. Do you want to have a go at doing this?

benjie:
(Before anything else, please rebase. I'm on the verge of releasing 4.9.0 so I don't think there's any more major changes coming in soon.)

enisdenjo:
Thats great! We should be leaving the experimental phase soon 😁. From the looks of it, there are 2 breaking changes:
- `subscriptions-transport-ws` is not supported anymore. The client counterpart **must** use this lib!
- [The hook name change from `postgraphile:ws:onOperation` to `postgraphile:ws:onSubscription`](https://github.com/graphile/postgraphile/pull/1338/files#diff-88720c8c3a816fa2d1f27f111f7e1cf4L57-R56)

enisdenjo:
Ready 👍

benjie:
The first of those:

> subscriptions-transport-ws is not supported anymore. The client counterpart must use this lib!

This is a major breaking change. We can't do it on a semver minor update. Protocol-wise, what specifically has broken? Can we use both libraries and use a header/sniff the traffic to pipe it into one or the other implementation? Maybe we can change the `--subscriptions` flag to accept something like `--subscriptions=v2` so they can opt in to the new library. Maybe we give them warnings about security/etc if they don't.

I'd like to sort this out conceptually before I review the specific code.

enisdenjo:
I like your idea of using the flag. I see it something like this:
```ts
type SubscriptionsFlag = 'v0' | 'v1' | boolean;
// true = 'v0' 👈  to reduce breaking changes. For all those who had this flag set to true - nothing will change.
```
I am thinking about `v0` and `v1` because `subscriptions-transport-ws` is still on the `0th` major version, while the `graphql-transport-ws` is on the `1st`.

Also, if you choose `v1`, the plugin hook would change from `postgraphile:ws:onOperation` to `postgraphile:ws:onSubscription`.

What do you say?

benjie:
This seems very reasonable. Could you attempt to integrate this in? Also I don't think we actually have tests for websocket connections right now, but with two separate websocket implementations in play we're going to need them - are you interested in taking this on or do you need me to attempt it? I'd like to see tests for the existing implementation merged separately (with this PR in mind) and then see how this PR changes those tests.

benjie:
Actually; can you add a `--websockets=none/v0/v1` flag, since the `--subscriptions` flag is already quite overloaded (it's treated as a feature flag that enables a bunch of stuff inside Graphile Engine that's arguably a breaking change but I don't expect anyone to notice). It should default to v0 if `--subscriptions` is specified, and `none` otherwise. We should also keep in mind the `websocketOperations` option proposed in #1233.

enisdenjo:
I'll do the `websocket` flag implementation for sure, no worries! Will try doing that today already.

Regarding the tests, I'd be happy to contribute with test cases; but, I would be very grateful if you could lay down the ground work? Just some basic subscriptions test setup to help me understand more about PostGraphile's approach to testing and to allow the tests to fit with the rest of the codebase.

My lib already fully supports all operations, so I can integrate that in this PR too.

I also oversaw the discussion about auth in #1233 - so I'd just like to add my 2 cents in with how I approach this in my production environments. Besides relying on CORS to protect against connections originating from unknown sources I currently is authenticate the requester on **every** subscription request and kick him out if I sense something fishy. However, the `connectionParams` I check against are set-up **once** (during the connection initialisation phase) and persisted during subsequent requests - **but**, is still think this is completely OK. For example: if you use JWT to authenticate and the JWT has an expiration date, it will expire for the sockets the same way it expires for the usual HTTP req/res. Since the lib supports silent auto-reconnects; further improvements could also potentially be gracefully closing the sockets with clients and communicating a "re-auth" challenge through the close event, something like: `4511: Authentication Required` - but this might cause losing some emitted events in critical moments (can be avoided by using `@pgSubscription(intialEvent: ...)`).

benjie:
Sounds good; hopefully I can take a look at this on Tuesday :+1:

enisdenjo:
Rebased.

enisdenjo:
Hey @benjie, I have addressed all raised concerns. Renamed the name and updated the PR description to reflect that. Would be awesome if you can check it out!

enisdenjo:
👋  @benjie, me again. I guess you are very busy and I don't want to be annoying, I just wanted to ping and check about this PR. I am looking forward to using the official release instead of a packaged fork 😁 .

enisdenjo:
Rebased.

jwrascoe:
@enisdenjo I was just speaking to @benjie and was looking to see how to support changing JWT tokens on the fly for a WS connection... basically check it every time the onOperation is called in subscriptions.ts
https://github.com/graphile/postgraphile/blob/1da44d37c18afb3627cd8e9fc41b275514faee7a/src/postgraphile/http/subscriptions.ts#L227

I would like to use this obviously to change roles when the query hits.

Have you considered this at all?

enisdenjo:
Hey @jwrascoe,

The `onOperation` equivalent in this case is the `onSubscribe` callback (and the matching `postgraphile:ws:onSubscription is` plugin hook):

https://github.com/enisdenjo/postgraphile/blob/36e3561f0ea17cbd74932dac98898bd01dbf218e/src/postgraphile/http/subscriptions.ts#L387-L416

By "changing JWT tokens on the fly" you mean changing the actual token in the original `connectionParams` request? You could do that on every subscribe request (operation) like this:

```ts
...
async onSubscribe(ctx, message, args) {
  // ctx is the context of the connection and the connectionParams are from the initialisation message
  ctx.connectionParams.authorization = yourBearerToken(request, message, args);
  // request is the initial connection HTTP request from the client
  ctx.request.headers.authorization = ctx.connectionParams.authorization;
  ...
}
...
```

Out of curiousity, why would you change the clients JWT token to switch roles?

jwrascoe:
@enisdenjo Interesting... let me see if I can figure out how to implement your suggestion.

Without a token, I have the default access set to only allowing the request of a token... once the user provides credentials, I look it up and then decide which role they need to be so I can enforce security..

The reason for changing the token is not to change roles (only if they log in and out as someone else) but rather to secure the data a bit more.

In the event that the JWT token was caught somehow in a packet trace, I only wanted it to be good for a few minutes.  The way I wrote my client it looks at the TTL in the JWT and renews it (if its able to) for continued access.

I could  close and restart the WS connection, but if I had outstanding subscriptions I would have to re-issue them and wanted to just supply the replacement(s) token to keep the connection alive with the subscriptions intact.

Maybe there is a better way? Or im just being paranoid?

enisdenjo:
I'd say you should always be aware of the exact details a client provides (or has provided) you; so, personally I wouldn't change the token on the fly **in the server** ever...

I would recommend relying on the library and the `@pgSubscription`'s `initialEvent` argument. Keep checking the JWT token validity on every subscription request using the subscription directive topic builder: `@pgSubscription(topic: ${embed(checkAuthAndBuildTopic)})`. If the checks fails because of whatever, terminate the WS connection completely, the client will silently try to reconnect again and all "hot" subscribers on the client side will re-subscribe. Using the `initialEvent` argument, you can issue an event on every subscription and therefore supplement any data that the client might have missed during the time it was trying to reconnect. Furthermore, the client has a separate process of requesting a new token on expiry. So, if the server kicks the client off, the reconnection attempts should already be using the new, refreshed, token.

jwrascoe:
@enisdenjo ok, thanks for the info... I'll see if I can come up with something that fits my use case.

jwrascoe:
@enisdenjo, @benjie ... I actually found a way to do it that required no code change... and solves my use case.

When I connect in my client code, I can specify a call back on the connection params... 
```
  const ws = new WebSocketLink({
    uri: environment.wsUrl,
    webSocketImpl: WebSocket,
    options: {
      minTimeout: 3000, 
      lazy: false,
      reconnect: true,
      connectionParams: () => {
        const token = coreService.appContext$.getValue().authToken;
        return token ? { Authorization: `Bearer ${token}` } : {};
      }
    }
  });
```
 Then whenever I ask for a new JWT token from the server I can simply do a soft close.. which forces a reconnect, runs the callback and will use the latest token.. and keeps the subscriptions outstanding alive.

```
ws.subscriptionClient.close(false, false);
```

@enisdenjo it was your comment that changed my direction on how to solve my problem....

"terminate the WS connection completely, the client will silently try to reconnect again and all "hot" subscribers on the client side will re-subscribe" 

Thank you both for the help, and I look forward to working with this PR when its ready... I will give it a good workout!

enisdenjo:
Rebased and updated lib to latest version.

enisdenjo:
Rebased.

src/postgraphile/http/subscriptions.ts
```diff
@@ -186,134 +184,278 @@ export async function enhanceHttpServerWithSubscriptions<
...
-          throw new Error('No request!');
-        }
-        const normalizedConnectionParams = lowerCaseKeys(connectionParams);
-        request['connectionParams'] = connectionParams;
-        request['normalizedConnectionParams'] = normalizedConnectionParams;
-        socket['__postgraphileReq'] = request;
-        if (!request.headers.authorization && normalizedConnectionParams['authorization']) {
+  switch (websockets) {
+    case 'none':
+      break;
```
benjie, line 189: This should check much earlier, and throw if `'none'`.

src/postgraphile/http/subscriptions.ts
```diff
@@ -186,134 +184,278 @@ export async function enhanceHttpServerWithSubscriptions<
...
+              ) =>
+                liveSubscribe({
+                  schema,
+                  document,
+                  rootValue,
+                  contextValue,
+                  variableValues,
+                  operationName,
+                  fieldResolver,
+                })
```
benjie, line 219: Ah; you've moved the logic here from before. Please put it back where it was for parity with https://github.com/graphql/graphql-js/blob/master/src/execution/execute.js

src/postgraphile/http/subscriptions.ts
```diff
@@ -186,134 +184,278 @@ export async function enhanceHttpServerWithSubscriptions<
...
+              // (there is no result to add the meta to)
+            });
+            const validationErrors = validate(
+              hookedArgs.schema,
+              hookedArgs.document,
+              // should the validation rules be applied
+              // in addition to the graphql defaults (as before)
+              // or should they be used exclusively?
+              validationRules.length ? validationRules : undefined,
             );
```
benjie, line 429: The query should already have been validated at this point, right?

If so, we only want to apply _additional_ validation rules here. So there should be an `if (moreValidationRules.length) {` check around this.

If not, we need to be much more careful with the validation here - at the moment adding a single validation rule will cause all the built in validations not to be used.

src/postgraphile/http/subscriptions.ts
```diff
@@ -186,134 +184,278 @@ export async function enhanceHttpServerWithSubscriptions<
...
+              // operation execution errors
+              const { req, res } = await reqResFromSocket(ctx.socket);
+              result.errors = handleErrors(result.errors, req, res);
+              return result;
+            }
+          },
+          onComplete({ socket }, msg) {
+            releaseContextForSocketAndOpId(socket, msg.id);
+          },
+          // default keepAlive fine? (12 seconds)
```
benjie, line 452: Seems fine.

src/postgraphile/http/subscriptions.ts
```diff
@@ -186,134 +184,278 @@ export async function enhanceHttpServerWithSubscriptions<
...
+        },
+        wss,
+      );
+      break;
+    }
+    default:
+      throw new Error(`Invalid websockets argument ${websockets}`);
```
benjie, line 459: We should throw this much earlier.

tsconfig.json
```diff
@@ -18,6 +18,7 @@
     "noFallthroughCasesInSwitch": true,
     "noUnusedParameters": false,
     "noUnusedLocals": false,
+    "skipLibCheck": true,
```
benjie, line 21: Why do we need this?

benjie (changes requested):
This is looking good! I think there's a few more changes in this PR than strictly necessary, I'd prefer that we keep the diff as small as possible to make it clearer what changes were actually required for this versus simply refactoring (any refactoring can be discussed in a follow-up PRs). There's also a couple breaking changes to public interfaces that need to be undone and solved a slightly different way. The main changes in subscriptions.ts will need some review too; at the moment the logic used differs between the Apollo version and your version; I need to understand why and make sure that we're not overlooking anything.

src/postgraphile/http/subscriptions.ts
enisdenjo: When supplying custom `onSubscribe` execution arguments, [you're trusted to prepare the execution arguments](https://github.com/enisdenjo/graphql-ws/blob/master/docs/interfaces/_server_.serveroptions.md#onsubscribe) for direct usage. This is a conscious decision because it allows you to apply optimizations (like executing some rules exclusively or trusting the client implicitly) or simply have control of when the validation happens when you introduce a custom implementation.

> The query should already have been validated at this point, right?

So no, at this point the query is untouched. As you may notice, we're using the fields from the raw message.

> If not, we need to be much more careful with the validation here - at the moment adding a single validation rule will cause all the built in validations not to be used.

Yeah, which is why I asked. Didn't really know how strict the hook is. [I now do the regular validation before performing optional additional ones.](https://github.com/graphile/postgraphile/pull/1338/commits/4639fcc1259ae05ab9b7f91bc1f4a86c4eb2d810)

src/interfaces.ts
```diff
@@ -69,6 +69,10 @@ export interface PostGraphileOptions<
   ownerConnectionString?: string;
   // Enable GraphQL websocket transport support for subscriptions (you still need a subscriptions plugin currently)
   subscriptions?: boolean;
+  // Choose the version of the websocket transport library. Requires `subscriptions` to be enabled. Defaults to `v0`
+  websockets?: 'none' | 'v0' | 'v1';
```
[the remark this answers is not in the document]
enisdenjo: I think it makes sense too. I changed the code accordingly.

tsconfig.json
enisdenjo: Because graphql-ws previously used the `WebSocket` and `CloseEvent` types from the global scope which does not exist outside the "dom" lib. However, this is not the case anymore as of [v1.9.3](https://github.com/enisdenjo/graphql-ws/releases/tag/v1.9.3), so I removed it and the config is now unchanged.

src/interfaces.ts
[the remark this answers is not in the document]
enisdenjo: I reckon that `enhanceHttpServerWithSubscriptions` should then be renamed to `enhanceHttpServerWithWebSocket` to accommodate the change?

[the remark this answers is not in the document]
benjie: I feel like that should be plural: `enhanceHttpServerWithWebSockets`, but yes; that's a better name. Please keep the old name as just a reference to the new function; we have a section in the code where we've done that in the past: 

https://github.com/graphile/postgraphile/blob/7dde41da9204bbd381c20e5783439118650fccd2/src/index.ts#L35-L39

src/postgraphile/http/subscriptions.ts
benjie: Got it. So there's a bit more work you need to do; your validation needs to use the static rules, like here:

https://github.com/graphile/postgraphile/blob/7dde41da9204bbd381c20e5783439118650fccd2/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts#L381

which come from:

https://github.com/graphile/postgraphile/blob/7dde41da9204bbd381c20e5783439118650fccd2/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts#L324-L326

(Note that they're based on GraphQL's default list of validation rules, but can then be manipulated via server plugins.)

(Further, note these rules only need to be calculated once, e.g. within the `enhanceHttpServerWithWebSockets` call. They do not need to be recalculated for each request.)

Then later comes the part I think you've now handled; i.e.

https://github.com/graphile/postgraphile/blob/7dde41da9204bbd381c20e5783439118650fccd2/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts#L820-L838

tsconfig.json
benjie: Awesome; thanks :raised_hands:

src/postgraphile/http/subscriptions.ts
enisdenjo: Awesome! [The `staticValidationRules` were already available for me, I just plugged them in.](https://github.com/graphile/postgraphile/pull/1338/commits/f78a4cc4d4ca55bd8d2701fab74d1d8d8826bf8e)

src/postgraphile/http/subscriptions.ts
```diff
@@ -186,134 +191,262 @@ export async function enhanceHttpServerWithSubscriptions<
...
+
+            // Override schema (for --watch)
+            const schema = await getGraphQLSchema();
+
+            const { payload } = msg;
+            const args: ExecutionArgs = {
+              schema,
+              contextValue: context,
+              operationName: payload.operationName,
+              document: parse(payload.query),
```
benjie, line 381: In the case of persisted operations, `payload.query` may not be present, it might be `payload.documentId` or something else for example. I'm not sure, but I think that calling `parse` here might throw an error if there is no `query`? We want effectively `document: (() => {try{return parse(payload.query)} catch { return null }})()`

src/postgraphile/http/subscriptions.ts
```diff
@@ -186,134 +191,262 @@ export async function enhanceHttpServerWithSubscriptions<
...
+              variableValues: payload.variables,
+            };
+
+            // for supplying custom execution arguments
+            const hookedArgs = pluginHook
+              ? pluginHook('postgraphile:ws:onSubscribe', args, {
+                  ...ctx,
```
benjie, line 388: Lets pass `ctx` directly rather than splatting it. That allows us to add things in future without worrying that they'll override things that `ctx` contains.
```suggestion
                  ctx,
```

src/postgraphile/http/subscriptions.ts
```diff
@@ -186,134 +191,262 @@ export async function enhanceHttpServerWithSubscriptions<
...
+                  message: msg,
+                  options,
+                })
+              : args;
+
+            // when supplying custom execution args from the
+            // onSubscribe, you're trusted to do the validation
+            let validationErrors = validate(
```
benjie, line 396:
```suggestion
            const validationErrors = validate(
```

src/postgraphile/http/subscriptions.ts
```diff
@@ -186,134 +191,262 @@ export async function enhanceHttpServerWithSubscriptions<
...
+              variables: hookedArgs.variableValues,
+              operationName: hookedArgs.operationName,
+              // no meta because validation errors returned from here will be
+              // served through the error message. it contains just the GraphQLErrors
+              // (there is no result to add the meta to)
+            });
+            if (validationRules.length) {
+              validationErrors = validate(hookedArgs.schema, hookedArgs.document, validationRules);
+              if (validationErrors.length) {
+                return validationErrors;
```
benjie, line 422: I prefer that `const` is used where possible; it makes the code easier for me to follow because I don't have to worry about rebinding.

```suggestion
            const moreValidationRules = pluginHook('postgraphile:validationRules', [], {
              options,
              req,
              res,
              variables: hookedArgs.variableValues,
              operationName: hookedArgs.operationName,
              // no meta because validation errors returned from here will be
              // served through the error message. it contains just the GraphQLErrors
              // (there is no result to add the meta to)
            });
            if (moreValidationRules.length) {
              const moreValidationErrors = validate(hookedArgs.schema, hookedArgs.document, moreValidationRules);
              if (moreValidationErrors.length) {
                return moreValidationErrors;
```

src/postgraphile/http/subscriptions.ts
```diff
@@ -186,134 +191,262 @@ export async function enhanceHttpServerWithSubscriptions<
...
+          async onNext(ctx, _msg, _args, result) {
+            if (result.errors) {
+              // operation execution errors
+              const { req, res } = await reqResFromSocket(ctx.socket);
+              result.errors = handleErrors(result.errors, req, res);
+              return result;
+            }
+          },
+          onComplete({ socket }, msg) {
+            releaseContextForSocketAndOpId(socket, msg.id);
```
benjie, line 442: We need this to happen when the socket disconnects unexpectedly too, I think. E.g. if the client closes the connection/times out/etc.

benjie (changes requested):
This is really close! From now on, please do not rebase so I can review only the changed code. (I'll be squash-merging at the end, so don't worry about having a tidy git history.)

A few more questions and change requests below, but we're basically there I think :raised_hands:

postgraphiql/src/components/PostGraphiQL.js
```diff
@@ -102,63 +103,92 @@ class PostGraphiQL extends React.PureComponent {
...
+            connecting: () => {
+              this.setState({ socketStatus: 'connecting' });
+            },
+            connected: () => {
+              this.setState({ socketStatus: 'connected', error: null });
+            },
+            closed: closeEvent => {
+              this.setState({
+                socketStatus: 'closed',
+                error: new Error(`Socket closed with ${closeEvent.code} ${closeEvent.reason}`),
```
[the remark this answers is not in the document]
enisdenjo: Yes. Most probably with code 1006 and no reason, but the props will be there.

With `graphql-ws` the server can emit errors in 2 ways:
- `GraphQLError`s sent through the error message (return them from `onSubscribe`)
- Fatal error that is represented as a [CloseEvent](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent) with the relevant code and the original error message in the `reason` (except for internal errors which, in production environments, will always be `Internal Error`).

src/postgraphile/http/subscriptions.ts
enisdenjo: When using persisted queries, you wont be _just_ parsing the payload's query. The preparation should happen beforehand, like checking the `documentId` field and supplying your own document during assembly. Would you agree?

You can check the ["Server and client usage with persisted queries" recipe](https://github.com/enisdenjo/graphql-ws#recipes) on how I envisioned this.

src/postgraphile/http/subscriptions.ts
```diff
@@ -186,134 +191,262 @@ export async function enhanceHttpServerWithSubscriptions<
...
+      createServer(
+        {
+          schema,
+          execute:
+            subscriptionServerOptions?.operations === 'all'
+              ? execute
+              : () => {
+                  throw new Error('Only subscriptions are allowed over WebSocket transport');
+                },
+          subscribe: options.live ? args => liveSubscribe(args, args.document) : graphqlSubscribe,
```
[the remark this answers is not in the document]
enisdenjo: [The `liveSubscribe` typescript requires it 😄.](https://github.com/enisdenjo/postgraphile/blob/776c884f89daa7a184820c24ef3da2bda84cbc12/src/postgraphile/http/liveSubscribe.ts#L28) We either change the types there (do some more refactoring) or we do a `@ts-expect-error` here.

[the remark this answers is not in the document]
benjie: Add a `?` to the document arg; I think that's the issue?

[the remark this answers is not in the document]
enisdenjo: Easy, just thinking about pushing less changes. 😄

src/postgraphile/http/subscriptions.ts
benjie: In that example you override it in the `onSubscribe` hook, which is what I'd like to do too. In the existing implementation we use the `postgraphile:ws:onOperation` hook for this, so I'd expect to achieve it through the `postgraphile:ws:onSubscribe` hook a couple lines below here - passing the args (sans valid query) and message (with `documentId` or whatever) and getting back in `hookedArgs` the valid args with new `query` property set.

benjie: I expect it'd look very much like this: https://github.com/graphile/persisted-operations/blob/ee2e404ccc5e4f7f7ffe65215a97ac5e6e0fc08e/src/index.ts#L256-L263

enisdenjo: Didn't saw that there! I'll make sure to add support for persisted queries using the `onSubscribe` plugin hook.

src/postgraphile/http/subscriptions.ts
```diff
@@ -53,6 +56,8 @@ export async function enhanceHttpServerWithSubscriptions<
   subscriptionServerOptions?: {
     keepAlive?: number;
     graphqlRoute?: string;
+    websockets: PostGraphileOptions['websockets'];
+    operations: PostGraphileOptions['websocketOperations'];
```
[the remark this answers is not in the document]
enisdenjo: Something like this https://github.com/graphile/postgraphile/pull/1338/commits/d44a8f245607db48474690d7eb14f74bff058095?

postgraphiql/src/components/PostGraphiQL.js
```diff
@@ -350,7 +390,13 @@ class PostGraphiQL extends React.PureComponent {
...
           };
 
-          const subscription = this.subscriptionsClient.request(graphQLParams).subscribe(observer);
+          let subscription;
+          if (POSTGRAPHILE_CONFIG.websockets === 'v0') {
+            subscription = this.subscriptionsClient.request(graphQLParams).subscribe(observer);
+          } else if (POSTGRAPHILE_CONFIG.websockets === 'v1') {
+            const unsubscribe = this.subscriptionsClient.subscribe(graphQLParams, observer);
+            subscription = { unsubscribe };
+          }
```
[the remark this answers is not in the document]
enisdenjo: This what you had in mind https://github.com/graphile/postgraphile/pull/1338/commits/5be16ccef8314d08b462cdeb977d6c17cb5ce826?

src/postgraphile/http/subscriptions.ts
enisdenjo: With https://github.com/graphile/postgraphile/pull/1338/commits/1074b013657302773054850a784578e3c2d8f4f5, you should be able to do something like:
```ts
"postgraphile:ws:onSubscribe" = (args, { message, options }) => {
  args.query = persistedOperationFromPayload(
    message.payload, // graphql-ws subscribe message payload
    options
  );
  return args;
};
```

src/postgraphile/http/subscriptions.ts
enisdenjo: Ah yes; because of the way `graphql-ws` is built, `onComplete` will be called on abrupt closures (from any side) too. [I had that confirmed with this test, just in case.](https://github.com/enisdenjo/graphql-ws/commit/3f5c42e3deb7d014258417d20df29ef9e1fb1ed7)

The only addition I had is https://github.com/graphile/postgraphile/pull/1338/commits/6b3c5a84e2d6e497da77d110c9b3bedc35f1390a. We should be tight now, the only 2 exits for an observable are `error` and `complete`.

postgraphiql/src/components/PostGraphiQL.js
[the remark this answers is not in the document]
benjie: Yep :+1:

src/postgraphile/http/subscriptions.ts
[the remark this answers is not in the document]
benjie: Exactly :+1:

src/postgraphile/http/subscriptions.ts
```diff
@@ -186,134 +191,262 @@ export async function enhanceHttpServerWithSubscriptions<
...
-           * Referer / etc)
+           * GraphQL Playground times out after 20s:
+           *   https://github.com/prisma/graphql-playground/blob/fa91e1b6d0488e6b5563d8b472682fe728ee0431/packages/graphql-playground-react/src/state/sessions/fetchingSagas.ts#L81
+           *
+           * Pick a number under these ceilings.
            */
-          request.headers.authorization = String(normalizedConnectionParams['authorization']);
-        }
+          keepAlive: 15000,
+        },
```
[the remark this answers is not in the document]
benjie: I think you added `...subscriptionServerOptions` to v1 instead of v0?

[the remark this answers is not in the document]
enisdenjo: Oops, misplaced it 😅. Fixed in https://github.com/graphile/postgraphile/pull/1338/commits/821ba65be2b39d7d5333312221771239cebe1bd7.

src/postgraphile/http/subscriptions.ts
benjie: That looks reasonable. It prevents passing the query identifier via the `query` parameter but I think that'd be a bit of an undesirable thing to do anyway. Apollo Client uses the `extensions` property on the message, and Relay uses a `documentId` field, and I've seen it just done via an `id` field before, so I think we're fine on this.

benjie:
Note to self: add

```
"postgraphile:ws:onSubscribe" = (args, { message, options }) => {
  args.query = persistedOperationFromPayload(
    message.payload, // graphql-ws subscribe message payload
    options
  );
  return args;
};
```

To `@graphile/persisted-operations`

src/postgraphile/http/subscriptions.ts
benjie: Great :raised_hands:

benjie:
@enisdenjo Regarding `onComplete`, you may want to update the docs to reflect that it also handles abrupt completion: https://github.com/enisdenjo/graphql-ws/blob/master/docs/interfaces/_server_.serveroptions.md#oncomplete

benjie (approved):
Amazing; this looks solid! Thanks so much for your work on this :raised_hands:  Any further changes will be from me as I test it/tweak docs if necessary/etc.

enisdenjo:
Brilliant! Thanks for welcoming this PR and for your great insights.

I guess the next part would be to refresh the docs too?

benjie:
We also need an upgraded version of graphql-ws that supports older GraphQL versions (PostGraphile needs support for GraphQL < 15).

enisdenjo:
The big [`graphql-ws@v1.11.0`](https://github.com/enisdenjo/graphql-ws/releases/tag/v1.11.0) has landed!

Node 10 is now the min supported node engine and the peer `graphql` dependency can be from v0.11.0 onwards.

benjie:
Have merged in the latest GraphiQL. There was a recent change that broke v0 websockets, but I've fixed that. The new websockets seem to work smoothly.

There's one significant concern I have: when you connect to a v1 sockets server using a v0 sockets client the connection is just terminated and re-attempted indefinitely but no error is output. It feels like the server shouldn't just close the connection but should also send a message informing the client _why_ the connection was closed. Obviously I knew what I had done in this circumstance as I did it deliberately, but people less familiar with GraphQL websockets are going to have a really hard time debugging this without more information.

There's also a change in behaviour when you're editing the headers: in v0 the existing subscription is re-subscribed when you reconnect, with v1 the existing subscription is terminated. I prefer the former behaviour in GraphiQL as it took a little while to realise that the subscription had stopped just because I'd added and then deleted one character in the headers editor.

enisdenjo:
I am happy to hear that its is working well!

`graphql-ws` server closes the connection with v0 immediately because its faced with an invalid subprotocol (v0 uses a different one). [It indeed says why](https://github.com/enisdenjo/graphql-ws/blob/master/src/server.ts#L371-L379) - communicating an unsupported protocol using the [spec's `1002: Protocol Error` close event](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent). Maybe the problem is with error reporting in the v0 client?

The reason they terminate is because the client is disposed. There is no imperative method which reconnects with new connection parameters in graphql-ws. My reasoning is that I see the client as "the" socket, it exists as long as the socket does. Likewise, it dies when the socket does too. Since, [the Protocol does not allow multiple `ConnectionInit` messages at any point in time within the socket](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#connectioninit), the socket has to be disposed to reconnect with new parameters - this behaviour feels very appropriate and simple. Whats your take on this? I will think about a possible solution on this and come back with my findings.

benjie:
> graphql-ws server closes the connection with v0 immediately because its faced with an invalid subprotocol (v0 uses a different one). It indeed says why - communicating an unsupported protocol using the spec's 1002: Protocol Error close event.

Can confirm; turns out Chrome devtools doesn't expose this but Firefox's does. :+1: 

> the socket has to be disposed to reconnect with new parameters - this behaviour feels very appropriate and simple. Whats your take on this?

Seems fine; I only care about this for GraphiQL - can you add a line to make it restart the subscription if one was active where you recreate the socket?

https://github.com/graphile/postgraphile/pull/1338/files#diff-cae0f37e71b3354a66efda9007b7a693dd1627df304107e84a732bd13896da0dR543-R549

benjie:
Currently thinking it might make sense to support both v0 and v1 protocol at the same time to make moving from v0 to v1 smoother.

benjie (approved):

enisdenjo:
Hey hey, great news!

[The server will now accept both `subscriptions-transport-ws` and `graphql-ws` clients!](https://github.com/enisdenjo/postgraphile/blob/e6a38d78034ef3d3624b98f6e47a163b2c54c7c0/src/postgraphile/http/subscriptions.ts#L184-L198).

Having done this, I took the liberty of completely dropping `subscriptions-transport-ws` from PostGraphiQL too! This removes the need of keeping the `websockets` option around; since the server supports both, it would end up only being used by the frontend. Not only this, but using just `graphql-ws` increases stability, simplifies code and reduces the bundle size by [17.7kB as per Bundlephobia](https://bundlephobia.com/result?p=subscriptions-transport-ws@0.9.18).

Furthermore, [the WebSocket client in PostGraphiQL will now gracefully handle header changes by restarting the socket and resubscribing active subscriptions](https://github.com/enisdenjo/postgraphile/blob/e6a38d78034ef3d3624b98f6e47a163b2c54c7c0/postgraphiql/src/components/PostGraphiQL.js#L125-L133).

enisdenjo:
After a brief discussion with @benjie over GraphQL Slack a decision was made to allow choosing the `websockets` library on the back with an array. (If subscriptions are active, it will default to `websockets=[v0, v1]`).

Furthermore, the GraphiQL was reverted to also support both libraries. `subscriptions-transport-ws` will be used only if the server `websockets` argument provides just v0 (`[v0]`) - in all other cases (with enabled subscriptions), v1 will be used.

yarn.lock
```diff
@@ -5563,7 +5568,12 @@ istanbul-reports@^2.2.6:
   dependencies:
     html-escaper "^2.0.0"
 
-iterall@^1.0.2, iterall@^1.2.1, iterall@^1.2.2:
+iterall@^1.0.2, iterall@^1.2.2:
+  version "1.2.2"
+  resolved "https://registry.yarnpkg.com/iterall/-/iterall-1.2.2.tgz#92d70deb8028e0c39ff3164fdbf4d8b088130cd7"
+  integrity sha512-yynBb1g+RFUPY64fTrFv7nsjRrENBQJaX2UL+2Szc9REFrSNm1rpSXHGzhmAy7a9uv3vlvgBlXnf9RqmPH1/DA==
+
+iterall@^1.2.1:
```
benjie: Duplicate versions

src/postgraphile/http/subscriptions.ts
```diff
@@ -169,151 +177,308 @@ export async function enhanceHttpServerWithSubscriptions<
...
+          wss.emit('connection', ws, req);
+        });
+      } else if (v1Wss) {
+        // v1 will welcome its own subprotocol `graphql-transport-ws`
+        // and gracefully reject invalid ones
+        const wss = v1Wss;
+        wss.handleUpgrade(req, socket, head, ws => {
+          wss.emit('connection', ws, req);
+        });
+      }
```
benjie, line 481: If both protocols are supported then v1 should prevail; also please extract the duplicated logic from the `if` statement (e.g. use a ternary to define `wss`)

benjie (changes requested):
Really close now :+1:

src/postgraphile/http/subscriptions.ts
```diff
@@ -169,151 +178,307 @@ export async function enhanceHttpServerWithSubscriptions<
...
+    const { pathname = '' } = parseUrl(req) || {};
+    const isGraphqlRoute = pathname === graphqlRoute;
+    if (isGraphqlRoute) {
+      const protocol = req.headers['sec-websocket-protocol'];
+      const protocols = Array.isArray(protocol)
+        ? protocol
+        : protocol?.split(',').map(p => p.trim());
+
+      const wss =
+        protocols?.includes('graphql-ws') && !protocols.includes(GRAPHQL_TRANSPORT_WS_PROTOCOL)
```
benjie, line 471:
```suggestion
        v0Wss && protocols?.includes('graphql-ws') && !protocols.includes(GRAPHQL_TRANSPORT_WS_PROTOCOL)
```

benjie (approved):
Awesome; looks good to me! Thanks for all your hard work on this :raised_hands:

PR #1343, opened 2020-09-04
fix(types): update next function for Express compatability

purge:


benjie (approved):
This is due to the broader typing of the `NextFunction` in https://github.com/DefinitelyTyped/DefinitelyTyped/blob/77357b2c3de87fdaee7aa75487b5e767fa26dbec/types/express-serve-static-core/index.d.ts#L34-L41

Changing to any to match makes sense to me; thanks @purge :raised_hands:

PR #1381, opened 2020-11-03
Add options.additionalOptionsFromRequest

heri16:
## Description

Allow dynamically updating options such as `jwtPublicKey` based on the content of each request.

Indirectly allows support for multiple jwk issuers.

### Usage

```js
const keyfetch = require('keyfetch');
const jwt = require('jsonwebtoken');

const additionalOptionsFromRequest = (req) => {
  const jwtToken = jwtSecret ? getJwtToken(req) : null;

  // OpenID Connect uses 'iss' within the payload instead of 'jku' within the header
  var token = jwt.decode(jwtToken, { json: true, complete: true })

  if (!isTrustedIssuer(token.payload.iss)) {
    throw new Error("untrusted issuer");
  }

  return {
    jwtPublicKey: ({ kid, jku }, callback) => {
      // OpenID Connect uses 'iss' within the payload instead of 'jku' within the header
      // See: https://github.com/auth0/node-jsonwebtoken/issues/43#issuecomment-458346362
      keyfetch.oidcJwk(kid, token.payload.iss).then((result) => callback(null, result.pem), (err) => callback(err));
    },
    jwtAudiences: ...,
    jwtRole: ['cognito:role'],
    jwtVerifyOptions: ...
  }
}

options = {
  pgDefaultRole,
  pgSettings,
  additionalOptionsFromRequest,
}
```

## Performance impact

Minimal

## Security impact

None

## Checklist

### DIFF
--- a/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
+++ b/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
@@ -126,10 +126,15 @@ function withPostGraphileContextFromReqResGenerator(
     pgSettings: pgSettingsGenerator,
     allowExplain: allowExplainGenerator,
     jwtSecret,
+    additionalOptionsFromRequest,
     additionalGraphQLContextFromRequest,
   } = options;
   return async (req, res, moreOptions, fn) => {
     const jwtToken = jwtSecret ? getJwtToken(req) : null;
+    const additionalOptions =
+      typeof additionalOptionsFromRequest === 'function'
+        ? await additionalOptionsFromRequest(req, res)
+        : {};
     const additionalContext =
       typeof additionalGraphQLContextFromRequest === 'function'
         ? await additionalGraphQLContextFromRequest(req, res)
@@ -149,6 +154,7 @@ function withPostGraphileContextFromReqResGenerator(
         pgSettings,
         explain: allowExplain && req.headers['x-postgraphile-explain'] === 'on',
         ...moreOptions,
+        ...additionalOptions,
       },
       context => {
         const graphqlContext = additionalContext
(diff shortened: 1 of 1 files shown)


heri16:
@benjie Not sure why the tests are failing.

benjie:
This is an interesting idea; however it has very significant repercussions for performance optimisations both now and in future. At the moment we can assume `options` is fixed so we can cache/memoize based on these values, however if we were to merge this then we can no longer make this assumption and many of the optimisations we've either made or may want to make in future would no longer be possible. Take for example the GraphiQL HTML which depends on various settings from `options` but is only compiled once - with this approach we'd either have to undo that optimisation, or we'd have to document that GraphiQL only respects the first set of `options` the server is provided with which could lead to very confusing results. Further many options may not make sense to change at runtime, or may be hard to achieve (e.g. if you changed `watchPg` we'd have to unsubscribe or resubscribe to watch mode, which is a whole thing, and not something that this PR handles; similarly `subscriptions: true` to `subscriptions: false` or vice versa would be hard to achieve AND would have repercussions for the previously compiled/loaded GraphiQL instance/etc).

Alas for the reasons above I have to reject this PR as it stands. However, I'm open to you making the `jwtPublicKey` accept a callback function to enable it specifically to change on a per-request basis. I'm going to close this PR because the direction to take from here is too far removed from the initial changes, but please do feel free to raise a follow up PR making jwtPublicKey accept a callback in addition to everything else it accepts currently.

Thank you for taking the time to contribute to PostGraphile.

benjie:
(The tests were failing because the code wasn't formatted with prettier (run `yarn lint:fix` to resolve), and because some of the changes did not pass the TypeScript checks.)

Issue #461, opened 2017-05-10
How to test?


kenips:
Hey guys, thanks for creating this awesome project. I'm wondering what's the best way to add tests to my own project. I see that this project itself is using Jest and have tests loading the example (the integration tests). However it seems to need support files in various directory and doing quite a few different things. Do you guys have an example on how you do tests on your own projects? Thanks in advance.

mull:
I think a general write up on how to practically build projects using postgraphql might be useful. Some of the questions that arose for me:

## Testing
How do you ensure that 1) you wrote your procedures/policies/grants correctly and 2) keep someone from breaking them

## Organization
How do you go about organizing your files, i.e. your DB schema? What is a good way to structure your project so that someone new to it can grok it, and so that code review is relatively easy. One file per table? One for tables, one for functions, one for policies, one for triggers?

## Talking to other services
While relatively easy it might help everyone to see a sample project that may for example support authenticating through Google OAuth.

FWIW with some loose pointers I would give number three a shot. I would imagine you just set up a server somewhere that handles the initial call to Google and the callback, and then calls a function in the DB.

benjie:
I use `jest` for testing, I run a bunch of tests against the database. I have the entire schema written out to a file `db/schema.sql` using `pg_dump -s -O -f db/schema.sql mindtree`; and then tests in `db/__tests__/*.test.js`, e.g. this one enforces that RLS is enabled on every table in my public schema:

```js
import { withRootDb } from '../test_helpers';

test('RLS is enabled for all tables in mindtree_public', () => withRootDb(async (client) => {
  const { rows } = await client.query(`
  WITH nsp AS (
    SELECT oid FROM pg_namespace WHERE nspname = 'mindtree_public'
  )
  SELECT relname AS "table"
  FROM pg_class
  INNER JOIN nsp
  ON (nsp.oid = pg_class.relnamespace)
  WHERE relkind = 'r'
  AND relrowsecurity IS NOT DISTINCT FROM FALSE
  `);
  expect(rows).toHaveLength(0);
}));
```

The `test_helpers` file in `db` has a bunch of DB test helpers, including:

```js
const withDbFromUrl = async (url, fn) => {
  const client = new pg.Client(url);
  await client.connect();
  await client.query('begin;');
  try {
    await fn(client);
  } finally {
    await client.query('rollback;');
    await client.end();
  }
};

[...]

PR #1396, opened 2020-11-18
Add a close function to the middleware

paulmelnikow:
## Description

This adds a `close()` function to the Postgraphile middleware. As discussed in https://github.com/graphile/postgraphile/issues/461#issuecomment-531247990 this allows for a clean shutdown of the PgPool and the watchers. It's particularly useful for unit tests.

I've marked it a work in progress. I'm curious to start getting comments and to think about how to incorporate [the idea of  shutdownActions from Graphile Starter](https://github.com/graphile/starter/blob/58699dd1b759a3dd8e6f50cdc0622979f037e897/%40app/server/README.md#shutdownactions).

I've tested this locally with my unit tests, and it does clean everything up locally when `watchPg` is on and when it's off.

Really eager for your feedback!

## Performance impact

None

## Security impact

None

## Checklist

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -320,6 +320,7 @@ export interface CreateRequestHandlerOptions extends PostGraphileOptions {
   // A Postgres client pool we use to connect Postgres clients.
   pgPool: Pool;
   _emitter: EventEmitter;
+  close: () => Promise<void>;
 }
 
 export interface GraphQLFormattedErrorExtended {
@@ -372,6 +373,7 @@ export interface HttpRequestHandler<
   faviconRouteHandler: ((res: PostGraphileResponse) => Promise<void>) | null;
   eventStreamRoute: string;
   eventStreamRouteHandler: ((res: PostGraphileResponse) => Promise<void>) | null;
+  close: Promise<void>;
 }
 
 /**
--- a/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
+++ b/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
@@ -177,6 +177,7 @@ export default function createPostGraphileHttpRequestHandler(
     pgPool,
     pgSettings,
     pgDefaultRole,
+    close,
     queryCacheMaxSize = 50 * MEGABYTE,
     extendedErrors,
     showErrorStack,
@@ -1102,6 +1103,7 @@ export default function createPostGraphileHttpRequestHandler(
   middleware.faviconRouteHandler = graphiql ? faviconRouteHandler : null;
   middleware.eventStreamRoute = eventStreamRoute;
   middleware.eventStreamRouteHandler = watchPg ? eventStreamRouteHandler : null;
+  middleware.close = close;
 
   const hookedMiddleware = pluginHook('postgraphile:middleware', middleware, {
     options,
--- a/src/postgraphile/postgraphile.ts
+++ b/src/postgraphile/postgraphile.ts
@@ -29,6 +29,7 @@ export interface PostgraphileSchemaBuilder<
   _emitter: EventEmitter;
   getGraphQLSchema: () => Promise<GraphQLSchema>;
   options: PostGraphileOptions<Request, Response>;
+  cancelWatch: null | (() => Promise<void>);
 }
 
 /**
@@ -43,18 +44,27 @@ export function getPostgraphileSchemaBuilder<
   pgPool: Pool,
   schema: string | Array<string>,
   incomingOptions: PostGraphileOptions<Request, Response>,
-  release: null | (() => void) = null,
+  releasePgPool: null | (() => void) = null,
 ): PostgraphileSchemaBuilder {
+  let cancelWatchFn: null | (() => Promise<void>) = null;
+  async function cancelWatch() {
+    if (cancelWatchFn) {
+      await cancelWatchFn();
+      cancelWatchFn = null;
+    }
+  }
+
   let released = false;
-  function releaseOnce() {
+  async function releaseOnce() {
     if (released) {
       throw new Error(
         'Already released this PostGraphile schema builder; should not have attempted a second release',
       );
     }
     released = true;
-    if (release) {
-      release();
+    await cancelWatch();
+    if (releasePgPool) {
+      releasePgPool();
     }
   }
   if (incomingOptions.live && incomingOptions.subscriptions == null) {
@@ -96,6 +106,7 @@ export function getPostgraphileSchemaBuilder<
     _emitter,
     getGraphQLSchema: () => (gqlSchema ? Promise.resolve(gqlSchema) : gqlSchemaPromise),
     options,
+    cancelWatch,
   };
 
   async function createGqlSchema(): Promise<GraphQLSchema> {
@@ -104,7 +115,10 @@ export function getPostgraphileSchemaBuilder<
     while (true) {
       try {
         if (options.watchPg) {
-          await watchPostGraphileSchema(pgPool, pgSchemas, options, newSchema => {
+          if (cancelWatchFn) {
+            throw new Error(`cancelWatchFn should not have been set at this point.`);
+          }
+          cancelWatchFn = await watchPostGraphileSchema(pgPool, pgSchemas, options, newSchema => {
             gqlSchema = newSchema;
             _emitter.emit('schemas:changed');
             exportGqlSchema(gqlSchema);
@@ -137,7 +151,7 @@ export function getPostgraphileSchemaBuilder<
             const diff = process.hrtime(start);
             const dur = diff[0] * 1e3 + diff[1] * 1e-6;
             if (!retry) {
-              releaseOnce();
+              await releaseOnce();
               throw error;
             } else {
               if (dur < 50) {
@@ -258,7 +272,7 @@ export default function postgraphile<
 
   // Do some things with `poolOrConfig` so that in the end, we actually get a
   // Postgres pool.
-  const { pgPool, release } = toPgPool(poolOrConfig);
+  const { pgPool, releasePgPool } = toPgPool(poolOrConfig);
 
   pgPool.on('error', err => {
     /*
@@ -278,18 +292,27 @@ export default function postgraphile<
     debugPgClient(pgClient, !!options.allowExplain);
   });
 
-  const { getGraphQLSchema, options, _emitter } = getPostgraphileSchemaBuilder<Request, Response>(
-    pgPool,
-    schema,
-    incomingOptions,
-    release,
-  );
+  const { getGraphQLSchema, options, _emitter, cancelWatch } = getPostgraphileSchemaBuilder<
+    Request,
+    Response
+  >(pgPool, schema, incomingOptions, releasePgPool);
+
+  async function close() {
+    if (cancelWatch) {
+      await cancelWatch();
+    }
+    if (releasePgPool) {
+      await releasePgPool();
+    }
+  }
+
   return createPostGraphileHttpRequestHandler({
     ...(typeof poolOrConfig === 'string' ? { ownerConnectionString: poolOrConfig } : {}),
     ...options,
     getGqlSchema: getGraphQLSchema,
     pgPool,
     _emitter,
+    close,
   });
 }
 
@@ -321,24 +344,24 @@ function constructorName(obj: mixed): string | null {
 }
 
 // tslint:disable-next-line no-any
-function toPgPool(poolOrConfig: any): { pgPool: Pool; release: null | (() => void) } {
+function toPgPool(poolOrConfig: any): { pgPool: Pool; releasePgPool: null | (() => void) } {
   if (quacksLikePgPool(poolOrConfig)) {
     // If it is already a `Pool`, just use it.
-    return { pgPool: poolOrConfig, release: null };
+    return { pgPool: poolOrConfig, releasePgPool: null };
   }
 
   if (typeof poolOrConfig === 'string') {
     // If it is a string, let us parse it to get a config to create a `Pool`.
     const pgPool = new Pool({ connectionString: poolOrConfig });
-    return { pgPool, release: () => pgPool.end() };
+    return { pgPool, releasePgPool: () => pgPool.end() };
   } else if (!poolOrConfig) {
     // Use an empty config and let the defaults take over.
     const pgPool = new Pool({});
-    return { pgPool, release: () => pgPool.end() };
+    return { pgPool, releasePgPool: () => pgPool.end() };
   } else if (isPlainObject(poolOrConfig)) {
     // The user handed over a configuration object, pass it through
     const pgPool = new Pool(poolOrConfig);
-    return { pgPool, release: () => pgPool.end() };
+    return { pgPool, releasePgPool: () => pgPool.end() };
   } else {
     throw new Error('Invalid connection string / Pool ');
   }


src/postgraphile/postgraphile.ts
```diff
@@ -104,7 +115,10 @@ export function getPostgraphileSchemaBuilder<
     while (true) {
       try {
         if (options.watchPg) {
-          await watchPostGraphileSchema(pgPool, pgSchemas, options, newSchema => {
+          if (cancelWatchFn) {
+            throw new Error(`cancelWatchFn should not have been set at this point.`);
+          }
```
paulmelnikow, line 120: It's not clear to me whether `watchPostGraphileSchema()` will ever be successfully invoked more than once. What I've got here is the brash thing, which is assert it doesn't happen. If we know it could be invoked more than once, then the thing to do, I think, would be invoke and nullify `cancelWatchFn` here instead.

src/postgraphile/postgraphile.ts
```diff
@@ -137,7 +152,7 @@ export function getPostgraphileSchemaBuilder<
             const diff = process.hrtime(start);
             const dur = diff[0] * 1e3 + diff[1] * 1e-6;
             if (!retry) {
-              releaseOnce();
+              await releaseOnce();
```
benjie, line 155: If `releaseOnce` were to throw an error, this new error would be raised to the user rather than the original error, and the new error will contain less useful/addressable information. Instead we should swallow any new errors and continue to throw the old.

```suggestion
              // Swallow new error so old error is still thrown
              await releaseOnce().catch(e => {console.error(e)});
```

benjie (changes requested):
I like this general approach, my main feedback is:

- releasers should be added to `shutdownActions` directly near where the actions are set up (rather than passing around separate release functions which are then aggregated later)
- shutdownActions should be accessible from server plugins, such that server plugins may add (and remove) their own shutdown actions

Having both `release` and `shutdownActions` exposed on `middleware` would be acceptable for this if it makes most sense, however, I think exposing `shutdownActions` through the pluginHook's third argument might be a better approach as it keeps the exposed interface for consumers cleaner; it's likely that it will only be needed in certain hooks; I'm thinking `postgraphile:middleware` and _maybe_ `withPostGraphileContext` are the only pluginHooks we care about here.

paulmelnikow:
Thanks for the feedback! I will aim to pick this up on Monday.

benjie:
@paulmelnikow Gentle nudge :wink:

benjie:
Awesome 👏 Merge preferred; we squash at the end so no need to keep history flat.

src/postgraphile/postgraphile.ts
paulmelnikow: > I can't remember if `getPostgraphileSchemaBuilder` is a public API or not off-hand

It looks to me like it's not public. It's imported in `cli.ts` and `postgraphile.ts` and as far as I can tell it's not re-exported anywhere. Could you confirm? @benjie

paulmelnikow:
Hi @benjie, I think this is ready for another review. `getPostgraphileSchemaBuilder()` is now relying on the ShutdownActions instance. I'm not totally sure that's the right thing, but it's really simple and the process is now exiting cleanly as intended.

> -    shutdownActions should be accessible from server plugins, such that server plugins may add (and remove) their own shutdown actions
>
> Having both release and shutdownActions exposed on middleware would be acceptable for this if it makes most sense, however, I think exposing shutdownActions through the pluginHook's third argument might be a better approach as it keeps the exposed interface for consumers cleaner; it's likely that it will only be needed in certain hooks; I'm thinking postgraphile:middleware and maybe withPostGraphileContext are the only pluginHooks we care about here.

Exposing the release function on the middleware is working well for my unit tests. I guess you're suggesting instead of invoking `release()` I could register a plugin hook instead. That seems possible, although it would increase the amount of boilerplate consumers would need to include in their test setup for this common use case.

In 6078d88a I added `shutdownActions` to the `postgraphile:middleware` function. Is this what you are looking for?

I'm not quite sure what would need to be done for `withPostGraphileContext`. Could you elaborate?

paulmelnikow:
Hey @benjie, just wanted to ping you about this 😉

benjie:
> In 6078d88 I added shutdownActions to the postgraphile:middleware function. Is this what you are looking for?
> I'm not quite sure what would need to be done for withPostGraphileContext. Could you elaborate?

Yes; and I want the same thing for the `withPostGraphileContext` hook here: https://github.com/graphile/postgraphile/blob/601a1086a2c10ebae4747364e199f2517d653861/src/postgraphile/withPostGraphileContext.ts#L272-L274

src/postgraphile/postgraphile.ts
```diff
@@ -104,11 +93,17 @@ export function getPostgraphileSchemaBuilder<
...
+            pgPool,
+            pgSchemas,
+            options,
+            newSchema => {
+              gqlSchema = newSchema;
+              _emitter.emit('schemas:changed');
+              exportGqlSchema(gqlSchema);
+            },
+          );
+          shutdownActions.add(releaseWatchFn);
```
benjie, line 106: To solve the race condition we should register a shutdown action immediately, not after the promise resolves.

```suggestion
          const releaseWatchFnPromise = watchPostGraphileSchema(
            pgPool,
            pgSchemas,
            options,
            newSchema => {
              gqlSchema = newSchema;
              _emitter.emit('schemas:changed');
              exportGqlSchema(gqlSchema);
            },
          );
          // We must register the shutdown action immediately to avoid a race condition.
          shutdownActions.add(async () => {
            const releaseWatchFn = await releaseWatchFnPromise;
            releaseWatchFn();
          });
          // Wait for the watch to be set up before progressing.
          await releaseWatchFnPromise;
```

src/postgraphile/shutdownActions.ts
```diff
@@ -0,0 +1,42 @@
...
+  private didInvoke = false;
+
+  add(action: Action): void {
+    this.actions.push(action);
+  }
+
+  remove(action: Action): void {
+    const index = this.actions.indexOf(action);
+    if (index === -1) {
+      throw Error('The specified shutdown action was not found.');
```
benjie, line 14:
```suggestion
      throw new Error('The specified shutdown action was not found.');
```

src/postgraphile/shutdownActions.ts
```diff
@@ -0,0 +1,42 @@
...
+        return fn();
+      } catch (e) {
+        return Promise.reject(e);
+      }
+    });
+  }
+
+  async invokeAll(): Promise<void> {
+    // This would be better if it used `Promise.allSettled()` but we can't use
+    // it yet.
```
benjie, line 39:
```suggestion
    // TODO:v5: This would be better if it used `Promise.allSettled()` but we can't
    // use it yet.
```

benjie (changes requested):
This is looking good to me; few more things to address before we can merge it but I like the approach.

src/postgraphile/postgraphile.ts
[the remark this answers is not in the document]
benjie: (Addressed below)

src/postgraphile/shutdownActions.ts
```diff
@@ -0,0 +1,42 @@
+export type Action = () => void | Promise<void>;
+
+export class ShutdownActions {
+  private actions: Action[] = [];
+  private didInvoke = false;
+
+  add(action: Action): void {
+    this.actions.push(action);
```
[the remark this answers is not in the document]
paulmelnikow: Another option could be to run the cleanup function and throw an error. Would that be safer?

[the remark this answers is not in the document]
benjie: I don’t think so; nothing would be able to catch the error?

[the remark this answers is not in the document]
paulmelnikow: Right. I was thinking this condition was a programmer error, in which case an unhandled exception may be more appropriate than a silent failure.

When would this happen? Is it when the programmer initiates a shutdown while some part of the server is still starting up? Is that a programmer error?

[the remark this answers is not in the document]
benjie: It'd be an error in one of the plugins that they're using; and given they may not be able to fix those plugins directly it'd be pretty mean to cause a service shutdown. I think a warning (slap on the wrist) and gracefully handling the problem is best. Basically shutdown actions should be registered _in the same tick_ as when the thing they start is started - like I've done with the `watchPostGraphileSchema` code below :+1: 

Keep in mind that this is useful for more than just tests; there are people who spin up and tear down postgraphile instances at runtime and this is a great feature for them :+1:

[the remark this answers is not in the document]
paulmelnikow: Sounds good, and thanks for clarifying! I'll aim to wrap up these changes later today.

src/postgraphile/postgraphile.ts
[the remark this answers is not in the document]
paulmelnikow: Committed your change and made a small tweak in 93cb7b7.

paulmelnikow:
I think I've addressed all your feedback! If you'd like me to add some documentation, could you point me to where it should go?

benjie:
This looks good; I need to give it a full check over and decide where to go from there, but from a quick scan I don't think any further changes are required :+1:  Please make sure I can push to this branch (i.e. make sure "Allow edits from maintainers." is checked in the sidebar).

paulmelnikow:
Sounds good.

For some reason I don't understand, I'm not seeing that "allow edits from maintainers" checkbox. I've actually not seen this before. Maybe it's because this is on an organization account? I generally turn this on when I create PRs but since the option isn't coming up, I don't see a way to check.

This post describes an action you can take to verify whether or not you have edit access. Could you check and see? https://stackoverflow.com/a/60064666/893113

benjie:
Thanks; I didn't know about that way to check! I've checked and alas it's not editable:

Screenshot_20210514_155515

I'm not sure I actually need to edit it, but always best to have the option!

paulmelnikow:
Boo! For some reason this PR is still not showing the checkbox that would give you access. I gave you write access to our fork temporarily, so if you accept the invitation you'll be able to push to this branch. The only other way I can see to do this is to move this branch to a different fork and try to open a new PR. Let me know if you'd prefer I do that!

src/postgraphile/postgraphile.ts
```diff
@@ -104,11 +93,23 @@ export function getPostgraphileSchemaBuilder<
...
+            newSchema => {
+              gqlSchema = newSchema;
+              _emitter.emit('schemas:changed');
+              exportGqlSchema(gqlSchema);
+            },
+          );
+          // We must register the shutdown action immediately to avoid a race condition.
+          shutdownActions.add(async () => {
+            const releaseWatchFn = await releaseWatchFnPromise;
+            releaseWatchFn();
```
benjie, line 109: We should return this, I think it's a promise in itself.

```suggestion
            return releaseWatchFn();
```

src/postgraphile/shutdownActions.ts
```diff
@@ -0,0 +1,54 @@
+export type Action = () => void | Promise<void>;
+
+export class ShutdownActions {
```
benjie, line 3:
```suggestion
/**
 * This class tracks actions that must be taken
 * when PostGraphile is shut down (released) in
 * order to make sure that all the resources it
 * has consumed has been cleaned up.
 *
 * Since PostGraphile is extensible via plugins
 * and plugins may have their own shutdown actions
 * we use this generic system to handle them all.
 * If your plugin sets up something that would
 * keep running after PostGraphile is `.release()`d
 * (such as a timer, interval, network connection,
 * filesystem monitoring, or similar activity)
 * then you should register an action via
 * `shutdownActions` to clean it up.
 */
export class ShutdownActions {
```

src/postgraphile/shutdownActions.ts
```diff
@@ -0,0 +1,54 @@
...
+  private actions: Action[] = [];
+  private didInvoke = false;
+
+  add(action: Action): void {
```
benjie, line 7:
```suggestion
  /**
   * Register a function to be called when PostGraphile is released.
   */
  add(action: Action): void {
```

src/postgraphile/shutdownActions.ts
```diff
@@ -0,0 +1,54 @@
...
+          console.error('Error occurred calling shutdown action after invoke:');
+          console.error(e);
+        });
+      });
+    } else {
+      this.actions.push(action);
+    }
+  }
+
+  remove(action: Action): void {
```
benjie, line 23:
```suggestion
  /**
   * If your action is no longer relevant (for example it has
   * completed, or it was only relevant whilst in a particular
   * mode) then be sure to remove it so it won't be called
   * when PostGraphile is released.
   */
  remove(action: Action): void {
```

src/postgraphile/shutdownActions.ts
```diff
@@ -0,0 +1,54 @@
...
+    const index = this.actions.indexOf(action);
+    if (index === -1) {
+      throw new Error('The specified shutdown action was not found.');
+    }
+    this.actions.splice(index, 1);
+  }
+
+  invoke(): Array<Promise<void> | void> {
```
benjie, line 31:
```suggestion
  /**
   * Calls the release actions in parallel and returns
   * the array of resulting promises/results. Will not
   * throw unless the shutdown actions have already
   * been invoked.
   */
  invoke(): Array<Promise<void> | void> {
```

src/postgraphile/shutdownActions.ts
```diff
@@ -0,0 +1,54 @@
...
+      // Ensure that all actions are called, even if a previous action throws an error
+      try {
+        return fn();
+      } catch (e) {
+        return Promise.reject(e);
+      }
+    });
+  }
+
+  async invokeAll(): Promise<void> {
```
benjie, line 49:
```suggestion
  /**
   * Calls all the release actions and resolves
   * when complete (rejecting if an error occurred).
   */
  async invokeAll(): Promise<void> {
```

benjie (changes requested):
LGTM! One missing `return` and it would benefit from some comments. I've added some but they'll need wordwrapping properly.

Please can you add some tests that exercise this code?

benjie:
Invitation accepted; files now editable but I'll leave you to make the changes anyway :+1:

paulmelnikow:
> Please can you add some tests that exercise this code?

I'm happy to add some tests but I need guidance on the testing strategy.

The tests in postgraphile-test and withPostGraphileContext-test seem to be mocking too much stuff to be useful here, and the integration tests seem more involved than what's needed. So I'm guessing I should create a new test file with a new kind of test, which like the integration tests, hits postgres and uses minimal mocks.

There are a lot of good things to test:

1. Can release() be invoked without causing an error
2. Does release() release the pg pool
3. Does release() release the pg pool with pgWatch
4. If you add a custom releaser using the hook, does it get invoked

I'll aim for the first one to start.

src/postgraphile/postgraphile.ts
benjie: I checked this; it’s fine 👍

benjie:
Yes; a new integration test is best 👍 Want to make sure it works for watch mode too. Have a go and see how far you get; I’m happy to chime in with guidance. @ me on Discord for faster responses.

src/__tests__/utils/pgPool.ts
```diff
@@ -3,10 +3,12 @@ import { parse as parsePgConnectionString } from 'pg-connection-string';
...
 
-const pgPool = new Pool({
+export const poolConfig = {
   ...parsePgConnectionString(pgUrl),
   max: 15,
   idleTimeoutMillis: 500,
-});
+};
+
+const pgPool = new Pool(poolConfig);
```
benjie, line 12: We need to swallow pool errors. https://github.com/graphile/postgraphile/blob/601a1086a2c10ebae4747364e199f2517d653861/src/postgraphile/postgraphile.ts#L264-L271

```suggestion
const pgPool = new Pool(poolConfig);
pgPool.on('error', () => {/* swallow error */});
```

src/postgraphile/__tests__/postgraphileIntegrationShutdown-test.js
```diff
@@ -0,0 +1,29 @@
...
+import pgPool, { poolConfig } from '../../__tests__/utils/pgPool';
+import { postgraphile } from '..';
+
+const COMMON_OPTIONS = {
+  exitOnFail: false,
+};
+
+test('When the handler is created using a Pool object, it can be released without triggering an error', async () => {
+  let handler = postgraphile(pgPool, COMMON_OPTIONS);
+  await handler.release();
```
benjie, line 12:
```suggestion
  let handler = postgraphile(pgPool, COMMON_OPTIONS);
  await handler.release();
  // This would throw an error if the pool was released, it not throwing is what passes the test.
  await pgPool.query("select 'Pool is still valid';");
```

src/postgraphile/__tests__/postgraphileIntegrationShutdown-test.js
```diff
@@ -0,0 +1,29 @@
...
+  expect(pgPool).toHaveProperty('ended', true);
+});
+
+test('When the handler is created in watch mode using Pool config, it releases the pool it creates', async () => {
+  let handler = postgraphile(poolConfig, { ...COMMON_OPTIONS, watchPg: true });
+  let { pgPool } = handler;
+  expect(pgPool).toHaveProperty('ended', false);
+  await handler.release();
+  expect(pgPool).toHaveProperty('ended', true);
+});
```
benjie, line 29: These tests are great; can you also add one that waits ~100ms before triggering? The watch mode one seems like a good one to pick for this.

src/postgraphile/postgraphile.ts
```diff
@@ -100,22 +89,52 @@ export function getPostgraphileSchemaBuilder<
...
+            pgSchemas,
+            options,
+            newSchema => {
+              gqlSchema = newSchema;
+              _emitter.emit('schemas:changed');
+              exportGqlSchema(gqlSchema);
+            },
+          );
+          // We must register the shutdown action immediately to avoid a race condition.
+          shutdownActions.add(async () => {
```
benjie: This registers a shutdown action during the loop but never releases it; this captures the closure (and its variables) forever and could result in memory exhaustion. I've refactored the code to only register the shutdown actions outside the loop.

benjie (changes requested):

benjie:
Please note that I pushed commits to this branch; please validate those commits work for you :+1:

benjie:
(We're basically there now, I think.)

src/interfaces.ts
```diff
@@ -377,13 +379,15 @@ export interface HttpRequestHandler<
   faviconRouteHandler: ((res: PostGraphileResponse) => Promise<void>) | null;
   eventStreamRoute: string;
   eventStreamRouteHandler: ((res: PostGraphileResponse) => Promise<void>) | null;
+  release: () => Promise<void>;
```
benjie, line 382:
```suggestion
  /** Experimental! */
  release: () => Promise<void>;
```

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -1120,9 +1121,12 @@ export default function createPostGraphileHttpRequestHandler(
   middleware.faviconRouteHandler = graphiql ? faviconRouteHandler : null;
   middleware.eventStreamRoute = eventStreamRoute;
   middleware.eventStreamRouteHandler = watchPg ? eventStreamRouteHandler : null;
+  middleware.shutdownActions = shutdownActions;
+  middleware.release = (): Promise<void> => shutdownActions.invokeAll();
```
benjie, line 1125:
```suggestion
  // Experimental
  middleware.release = (): Promise<void> => shutdownActions.invokeAll();
```

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -1120,9 +1121,12 @@ export default function createPostGraphileHttpRequestHandler(
...
 
   const hookedMiddleware = pluginHook('postgraphile:middleware', middleware, {
     options,
+    shutdownActions,
```
benjie, line 1129:
```suggestion
    // Experimental
    shutdownActions,
```

src/postgraphile/shutdownActions.ts
```diff
@@ -0,0 +1,84 @@
...
+ * This class tracks actions that must be taken when PostGraphile is shut down
+ * (released) in order to make sure that all the resources it has consumed has
+ * been cleaned up.
+ *
+ * Since PostGraphile is extensible via plugins and plugins may have their own
+ * shutdown actions we use this generic system to handle them all. If your
+ * plugin sets up something that would keep running after PostGraphile is
+ * `.release()`d (such as a timer, interval, network connection, filesystem
+ * monitoring, or similar activity) then you should register an action via
+ * `shutdownActions` to clean it up.
```
benjie, line 13:
```suggestion
 * `shutdownActions` to clean it up.
 * 
 * This class currently has "experimental" status, it may have breaking
 * changes in future semver minor releases.
```

benjie (changes requested):
Note that this is experimental

benjie (approved):

paulmelnikow:
Sorry I didn't get to run this yesterday! And thanks for merging this!

benjie:
Thanks for writing it and putting up with my pedantic feedback :grin:

paulmelnikow:
Happy to work with you on this! It's a good way to slowly learn about the project. And in fairness, it's a pedantic feature. 😉

PR #1407, opened 2020-11-30
fix(docker): remove GRAPHILE_TURBO

benjie:
## Description

Dockerfile references Node 12; GRAPHILE_TURBO now requires Node 14

## Performance impact

The Docker Hub image will run slower; npm build is unaffected.

## Security impact

None.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,7 +44,6 @@ FROM node:12-alpine
 LABEL description="Instant extensible high-performance GraphQL API for your PostgreSQL database https://graphile.org/postgraphile"
 
 EXPOSE 5000
-ENV GRAPHILE_TURBO=1
 WORKDIR /postgraphile/
 ENTRYPOINT ["./cli.js"]
 


EyMaddis:
@benjie  According to this file it is still v12:
https://github.com/graphile/postgraphile/blob/af39c3dcea40646c9217785fa93a9601612c2b83/isTurbo.js#L4

Cannot document what this actually does? I only found https://medium.com/graphile/how-i-made-postgraphile-faster-than-prisma-1-year-later-e6bf3a802f58 is still still correct?

benjie:
GRAPHILE_TURBO is a flag that opts you into optimisations only available in the latest release of Node.js (changes over time). It’s a moving target, hence it is an opt-in flag. Remove it and all should be well.

EyMaddis:
I did not mean that I got the error, I wanted to point out that the error message is wrong

benjie:
Ah yeah; I have a PR for that: https://github.com/graphile/postgraphile/pull/1411

PR #1432, opened 2021-01-20
chore(deps): upgrade a load of dependencies/devDependencies

benjie:
## Description

Most things, to be honest.

## Performance impact

Unknown.

## Security impact

Hopefully an improvement :crossed_fingers: 

## Checklist


benjie:
We had to use non-latest versions of some things:

- Webpack v5 breaks a bunch of stuff I couldn't be bothered to resolve, so reverted to v4
- Upgrading @types/node to 14.x causes lots of TypeScript issues, so I reverted
- Jest 26.x drops support for Node 8 so CI starts failing; so we've upgraded from Jest 24 to 25 but we're stuck there.

PR #1497, opened 2021-06-09
chore(deps): bump graphql-ws to 5.1.2

enisdenjo:
## Description

[`graphql-ws@v5.0.0`](https://github.com/enisdenjo/graphql-ws/releases/tag/v5.0.0) introduces a few, minimal, breaking changes. One of the quite important one is the addition of the subprotocol level ping and pong messages (https://github.com/enisdenjo/graphql-ws/pull/201#issue-663733666).

The implementation in `postgraphile` is untouched.

Please note that the client will NOT ping the server by default. Meaning, this upgrade is not critical.

## Performance impact

None.

## Security impact

None.

## Checklist

- [ ] ~My code matches the project's code style and `yarn lint:fix` passes.~


benjie:
Should we leave this for the next PostGraphile major bump, or can this be merged in such a way that it's backwards compatible with existing clients? e.g. if we default to not issuing ping/pongs then will that mean the existing protocol is not broken, and users can opt-in to this behaviour?

I'm glad you're adding ping/pong to the protocol :+1:

enisdenjo:
The server is 100% backwards compatible with all clients. [Server-side still relies on the WebSocket transport level ping/pongs.](https://github.com/enisdenjo/graphql-ws/blob/d3214731388070df5dfafde4762d47d6e81af88c/src/use/ws.ts#L67-L90) Quite literally, nothing changed there (except for the fact that it responds to subprotocol ping/pong messages).

The client options, by default, are also backwards compatible with all `graphql-ws` versions. However, if you change the [`keepAlive` client option](https://github.com/enisdenjo/graphql-ws/blob/master/docs/interfaces/client.clientoptions.md#keepalive), it will issue subprotocol level ping and pong messages to the server which will, due to validation, terminate the connection.

In essence, this update may as well be released in the next minor/patch bump. Furthermore, I'd also recommend not waiting for the major simply because nothing breaks and PostGraphile would just start accepting the new ping/pong message types.

benjie (approved):

benjie:
Thanks @enisdenjo; that's what I suspected but wanted to confirm!

PR #1548, opened 2021-10-23
Fix koa middleware going to next middleware

jsmnbom:
## Description

Currently the koa middleware will always go to the next middleware in the line. so if you have
```
app.use(postgraphile(...);
app.use(async (ctx) => {
    ctx.redirect('http://somewhere');
});
```
the request will always get redirected, even if the request was for /graphql and postgraphile actually set a response.

This PR fixes this issue, making /graphql not redirect, while everything not used in the postgraphile middleware will be redirected.

Also I noticed the koa middleware doesn't use async/await, despite koa heavily relying on it, any particular reason why?

## Performance impact
None

## Security impact
None

## Checklist

- [x] My code matches the project's code style and `yarn lint:fix` passes. *I didn't have the time to setup a dev env, so I used githubs interface, it should pass everything though*
- [x] I've added tests for the new feature, and `yarn test` passes. *does it need tests?*
- [x] If this is a breaking change I've explained why. *NOT BREAKING*

### DIFF
--- a/src/postgraphile/http/frameworks.ts
+++ b/src/postgraphile/http/frameworks.ts
@@ -295,7 +295,6 @@ export class PostGraphileResponseKoa extends PostGraphileResponse {
 
   setBody(body: Stream | Buffer | string | undefined) {
     this._ctx.body = body || '';
-    this._next();
   }
 }
 


cdaringe:
i posit that this actually correct behavior.

- middlewares call `next`
- handlers terminate

perhaps a `terminate: true` option drops `next`, but to not call next dangerously assumes some other middleware are not interested in continuing to forward the response down the pipe. unlikely, but feasible.

to not call next means it's not a middleware

benjie:
It’s not strictly a middleware, it’s a number of bundled *terminal* request handlers for Node.js’ http library, with an optional pass-through (e.g. calling next()) if the handlers are not matched in order to be compatible with what express would call a middleware. Then this has been adapted in a backwards-compatible way to work with other Node.js server frameworks as best we could manage without writing too much code to adapt it for each individual server - mostly these adaptors try and extract the underlying request/response objects from Node because that’s what the library was written for all those years ago.

benjie:
V5's koa integration doesn't call `next()` unless the route isn't handled; it does resolve to a promise though. From what I can see in the Koa docs, the behavior looks correct so I'm going to close this; but if you believe it's incorrect please either re-open this or submit a PR; either way include links to authoritative Koa sources indicating why it should be changed. Thanks!

https://github.com/graphile/crystal/blob/a42223b4c3c67f1df22e58316340c4510bd23e67/grafast/grafserv/src/servers/koa/v2/index.ts#L75

https://koajs.com/#cascading

PR #1550, opened 2021-10-28
Ignore ws crashes

purge:
- catch ws errors and ignore them
- allow handler to be overridden

## Checklist

### DIFF
--- a/src/postgraphile/http/subscriptions.ts
+++ b/src/postgraphile/http/subscriptions.ts
@@ -489,7 +489,7 @@ export async function enhanceHttpServerWithWebSockets<
               wsOptions.onError(e);
             } else {
               console.info(
-                'the following error thrown from `ws` was caught and ignored. if you want to handle it yourself pass an onError handler',
+                'The following error thrown from `ws` was caught and ignored.',
               );
               console.error(e);
             }
--- a/src/postgraphile/http/subscriptions.ts
+++ b/src/postgraphile/http/subscriptions.ts
@@ -58,6 +58,9 @@ export async function enhanceHttpServerWithWebSockets<
     keepAlive?: number;
     graphqlRoute?: string;
   },
+  wsOptions?: {
+    onError?: (e: unknown) => void;
+  },
 ): Promise<void> {
   if (websocketServer['__postgraphileSubscriptionsEnabled']) {
     return;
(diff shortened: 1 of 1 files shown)


src/postgraphile/http/subscriptions.ts
```diff
@@ -478,7 +481,21 @@ export async function enhanceHttpServerWithWebSockets<
...
             v1Wss;
+
       if (wss) {
+        wss.on('connection', function (ws) {
+          ws.on('error', function (e) {
+            if (wsOptions && wsOptions.onError) {
+              wsOptions.onError(e);
+            } else {
+              console.info(
+                'the following error thrown from `ws` was caught and ignored. if you want to handle it yourself pass an onError handler',
```
benjie, line 492:
```suggestion
                'The following error thrown from `ws` was caught and ignored.',
```

benjie (changes requested):
This looks good; are there situations in which we need to do cleanup in the 'error' handler though? Maybe @enisdenjo would know?

Interestingly neither the `ws` docs nor the `graphql-ws` docs seem to show examples adding error handlers...

benjie (approved):

enisdenjo:
Actually, [`graphql-ws@v5.5.4`](https://github.com/enisdenjo/graphql-ws/releases/tag/v5.5.4) started [doing something similar](https://github.com/enisdenjo/graphql-ws/commit/a22c00ff406f19ca706afdf6fdd7de98571bce78). Updating to the latest version, this PR might want to reconsider the then [extraneous `console.log`](https://github.com/purge/postgraphile/blob/eddfb126b48890b04ffb62a458497d5f3da1a809/src/postgraphile/http/subscriptions.ts#L491-L494) because [`useServer` leaves logs too](https://github.com/enisdenjo/graphql-ws/blob/239dafeb65d0cf057962f73e5ba7d6f2bd00c82d/src/use/ws.ts#L55-L95).

Furthermore, `graphql-ws` itself has no error handling examples because the lib logs to the console now too, and because all errors emitted on the WebSocket server (or its sockets) are considered fatal to the connection(s) anyways. After observing the error, there's nothing you can do to keep the connection alive. With this in mind, I just decided to not bother and keep the docs lean. But, maybe I should bother? 🤔

benjie:
@enisdenjo A paragraph about it might help :) "Since graphql-ws@x.y.z it's no longer necessary to add 'event' handlers to ws connections as we handle this internally because ..."

So what you're saying @enisdenjo is we should just upgrade to the latest version.

@purge since graphql-ws has a `^` range, does upgrading graphql-ws address your issue without these changes being required?

purge:
Yep, upgrading to `graphql-ws >= 5.5.4` no longer crashes the server. i'll re-raise as an updated minimum dep

PR #1559, opened 2021-11-30
fix(ws): only use singleStatement for subscriptions

benjie:
## Description

singleStatement causes a "fake" pgClient to be used; this client instigates a transaction for each and every SQL statement issued. For queries/mutations this is not desired:

- it decreases performance
- it breaks atomicity (later queries might return inconsistent results, not a major issue but not desired)
- it results in savepoint releases failing because the savepoints don't exist (this, surprisingly, does not break the transaction and thus is just a logging issue)

:rotating_light: POTENTIAL BREAKING CHANGE to plugins. It's a very subtle change, but basically the GraphQL context (which contains the `pgClient`, the `jwtClaims`, and anything that you return from `additionalGraphQLContextFromRequest`) cannot be built until _after_ the `postgraphile:ws:onOperation` / `postgraphile:ws:onSubscribe` hook is called, so these hooks will no longer have access to the final context. The only use of these hooks that I'm aware of is for implementing persisted operations which do not use context, and even if users use this hook to add other things to context because we merge context those values will still exist, so I don't anticipate this causing anyone issues.

## Performance impact

Improves performance for `websocketOperations: 'all'`

## Security impact

None known (other than performance-related and logging-related issues being solved).

## Checklist

- [x] ~~I have detailed the new feature in the relevant documentation.~~
- [x] ~~I have added this feature to 'Pending' in the `RELEASE_NOTES.md` file (if one exists).~~


benjie:
Searched for this hook in sourcegraph and there's no additional hits, so I think we can safely release this as a patch.

- https://sourcegraph.com/search?q=context:global+postgraphile:ws:onOperation&patternType=literal
- https://sourcegraph.com/search?q=context:global+postgraphile:ws:onSubscribe&patternType=literal

benjie:
Turns out we hadn't hooked up the `--websocket-operations` CLI flag either; fixed above.

benjie:
Tested with this script:

```js
import WebSocket from 'ws';
import { createClient } from 'graphql-ws';
import util from 'util';

const client = createClient({
  url: 'ws://localhost:5000/graphql',
  webSocketImpl: WebSocket,
});

// mutation
(async () => {
  const result = await new Promise((resolve, reject) => {
    let result;
    client.subscribe(
      {
        query: `
mutation {
  createPost(input:{
    post:{
      headline:"Mutations now work? ${new Date().toISOString()}"
    }
  }) {
    post {
      id
      headline
    }
  }
}
`,
      },
      {
        next: data => (result = data),
        error: reject,
        complete: () => resolve(result),
      },
    );
  });

  console.log(util.inspect(result, { depth: 12 }));
})().catch(e => {
  console.error(e);
  process.exit(1);
});
```

PR #1562, opened 2021-11-30
feat: integrate `graphql-sse`

enisdenjo:
This PR hopes to integrate [`graphql-sse`](https://github.com/enisdenjo/graphql-sse) for transporting [GraphQL over Server-Sent Events](https://github.com/graphql/graphql-over-http/pull/163) behind the `eventSourceRoute`.

A few key takeaways:
- Doesn't mess with the schema update stream on `eventSourceRoute`, both can exist on the same route
- Added an `sse` option for enabling/disabling SSE _(disabled by default)_
- When `sse` is enabled, the `PUT` and `DELETE` methods are allowed through CORS in order to support the [HTTP/1 safe "single connection mode"](https://github.com/enisdenjo/graphql-sse/blob/master/PROTOCOL.md#single-connection-mode)
- The client can use the `X-GraphQL-Event-Stream` header for SSE endpoint discovery

### TODOs
- [ ] Disallow `PUT` and `DELETE` methods when PostGraphile is running on a HTTP/2 enabled server (or is behind a HTTP/2 proxy)
- [ ] [Implement the `postgraphile:validationRules` hook](https://github.com/enisdenjo/postgraphile/blob/145199eba580037ed7139fdb04ccaa09aed3c8a9/src/postgraphile/http/setupServerSentEvents.ts#L114-L130)


benjie:
I'd rather see this implemented as a plugin, you should be able to do various of the parts of it as such - adding the CLI flags/options, hooking `postgraphile:http:eventStreamRouteHandler` (and returning null) to add the event stream processing, hooking `postgraphile:http:graphqlRouteHandler` to add the `X-GraphQL-Event-Stream` header when watchPg is not set, etc.

There may be a few things you can't do this way _currently_, for example changes to CORS handling, and I'd be happy to see very small server hooks added to enable this functionality to be exposed.

In general I'd rather see this kind of major feature being a plugin first and then being merged into core once it's proven itself - it allows the plugin to go through major versions without the library itself having to do so.

enisdenjo:
Understood, thanks for the feedback! 

I'll go ahead and make a plugin out of it. If you agree, I'd be happy to do it under `@graphile-contrib/graphql-sse`, what do you think?

benjie:
Absolutely! https://github.com/graphile-contrib/graphql-sse

I will need to be the person that publishes the first version of the plugin, but after that I can give you rights on npm I think.

enisdenjo:
Brilliant! Thanks, Benjie.

I'll ping you when v1 is ready. 👍

enisdenjo:
Ping, this feature is now a plugin over at https://github.com/graphile-contrib/graphql-sse. 🚀

Issue #1455, opened 2021-03-25
Connecting to digital ocean managed database


ajaymore:
### Summary

I am getting below error

Error: self signed certificate in certificate chain
    at TLSSocket.onConnectSecure (_tls_wrap.js:1473:34)
    at TLSSocket.emit (events.js:311:20)
    at TLSSocket._finishInit (_tls_wrap.js:916:8)
    at TLSWrap.ssl.onhandshakedone (_tls_wrap.js:686:12)

Connection details

npx postgraphile -c "postgres://doadmin:secret@xxx-postgres-do-user-4864288-0.db.ondigitalocean.com:25060/apu-prod?ssl=1&sslrootcert=ca-certificate.crt" -s 'my_db' --retry-on-init-fail --watch --enhance-graphiql --dynamic-json

Tried few more combinations
npx postgraphile -c "postgres://doadmin:secret@xxx-postgres-do-user-4864288-0.db.ondigitalocean.com:25060/apu-prod?ssl=1&sslmode=require&sslrootcert=ca-certificate.crt" -s 'my_db' --retry-on-init-fail --watch --enhance-graphiql --dynamic-json

npx postgraphile -c "postgres://doadmin:secret@xxx-postgres-do-user-4864288-0.db.ondigitalocean.com:25060/apu-prod?sslmode=prefer" -s 'my_db' --retry-on-init-fail --watch --enhance-graphiql --dynamic-json

Below is connection string copied from digital ocean console.
postgresql://doadmin:secret@xxx-postgres-do-user-4864288-0.db.ondigitalocean.com:25060/apu-prod?sslmode=require

Am I missing something while running postgraphile?

benjie:
Try giving an absolute path to the cert rather than relative

benjie:
You can use sslmode=no-verify if you’re not concerned about verification

ajaymore:
> Try giving an absolute path to the cert rather than relative

Tried that as well. 

```
npx postgraphile -c 'postgres://secret@parallelearning-postgres-do-user-4864288-0.db.ondigitalocean.com:25060/apu-prod?sslmode=require&sslrootcert=/Users/mystical/Downloads/ca-certificate.crt' -s 'my_db' --retry-on-init-fail --watch --enhance-graphiql --dynamic-json
```

sslmode=no-verify does not work as digital ocean managed database strictly enforces SSL

benjie:
[...]

PR #1564, opened 2021-12-08
fix(cli): handling of SSL options from the connection string

maximsmol:
## Description

Addresses the CLI side of https://github.com/graphile/postgraphile/issues/1455

## Performance impact

None

## Security impact

None

## Checklist

- [X] I've added tests for the new feature, and `yarn test` passes.
- [X] I have added this feature to 'Pending' in the `RELEASE_NOTES.md` file (if one exists).
- [X] If this is a breaking change I've explained why.

### DIFF
--- a/src/postgraphile/cli.ts
+++ b/src/postgraphile/cli.ts
@@ -577,9 +577,7 @@ const coerce = (o: ReturnType<typeof parsePgConnectionString>): PoolConfig => {
         ? undefined
         : o.ssl.rejectUnauthorized == null
         ? !!o.ssl
-        : {
-            rejectUnauthorized: !!o.ssl.rejectUnauthorized,
-          },
+        : o.ssl,
     user: typeof o.user === 'string' ? o.user : undefined,
     database: typeof o.database === 'string' ? o.database : undefined,
     password: typeof o.password === 'string' ? o.password : undefined,
--- a/src/postgraphile/cli.ts
+++ b/src/postgraphile/cli.ts
@@ -575,9 +575,9 @@ const coerce = (o: ReturnType<typeof parsePgConnectionString>): PoolConfig => {
     ssl:
       o.ssl == null
         ? undefined
-        : o.ssl.rejectUnauthorized == null
+        : (o.ssl as any).rejectUnauthorized == null
         ? !!o.ssl
-        : o.ssl,
+        : (o.ssl as any),
     user: typeof o.user === 'string' ? o.user : undefined,
     database: typeof o.database === 'string' ? o.database : undefined,
     password: typeof o.password === 'string' ? o.password : undefined,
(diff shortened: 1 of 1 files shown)


src/postgraphile/cli.ts
```diff
@@ -572,7 +572,14 @@ const coerce = (o: ReturnType<typeof parsePgConnectionString>): PoolConfig => {
...
     application_name: o['application_name'] || undefined,
-    ssl: o.ssl != null ? !!o.ssl : undefined,
+    ssl:
+      o.ssl == null
+        ? undefined
+        : o.ssl.rejectUnauthorized == null
+        ? !!o.ssl
+        : {
+            rejectUnauthorized: !!o.ssl.rejectUnauthorized,
+          },
```
benjie, line 582: Maybe we should pass it through unmodified if it's an object?

```suggestion
    ssl:
      o.ssl == null
        ? undefined
        : typeof o.ssl === 'object' && !Array.isArray(o.ssl)
        ? o.ssl
        : !!o.ssl,
```

benjie (changes requested):

src/postgraphile/cli.ts
maximsmol: I think it depends on why `coerce` exists in the first place. Seems like it doesn't really do anything in most cases i.e. when `pg-connection-string` is well-behaved?

Passing the object as-is would obviously be more "future-proof" but it is equivalent as of now since `rejectUnauthorized` is the only thing that can be returned in this object.

Feel free to make a decision either way, I have no preference.

maximsmol: Any thoughts on the above? @benjie Happy to just commit your suggestion if we are fine with not validating the object

benjie: Leave it be. It’s been a hectic week; sorry!

benjie, line 582:
```suggestion
        ? !!o.ssl
        : o.ssl,
```

benjie (changes requested):

src/postgraphile/cli.ts
benjie: I went in a slightly different direction; if `rejectUnauthorized` is set then we handle the whole object. It's a bit weird but should be backwards-compatible (at least with _intent_).

benjie (approved):

benjie:
Ugh it doesn't pass the type checks. This'll need more work to unpick. I'll check it out and fix.

benjie:
As a summary of the issue; the TypeScript types indicate that the type from pg-connection-string is `string | boolean` but actually objects can come too; and to complicate things further `string` isn't an accepted argument for `ssl` option for `pg`.

benjie:
Released in postgraphile@4.12.8

PR #1577, opened 2022-01-22
How to use the tutorial

MeaningOfLights:
## Description

The Tutorial is great, I thought it necessary to "explicitly" tell people How To Use The System.

## Performance impact

No code change.

## Security impact

No code change.

## Checklist

- [ YES] I have detailed the feature in the relevant documentation.


benjie:
Sorry it's taking me a while to get around to reviewing this; I'm flat out on version 5 development right now. I will get to it in the next couple weeks :+1:

benjie (changes requested):
You currently have `PostGraphile` in places where `postgraphile` would be expected - e.g. the name of the npm package, the binary name, the name of database roles. Please note that `PostGraphile` and `postgraphile` each have their own relevant domains, the one that's _always_ incorrect is `Postgraphile`.

triptec:
If I were to do the requested changes, how would I go about doing that? Checkout @MeaningOfLights branch, add commits and create a new PR against main? Or is there a way to make a PR against a PR?

benjie:
I think you can raise a PR against @MeaningOfLights 's fork, which if merged would auto update this. But sending a PR direct would also be fine, @MeaningOfLights should still get credit by being in the git history 👍

PR #1597, opened 2022-03-18
chore: replace deprecated String.prototype.substr()

CommanderRoot:
## Description

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

## Performance impact

From past experience there is no performance impact due to switching to slice()

## Security impact

None as the returned values are still the same

## Checklist

### DIFF
--- a/scripts/make-assets
+++ b/scripts/make-assets
@@ -19,7 +19,7 @@ async function main() {
       tier = [];
       tiers.push(tier);
     } else if (line.startsWith('* ') || line.startsWith('- ')) {
-      tier.push(line.substr(2).replace(/\s?\(.*$/, ''));
+      tier.push(line.slice(2).replace(/\s?\(.*$/, ''));
     }
   }
   tiers.reverse();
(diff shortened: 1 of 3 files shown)


benjie (approved):
Thanks!

TIL the difference between `slice()` and `substring()` is in the handling of negative values: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#differences_between_substring_and_slice

(Mentioned because normally I'd move `substr` calls to be `substring` calls, but it seems that `slice` is actually preferable and is also shorter and compresses better due to overlap with arrays. No more substring for this person!)

PR #1616, opened 2022-04-13
refactor(ws): make sure operation is available before validating

enisdenjo:
## Description

[`graphql.validate`](https://github.com/graphql/graphql-js/blob/b38429f4d46b17f23d15406a2ca0d1eb201864b6/src/validation/validate.ts#L38-L46) will throw if its typings are not respected.

This PR adds pre-validation steps making sure the operation document, as well as the operation AST, is available.

Note that we don't do the same for the schema - not having it is a fatal error still.

## Performance impact

None.

## Security impact

None.

## Checklist


benjie (approved):
Looks good to me - thanks! :raised_hands:

benjie (approved):

Issue #1235, opened 2020-02-21
GraphiQL / Explain Improvements


benjie:
- [ ] Add option to enable analyze _on queries only_
- [ ] Add option to export as JSON rather than text (for import into https://tatiyants.com/pev/#/plans/new)
- [ ] Add button (or similar) to easily copy the SQL query or explain results

hyusetiawan:
maybe im just not used to it, but is it possible to improve the readability?
```
select to_json(dob) as "__dob"
from users as __local_0__
where (
  __local_0__.id = $1
) and (TRUE) and (TRUE)
```
is hard to parse what is going on, or perhaps there is a guide on how to understand it?

benjie:
Related:

- https://github.com/graphile/crystal-pre-merge/issues/388
- https://github.com/graphile/crystal-pre-merge/issues/413
- https://github.com/graphile/crystal/pull/1618

benjie:
@hyusetiawan FYI the SQL in V5 is significantly easier to understand:

# V4

![image](https://github.com/graphile/crystal/assets/129910/6f893e16-f11a-4c10-a5cc-159ff2f5040a)

# V5

![image](https://github.com/graphile/crystal/assets/129910/39cfc3cd-08de-4c66-b0ed-e63882ade352)

PR #1618, opened 2022-04-14
Add support for EXPLAIN options to GraphiQL

angelyan:
## Description

* Allows setting all available `EXPLAIN` options (analyze, verbose, costs, settings, buffers, wal, timing, summary, format) from the toolbar.
* Adds buttons to copy the SQL query and plan to the clipboard.

### Screenshots

"Options" and "Format" menu items are added next to "Explain".
Screenshot 2022-04-14 at 17-53-14 PostGraphile GraphiQL

The "Options" menu allows toggling of all boolean options. Note: An asterisk appears to the left of "TIMING" as it is enabled but it requires "ANALYZE" to be enabled as well.
Screenshot 2022-04-14 at 17-53-40 PostGraphile GraphiQL

The "Format" menu allows changing between all output formats (text, json, yaml, xml).
Screenshot 2022-04-14 at 17-55-35 PostGraphile GraphiQL

Fixes https://github.com/graphile/postgraphile/issues/1235.

## Performance impact

unknown

## Security impact

unknown

## Checklist

### DIFF
--- a/postgraphiql/src/components/PostGraphiQL.js
+++ b/postgraphiql/src/components/PostGraphiQL.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import GraphiQL from 'graphiql';
+import * as querystring from 'querystring';
 import {buildClientSchema, getIntrospectionQuery, getOperationAST, GraphQLObjectType, isType, parse} from 'graphql';
 import GraphiQLExplorer from 'graphiql-explorer';
 import StorageAPI from 'graphiql/dist/utility/StorageAPI';
@@ -89,9 +90,9 @@ function explainOptionRequiresAnalyze(option) {
 
 function buildExplainOptionsHeader(explainOptions) {
   const analyzeEnabled = explainOptions.analyze;
-  return Object.entries(explainOptions)
-    .filter(([option]) => !explainOptionRequiresAnalyze(option) || analyzeEnabled)
-    .map(([option, value]) => `${option}=${value}`).join(',')
+  const validOptions = Object.fromEntries(Object.entries(explainOptions)
+    .filter(([option]) => !explainOptionRequiresAnalyze(option) || analyzeEnabled));
+  return querystring.stringify(validOptions, ';', '=');
 }
 
 function writeToClipboard(content) {
@@ -129,7 +130,12 @@ class PostGraphiQL extends React.PureComponent {
 
   parseFromStorage(key) {
     const value = this._storage.get(key);
-    return value ? JSON.parse(value) : undefined;
+    try {
+      return value ? JSON.parse(value) : undefined;
+    } catch (e) {
+      console.warn(`Failed to parse key '${key}' from storage: ${value}`)
+      return undefined;
+    }
   }
 
   restartRequested = false;
@@ -398,9 +404,8 @@ class PostGraphiQL extends React.PureComponent {
           Accept: 'application/json',
           'Content-Type': 'application/json',
           ...(this.state.explain && POSTGRAPHILE_CONFIG.allowExplain
-            ? { 'X-PostGraphile-Explain': 'on' }
+            ? { 'X-PostGraphile-Explain': `on;${buildExplainOptionsHeader(this.state.explainOptions)}` }
             : null),
-          'X-PostGraphile-Explain-Options': buildExplainOptionsHeader(this.state.explainOptions)
         },
         extraHeaders,
       ),
--- a/postgraphiql/src/components/PostGraphiQL.js
+++ b/postgraphiql/src/components/PostGraphiQL.js
@@ -1,12 +1,11 @@
 import React from 'react';
 import GraphiQL from 'graphiql';
-import { getOperationAST, parse } from 'graphql';
+import {buildClientSchema, getIntrospectionQuery, getOperationAST, GraphQLObjectType, isType, parse} from 'graphql';
 import GraphiQLExplorer from 'graphiql-explorer';
 import StorageAPI from 'graphiql/dist/utility/StorageAPI';
 import './postgraphiql.css';
-import { buildClientSchema, getIntrospectionQuery, isType, GraphQLObjectType } from 'graphql';
-import { SubscriptionClient } from 'subscriptions-transport-ws';
-import { createClient } from 'graphql-ws';
+import {SubscriptionClient} from 'subscriptions-transport-ws';
+import {createClient} from 'graphql-ws';
 import formatSQL from '../formatSQL';
 
 const defaultQuery = `\
@@ -81,7 +80,19 @@ const STORAGE_KEYS = {
   SAVE_HEADERS_TEXT: 'PostGraphiQL:saveHeadersText',
   HEADERS_TEXT: 'PostGraphiQL:headersText',
   EXPLAIN: 'PostGraphiQL:explain',
+  EXPLAIN_OPTIONS: 'PostGraphiQL:explainOptions',
 };
+
+function explainOptionRequiresAnalyze(option) {
+  return ['wal', 'timing'].includes(option);
+}
+
+function buildExplainOptionsHeader(explainOptions) {
+  const analyzeEnabled = explainOptions.analyze;
+  return Object.entries(explainOptions)
+    .filter(([option]) => !explainOptionRequiresAnalyze(option) || analyzeEnabled)
+    .map(([option, value]) => `${option}=${value}`).join(',')
+}
 
 /**
  * The standard GraphiQL interface wrapped with some PostGraphile extensions.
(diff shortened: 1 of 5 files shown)


benjie:
This is very cool! One significant concern that I have currently is that if we allow `explain analyze` then for mutations the operation will run twice (once for the `explain analyze` and once again for the mutation itself) - and this is the main reason that I shyed away from adding that option. I'd like this to be addressed before anything that allows `explain analyze` to be enabled, but I don't know how best to do this. We could just disable it for mutation operations on the server side, but I'm not sure if this is sufficiently obvious to the user?

postgraphiql/src/components/PostGraphiQL.js
```diff
@@ -1,12 +1,11 @@
 import React from 'react';
 import GraphiQL from 'graphiql';
-import { getOperationAST, parse } from 'graphql';
+import {buildClientSchema, getIntrospectionQuery, getOperationAST, GraphQLObjectType, isType, parse} from 'graphql';
```
benjie, line 3: I do not understand why prettier is not formatting this file in the way I'd expect. It doesn't seem to be in the `prettierignore` file, and it doesn't seem to be excluded from the prettier glob

https://github.com/graphile/postgraphile/blob/55bff41460b113481c8161ef8f178f5af0a17df3/package.json#L48

And yet, CI passes. Very strange. I will have to investigate at some point.

angelyan:
> One significant concern that I have currently is that if we allow `explain analyze` then for mutations the operation will run twice

What about running the `EXPLAIN` in a transaction and then rolling it back?

benjie:
Other than leaking through things like sequences (`serial` type, for example) I can’t think of a particularly good reason not to do that. We’d have to know if we’re in a transaction or not: if we are we should always savepoint, and if not then it’s not a mutation so no need to worry. Good idea 👍

angelyan:
@benjie I added the savepoint. Could you do another pass when you get a chance?

src/postgraphile/withPostGraphileContext.ts
```diff
@@ -549,18 +550,34 @@ export function debugPgClient(pgClient: PoolClient, allowExplain = false): PoolC
...
+                    .then((data: any) => pgClient[$$pgClientOrigQuery]
+                      .call(this, 'rollback to savepoint postgraphile_explain')
+                      .then(() => data)),
+                  // Failed to create savepoint - we are not in a transaction, which means this is a query. No need to
+                  // rollback the EXPLAIN.
+                  () => pgClient[$$pgClientOrigQuery].call(this, explain, values)
+                )
+                .then((data: any) => data.rows)
+                // swallow errors during explain
+                .catch(() => null);
```
benjie, line 574: This isn't safe because our queries can occur in parallel, and thus the rollback might be issued out of order (or there may even be concurrent rollbacks for the same savepoint!)

Instead you must ensure that they are all added in the same tick so that they get issued in order without a chance for other (parallel) queries to be inserted in between:

```js
// IMPORTANT: the following 3 lines MUST BE CALLED SYNCHRONOUSLY.
pgClient[$$pgClientOrigQuery].query('savepoint postgraphile_explain');
const explainResultPromise = pgClient[$$pgClientOrigQuery].query(explain, values);
pgClient[$$pgClientOrigQuery].query('rollback savepoint postgraphile_explain');

// Now us `explainResultPromise` however you want.
```

benjie (changes requested):

angelyan:
@benjie this is ready for another pass!

postgraphiql/src/components/PostGraphiQL.js
```diff
@@ -101,12 +116,22 @@ class PostGraphiQL extends React.PureComponent {
...
     headersTextValid: true,
-    explorerIsOpen: this._storage.get('explorerIsOpen') === 'false' ? false : true,
+    explorerIsOpen: this._storage.get('explorerIsOpen') !== 'false',
     haveActiveSubscription: false,
     socketStatus: POSTGRAPHILE_CONFIG.websockets === 'none' ? null : 'pending',
   };
 
+  parseFromStorage(key) {
+    const value = this._storage.get(key);
+    return value ? JSON.parse(value) : undefined;
```
benjie, line 132: This should probably be wrapped with a `try/catch`

postgraphiql/src/components/PostGraphiQL.js
```diff
@@ -375,6 +400,7 @@ class PostGraphiQL extends React.PureComponent {
           ...(this.state.explain && POSTGRAPHILE_CONFIG.allowExplain
             ? { 'X-PostGraphile-Explain': 'on' }
             : null),
+          'X-PostGraphile-Explain-Options': buildExplainOptionsHeader(this.state.explainOptions)
```
benjie, line 403: I wonder if we should just replace the `{ 'X-PostGraphile-Explain': 'on' }` header with
```
{ 'X-PostGraphile-Explain': `on; ${buildExplainOptionsHeader(this.state.explainOptions)}` }
```

?

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -110,6 +115,34 @@ const isPostGraphileDevelopmentMode = process.env.POSTGRAPHILE_ENV === 'developm
...
+      default:
+        if (['true', 'false'].includes(value)) {
+          explainOptions[option] = value === 'true';
+        } else {
+          console.warn(`Ignoring invalid value '${value}' for explain option '${option}'.`);
+        }
+    }
+  }
+  return explainOptions;
+}
```
benjie, line 144: Do we really need custom parsing for this? I feel like we should just use `querystring` or JSON or similar.

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -150,6 +183,7 @@ function withPostGraphileContextFromReqResGenerator(
         jwtToken,
         pgSettings,
         explain: allowExplain && req.headers['x-postgraphile-explain'] === 'on',
+        explainOptions: parseExplainOptions(req),
```
benjie, line 186: Let's merge explain/explainOptions together. We don't need to support any legacy here since PostGraphile and PostGraphiQL are deeply integrated and tied to the exact same versions, so unified changes are fine - no need for backwards compatibility.

```suggestion
        explain: allowExplain && parseExplainOptions(req),
```

src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
```diff
@@ -1175,6 +1209,7 @@ function addCORSHeaders(res: PostGraphileResponse): void {
       'Content-Length',
       // For our 'Explain' feature
       'X-PostGraphile-Explain',
+      'X-PostGraphile-Explain-Options',
```
benjie, line 1212: Thinking about it, I'm very keen that we don't introduce another header because it means people may have to rewrite their nginx, CORS and similar config.

benjie (changes requested):
Excellent work - looking good! Ideally you should force a transaction for queries that involve `EXPLAIN` so we don't get the errors from `SAVEPOINT` / `ROLLBACK` (which I'm almost certain I'll start hearing about in a few months time...); but if you can't figure out how to do that I'm happy to do that as part of the final cleanup before merge 👍

angelyan:
@benjie I addressed everything you commented except for forcing the transaction. Couldn't figure out a nice way to do that, so it's all yours.

benjie:
Thanks! Just a heads up that it's going to be a little while before I get around to re-reviewing this, I'm focussed on V5 work currently. I'm excited though!

src/postgraphile/withPostGraphileContext.ts
```diff
@@ -532,15 +539,38 @@ export function debugPgClient(pgClient: PoolClient, allowExplain = false): PoolC
             const query = a && a.text ? a.text : a;
             const values = a && a.text ? a.values : b;
             if (query.match(/^\s*(select|insert|update|delete|with)\s/i) && !query.includes(';')) {
+              // Build the EXPLAIN command
+              let explainCommand = 'explain';
+              const explainOptionClauses = Object.entries(pgClient._explainOptions ?? {}).map(
+                ([option, value]) => `${option} ${value}`,
```
benjie, line 545: `_explainOptions` isn't safe to use like this, this opens PostGraphile up to SQL injection attacks.

benjie (changes requested):
So sorry, this slipped off of my TODO list (the notification got removed from my GitHub notifications somehow), I only just noticed it again :sweat: 

I don't have permission to push to your `master` branch so I've applied a number of minor fixes and tweaks to the `explain` branch: https://github.com/graphile/crystal/tree/explain

However, the main reason that this can't be merged currently is that it looks ripe for SQL injection attacks. Inserting user-submitted content (sent from a header) directly into an SQL query's raw text is a recipe for disaster. We need server-side validation of the values, and I'd rather see an exhaustive "if options.analyze then sql.push('analyze')" for each option that makes it extremely explicit exactly what text is being pushed into the explain query.

angelyan:
Hi @benjie. I will not continue working on this PR. I'm happy to hand it off to you if you want. Otherwise, I'll be closing it and deleting my fork.

benjie:
Hi @angelyan; no worries and sorry I dropped the ball here on the review time, I really don't know how the notification disappeared from my inbox. I'll be adding something like this to PostGraphile V5 and your code now exists on the `explain` branch so you can feel free to close and delete the fork. All the best!

angelyan:
No worries. Thank you!

PR #1640, opened 2022-06-13
Remove jwt not supported claims names

jeaneric:
Exclude jwt claims beginning with "_" as they are not supported by postgresql

## Description

Problem when reusing a token that have a claim beginning by an underscore. Fix by just ignoring those claims

## Performance impact

None

## Security impact

None

## Checklist

### DIFF
--- a/src/postgraphile/withPostGraphileContext.ts
+++ b/src/postgraphile/withPostGraphileContext.ts
@@ -430,7 +430,8 @@ async function getSettingsForPgClientTransaction({
   // If we have some JWT claims, we want to set those claims as local
   // settings with the namespace `jwt.claims`.
   for (const key in jwtClaims) {
-    if (Object.prototype.hasOwnProperty.call(jwtClaims, key)) {
+    // Exclude claims beginning with "_" as they are not supported by postgresql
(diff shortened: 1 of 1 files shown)


benjie:
Hi, thanks for contributing! Unfortunatly this would be a breaking change for anyone using an older version of Postgres which _does_ in fact support these names:

```
[postgres] # select set_config('jwt.claims._foo'::text, '27'::text, false);
┌────────────┐
│ set_config │
├────────────┤
│ 27         │
└────────────┘
(1 row)

Time: 0.322 ms
[postgres] # select current_setting('jwt.claims._foo'::text);
┌─────────────────┐
│ current_setting │
├─────────────────┤
│ 27              │
└─────────────────┘
(1 row)

Time: 0.375 ms
[postgres] # 
```

I considered doing a postgres version check and enabling the functionality only in incompatible Postgres versions, but doing so would mean that you could test something locally on PG12 and think it's fine, and then push it to production with PG15 and not know that some of your claims were being silently ignored - this is unacceptable, so we cannot follow this route either.

Instead you must either: 1. make sure your JWT is an acceptable shape for PostGraphile's JWT handling (including Postgres' needs), or 2. implement your own JWT handling by using PostGraphile in library mode and putting a JWT middleware in front of it.

I'm going to close this, because I think there's nothing we can do about it in PostGraphile itself really.

PR #1665, opened 2022-08-25
Exports `extendedFormatError`

arcanis:
## Description

If we set `handleErrors`, we can't set `extendedErrors`. But since the extended formatting function isn't exported, we have to reimplement its logic.

## Performance impact

n/a

## Security impact

n/a

## Checklist

- n/a I've added tests for the new feature, and `yarn test` passes.
- n/a If this is a breaking change I've explained why.


benjie (changes requested):
You also need an export in `src/index.ts`; e.g. https://github.com/graphile/postgraphile/blob/4a809aa63e3e97cc264ed1e8d92b6b5eb5894dbb/src/index.ts#L28-L41

benjie (approved):
I've made the changes; had to do it through the GitHub interface because for some reason git wouldn't let me push to your branch.

PR #1678, opened 2022-11-03
chore: update to Node 16 LTS in CI & published docker images

Jackbennett:
## Description

The issue is with Node 12 being so old I had TLS connection issues that would hang after startup and silently use all available memory. The node 12 image is using alpine3.15 anyway so this PR keeps alpine the same. While changing versions just moved any reference to node to the current LTS.

## Performance impact

Would expect an improvement.

## Security impact

Published docker images would be using a supported version of Node.
Potentially could remove node 8, 10 and 12 from the matrix as they go out of support.

## Checklist

## Comments

All I've tested is manually installing node 17 within the existing container and things seemed to work fine. Hopefully the CI tests can rundown any issues.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,10 +15,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Use Node.js 14.x
+      - name: Use Node.js 16.x
         uses: actions/setup-node@v1
         with:
-          node-version: 14.x
+          node-version: 16.x
       - run: yarn --immutable --ignore-engines
 
       - name: "Check Code Format"
@@ -42,7 +42,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        postgres-version: 
+        postgres-version:
           - 9.4
           - 10
           - 11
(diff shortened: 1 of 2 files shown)


benjie:
Seems Node 18 and PG 15 are failing across the board, please use Node 16 and limit the scope of this change as much as possible (i.e. don't mess with the Postgres versions). If you'd like to follow up with a PR adding Node 18 support and another PR adding PG15 support to CI those would be great ❤️

Jackbennett:
I did a bit of digging and looks like Webpack 4 -> 5 is required to use node 18, so starting to do that moves quite a lot of package versions. I'm not sure how you want to handle that.

PG15 might be as simple as something in the test init script.  Changed the commits to node 16 only.

benjie:
Let's skip the Node 18 stuff for now; hopefully V5 isn't too far out and that supports Node 18 OOTB already. I think supporting Node 8 -> Node 16 is a wide enough range for V4 😅 

PG15 would be great as a separate PR if you're interested ❤️

Jackbennett:
I'll take a look at pg15.

If you drop node <14 it looks like you can remove whatever the turbo-js stuff is and be faster all the time? Nobody should be using unspoorted node much less as a public api surface.

Tests passed so please let me know if there's anything else you want on this PR, it didn't seem much else of the checklist applied.

PR #1713, opened 2023-04-05
Allow pgDefaultRole with pgSettings(req)

mvnetbiz:
## Description
Remove error which stops use of both `pgDefaultRole` and `pgSettings(req)` (originally added in #606)

Role is already set in order
1. pgDefaultRole: https://github.com/graphile/postgraphile/blob/2e89f875050ecfc554cfff6b96b9a6335abf7bcc/src/postgraphile/withPostGraphileContext.ts#L320
2. JWT role claim: https://github.com/graphile/postgraphile/blob/2e89f875050ecfc554cfff6b96b9a6335abf7bcc/src/postgraphile/withPostGraphileContext.ts#L387
3. pgSettings(req) role key: https://github.com/graphile/postgraphile/blob/2e89f875050ecfc554cfff6b96b9a6335abf7bcc/src/postgraphile/withPostGraphileContext.ts#L416

Allows use of both `pgDefaultRole` and `pgSettings(req)` options. I am using `pgSettings(req)`, but I would like to use the default JWT role claim parsing, as well have have a default role to fall back to in case of no JWT.

## Performance impact

none

## Security impact

unknown

## Checklist

### DIFF
--- a/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
+++ b/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts
@@ -222,11 +222,6 @@ export default function createPostGraphileHttpRequestHandler(
 
   const origGraphiqlHtml = pluginHook('postgraphile:graphiql:html', baseGraphiqlHtml, { options });
 
-  if (pgDefaultRole && typeof pgSettings === 'function') {
-    throw new Error(
-      'pgDefaultRole cannot be combined with pgSettings(req) - please remove pgDefaultRole and instead always return a `role` key from pgSettings(req).',
-    );
-  }
   if (
     pgDefaultRole &&
     pgSettings &&


mvnetbiz:
I didn't notice before, but if I return other keys besides 'role' from pgSettings(req), for some reason they get ignored if a JWT is present.

benjie:
No; these errors are there for a reason - to resolve an ambiguity that can cause users to set up their system insecurely due to not understanding how it works.

1. We set role = pgDefaultRole (https://github.com/graphile/postgraphile/blob/263ba7477bc2133eebdf89d29acd0460e58501ec/src/postgraphile/withPostGraphileContext.ts#L320)
2. We override this role with the role from the JWT, if present (https://github.com/graphile/postgraphile/blob/263ba7477bc2133eebdf89d29acd0460e58501ec/src/postgraphile/withPostGraphileContext.ts#L387)
3. We override this role with the role from `pgSettings`, if present (https://github.com/graphile/postgraphile/blob/263ba7477bc2133eebdf89d29acd0460e58501ec/src/postgraphile/withPostGraphileContext.ts#L416)

If `pgSettings` (object) contains a `role` key, then `pgDefaultRole` will _never be used_. I see you did not remove this code. However, the same is true when `pgSettings` is a function - it's very common instead of using `pgSettings: {...}` to use `pgSettings: (req) => ({...})` to allow you to add request-based details to a pgSettings object. But depending on your implementation of `pgSettings` (which we do not know!) if you always return a `role` key then `pgDefaultRole` will never be used. Users have tripped over this in the past, which is why these errors were added.

You should either use `pgDefaultRole`, `pgSettings` as an object, and the built in JWT parsing; _or_ you should use `pgSettings` as a function to handle all of this, with your own JWT parsing (or, _even better_, [don't use JWTs at all](https://sprout.io/Benjie/13938-why-i-dont-use-jwt-for-sessions)!).

mvnetbiz:
Understood. I am curious though, what did "In a later release we might let these options co-exist again (but the right way around!)" mean? Is this PR "the right way around" as understood at the time, but due to the ambiguity you described is no longer in consideration? Or does it mean something else entirely? Either way, I'll do as suggested. Thanks!

benjie:
"The right way around" would be processing the JWT, then feeding the result of that to pgSettings and allowing pgSettings to selectively override things (and add more), and then if there is still no `role` adding the `pgDefaultRole`. We cannot do that in the V4 release cycle (it would be a breaking change) and V5 is configured in a way that eradicates this problem.

PR #1714, opened 2023-05-16
Add Postgres 15 to CI

benjie:


jmealo:
I opened #1716 which uses the latest commit for `wal2json`. This may not correspond to a published version, but, I figured it's a good place to start.

benjie:
Replaced by #1716

PR #1756, opened 2023-09-22
fix(dataplan-pg): type restrictions

### LINKED ISSUE #1755: Broken type helper (GetPgResourceRelations)
### Summary

Various type helpers appear to be broken (`GetPgResourceRelations`) for example always returns `any` even when supplying a fully typed `PgResource` with explicit relations. I would advice to use `infer` instead of property access to prevent type erasure.

Below works:

```typescript
type GetRegistry<U> = U extends PgResource<any, any, any, any, infer R>
  ? R
  : never
type GetRelations<U> = U extends PgRegistry<any, any, infer R> ? R : never
type GetCodec<U> = U extends PgResource<any, infer C, any, any, any>
  ? C
  : never
type GetName<U> = U extends PgCodec<infer N, any, any, any, any, any, any>
  ? N
  : never
type GetPgResourceRelations<U> = U extends PgResource<
  any,
  any,
  any,
  any,
  any
>
  ? GetRelations<GetRegistry<U>>[GetName<GetCodec<U>>]
  : never
```

Whereas this doesnt:

```typescript
type GetPgResourceRelations<TResource extends PgResource<any, any, any, any, any>> = TResource["registry"]["pgRelations"][TResource["codec"]["name"]];
```

wesselvdv:
## Description

fixes #1755 

## Checklist

### DIFF
--- a/grafast/dataplan-pg/src/interfaces.ts
+++ b/grafast/dataplan-pg/src/interfaces.ts
@@ -760,41 +760,85 @@ export interface PgRegistry<
   };
 }
 
-export type GetPgRegistryCodecs<TRegistry extends PgRegistry<any, any, any>> =
-  TRegistry["pgCodecs"];
-
-export type GetPgRegistrySources<TRegistry extends PgRegistry<any, any, any>> =
-  TRegistry["pgResources"];
-
+export type GetPgRegistryCodecs<U> = U extends PgRegistry<infer C, any, any>
+  ? C
+  : never;
+export type GetPgRegistrySources<U> = U extends PgRegistry<any, infer R, any>
+  ? R
+  : never;
+export type GetPgRegistryRelations<U> = U extends PgRegistry<any, any, infer R>
+  ? R
+  : never;
+export type GetPgCodecName<U> = U extends PgCodec<
+  infer N,
+  any,
+  any,
+  any,
+  any,
+  any,
+  any
+>
+  ? N
+  : never
 export type GetPgRegistryCodecRelations<
   TRegistry extends PgRegistry<any, any, any>,
   TCodec extends PgCodec<any, any, any, any, any, any, any>,
-> = TRegistry["pgRelations"][TCodec["name"]];
-
-export type GetPgCodecAttributes<
-  TCodec extends PgCodec<any, any, any, any, any, any, any>,
-> = TCodec extends PgCodec<any, infer UAttributes, any, any, any, any, any>
-  ? UAttributes extends undefined
-    ? never
-    : UAttributes
-  : PgCodecAttributes;
-
-export type GetPgResourceRegistry<
-  TResource extends PgResource<any, any, any, any, any>,
-> = TResource["registry"];
-
-export type GetPgResourceCodec<
-  TResource extends PgResource<any, any, any, any, any>,
-> = TResource["codec"];
-
-export type GetPgResourceAttributes<
-  TResource extends PgResource<any, any, any, any, any>,
-> = GetPgCodecAttributes<TResource["codec"]>;
-
-export type GetPgResourceRelations<
-  TResource extends PgResource<any, any, any, any, any>,
-> = TResource["registry"]["pgRelations"][TResource["codec"]["name"]];
-
-export type GetPgResourceUniques<
-  TResource extends PgResource<any, any, any, any, any>,
-> = TResource["uniques"];
+> = GetPgRegistryRelations<TRegistry>[GetPgCodecName<TCodec>]
+export type GetPgCodecAttributes<U> = U extends PgCodec<
+  any,
+  infer A,
+  any,
+  any,
+  any,
+  any,
+  any
+>
+  ? A
+  : never;
+export type GetPgResourceRegistry<U> = U extends PgResource<
+  any,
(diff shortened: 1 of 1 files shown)


wesselvdv:
Applying this fix allows you to get actual type completion on a typed `PgSelectStep<A>` for `singleRelation` where the input of `singleRelation` is only valid if it exists as a relation in the associated `PgRegistry` for type `A`

benjie:
Doesn't this _reduce_ type safety by removing the constraints on the generics though?

wesselvdv:
That's true, but the current constraints are too strict causing it to not be workable. I can't pass a object that can satisfy the constraints without causing it to be useless because I have to union `"asd"` with `string` keys causing it to be generic again.

wesselvdv:
I also see that the use of `undefined` as bottom type is problematic in many cases, I think it's better to use `never` in such cases. And you don't need to explicitly indicate that it might be `never` in a union.

benjie:
... but it's not "never" - it's literally that those things can be undefined. A function accepts an array of parameters (even an empty array), a table does _not_ accept an array of parameters - the parameters are undefined - we should be able to differentiate this at a type level, otherwise we can't tell the difference between function-like and table-like... can we?

:thinking: _Thinking pause..._

Maybe you're right... a table "never" accepts parameters... It's a bit weird because `foo 
 never = foo` for all `foo`... but I think we can still check it:

https://www.typescriptlang.org/play?#code/C4TwDgpgBAYlC8UB2BXAtgIwgJygH2QgDccBuAKAHpKpaA9AfnNEigHEFYoIAPYCJABMAzoRK4GUYNhTQAXFABmAQwA2wiBWq0ojZuGgAJTkmJkqNekxbQAkp2O9+Q0afFRJ02VAUr1mix09IA

My main concern is that this might look confusing to the consumer - they might think parameters is _always_ required (since `| never` evaporates) and thus pass an empty array when they should be passing `never`. Currently the `| undefined` is very explicit that that's an option. Having said that, if it makes working with the types easier and doesn't actually break anything, I'm open to it.

benjie:
Is it the `[name in string]` causing the issues for you; would `[name: string]` be any better? I'm not sufficiently knowledgeable about TypeScript to fully understand the difference.

wesselvdv:
> Is it the `[name in string]` causing the issues for you; would `[name: string]` be any better? I'm not sufficiently knowledgeable about TypeScript to fully understand the difference.

I am playing around with another one:

```typescript
export interface PgRegistry<
  TCodecs extends Record<
    keyof TCodecs,
    PgCodec<any, PgCodecAttributes, any, any, any, any, any>
  >,
  TResourceOptions extends Record<
    keyof TResourceOptions,
    PgResourceOptions<
      string,
      TCodecs[keyof TCodecs],
      ReadonlyArray<PgResourceUnique<PgCodecAttributes>>,
      ReadonlyArray<PgResourceParameter>
    >
  >,
  TRelations extends Record<
    keyof TRelations,
    Record<
      keyof TRelations[keyof TRelations],
      PgCodecRelationConfig<
        any,
        PgCodec<string, PgCodecAttributes, any, any, never, any, never>,
        PgResourceOptions<any, PgCodecWithAttributes<any>, any, any>
      >
    >
  >,
> {
  pgCodecs: TCodecs
  pgResources: {
    [name in keyof TResourceOptions]: TResourceOptions[name] extends PgResourceOptions<
      infer UName,
      infer UCodec,
      infer UUniques,
      infer UParameters
    >
      ? PgResource<
          UName,
          UCodec,
          UUniques,
          UParameters,
          PgRegistry<TCodecs, TResourceOptions, TRelations>
        >
      : never
  }
  pgRelations: {
    [codecName in keyof TRelations]: {
      [relationName in keyof TRelations[codecName]]: Expand<
        Omit<TRelations[codecName][relationName], "remoteResourceOptions"> & {
          remoteResource: TRelations[codecName][relationName] extends {
            remoteResourceOptions: PgResourceOptions<
              infer UName,
              infer UCodec,
              infer UUniques,
              infer UParameters
            >
          }
            ? PgResource<
                UName,
                UCodec,
                UUniques,
                UParameters,
                PgRegistry<TCodecs, TResourceOptions, TRelations>
              >
            : never
        }
      >
    }
  }
}
```

benjie:
`Foo extends Record<keyof Foo, ...>`

But the... wait the key... the... of... of the... ... what?! :exploding_head:

benjie:
I appreciate your work on this @wesselvdv; I'm going to convert this PR to a "draft" to indicate that research is still in progress. You can probably see I tried a bunch of approaches and the current one is the best I could come up with, but I still very much do not like it, and it requires too much manual `as ...` casting - not to mention not accepting perfectly valid things that it should accept.

wesselvdv:
> I appreciate your work on this @wesselvdv; I'm going to convert this PR to a "draft" to indicate that research is still in progress. You can probably see I tried a bunch of approaches and the current one is the best I could come up with, but I still very much do not like it, and it requires too much manual `as ...` casting - not to mention not accepting perfectly valid things that it should accept.

I closed this one in favour for my other branch, for which I have created a new PR.https://github.com/graphile/crystal/pull/1784

PR #1776, opened 2023-10-01
Fix docs for configuring list behavior in preset

adrianmfi:
## Description

The field to use in the preset file for configuring list behaviour is `defaultBehavior` instead of `globalBehavior`,
as described here https://postgraphile.org/postgraphile/next/behavior#global-default-behavior

## Performance impact
None 

## Security impact
None

## Checklist


benjie (approved):
Thanks! We use `defaultBehavior` in presets and `globalBehavior` in plugins and I frequently get them confused. I'm considering removing `globalBehavior` entirely now we have an entity behavior system.

PR #1777, opened 2023-10-02
Added lambda adapter

innermatrix:
## Description

Lambda adapter for Grafserv (no WebSockets support)

Todo: validate that the example code works outside of Postgraphile. 

## Performance impact

N/A

## Security impact

N/A

## Checklist


grafast/grafserv/package.json
```diff
@@ -96,6 +98,7 @@
     }
   },
   "devDependencies": {
+    "@types/aws-lambda": "^8.10.122",
```
benjie, line 101: Is it right that the types are v8 but the library is v1?

grafast/grafserv/examples/example-lambda.mjs
```diff
@@ -0,0 +1,10 @@
+import { grafserv } from "grafserv/lambda";
```
benjie, line 1: Normally we version our adaptors; is it likely we would need this with lambda? In general I'd err on the side of adding a version (even `v1`) just to leave space for future changes.

grafast/grafserv/package.json
```diff
@@ -70,10 +70,12 @@
   "homepage": "https://github.com/graphile/crystal/tree/main/grafast/grafserv",
   "dependencies": {
     "@graphile/lru": "workspace:^",
+    "aws-lambda": "^1.0.7",
     "debug": "^4.3.4",
     "eventemitter3": "^5.0.1",
     "graphile-config": "workspace:^",
     "graphql-ws": "^5.14.0",
+    "modern-async": "^1.1.4",
```
benjie, line 78: `aws-lambda` and `modern-async` should **not** be dependencies - only the lambda adaptor needs them, we shouldn't install them for everyone else. Please move them to `peerDependencies` and make them optional via `peerDependenciesMeta`.

grafast/grafserv/src/servers/lambda/index.ts
```diff
@@ -0,0 +1,132 @@
...
+  APIGatewayProxyResult,
+  Context as LambdaContext,
+} from "aws-lambda";
+import type {
+  GrafservConfig,
+  RequestDigest} from "grafserv";
+import {
+  GrafservBase,
+  processHeaders
+} from "grafserv";
```
benjie, line 13: Self-imports can cause issues for various pieces of tooling (that haven't been implemented to fully support Node's resolution); please use relative paths instead

```suggestion
import type { GrafservConfig, RequestDigest } from "../../interfaces.js";
import { GrafservBase } from "../../core/base.js";
import { processHeaders } from "../../utils.js";
```

grafast/grafserv/src/servers/lambda/index.ts
```diff
@@ -0,0 +1,132 @@
...
+import { toArray } from "modern-async";
+
+export class LambdaGrafserv extends GrafservBase {
+  createHandler() {
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    const serv = this;
+    return async function (
+      event: APIGatewayEvent,
+      context: LambdaContext,
+    ): Promise<APIGatewayProxyResult> {
```
benjie, line 23: Rather than using a `serv = this`, just use an arrow function:

```suggestion
    return async (
      event: APIGatewayEvent,
      context: LambdaContext,
    ): Promise<APIGatewayProxyResult> => {
```

grafast/grafserv/src/servers/lambda/index.ts
```diff
@@ -0,0 +1,132 @@
...
+            text: event.body ?? "",
+          };
+        },
+        requestContext: {
+          lambdaCtx: { event, context },
+        },
+        preferJSON: true,
+      };
+
+      const response = await serv.processLambdaRequest(event, context, request);
```
benjie, line 54:
```suggestion
      const response = await this.processLambdaRequest(event, context, request);
```

grafast/grafserv/src/servers/lambda/index.ts
```diff
@@ -0,0 +1,132 @@
...
+export class LambdaGrafserv extends GrafservBase {
+  createHandler() {
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    const serv = this;
+    return async function (
+      event: APIGatewayEvent,
+      context: LambdaContext,
+    ): Promise<APIGatewayProxyResult> {
+      const version = event.requestContext.protocol.match(
+        "^HTTP/(?<major>[0-9]+).(?<minor>[0-9]+)$",
```
benjie, line 25: Let's escape the `.` and use a regex literal rather than casting a string:

```suggestion
        /^HTTP\/(?<major>[0-9]+)\.(?<minor>[0-9]+)$/,
```

benjie (changes requested):
This is a great first draft! Please try and follow conventions in the other adaptors, such as putting the declaration merged types near the top of the file, using versioning for the adaptor, etc. Thanks for doing this!

I'd also like to see a cut down version of the lambda handler that _only_ serves GraphQL and drops the GraphiQL/event stream/etc stuff.

Adding details to the documentation about how to configure the lambda via the AWS console or similar would also be great, but isn't required to get this merged.

:raised_hands:

grafast/grafserv/package.json
innermatrix: I am not a huge fan of the idea of copy-pasting type declarations from a 3rd party package, since a peer dep will accomplish all the same things wrt isolating graphile from dep sprawl, while making it easier to track updates in the 3rd party package. So if it's all the same to you, I'll make it a peer dep and keep the import.

grafast/grafserv/src/servers/lambda/index.ts
```diff
@@ -0,0 +1,132 @@
...
+          ) as Record<string, string>;
+        },
+        getBody() {
+          return {
+            type: "text",
+            text: event.body ?? "",
+          };
+        },
+        requestContext: {
+          lambdaCtx: { event, context },
```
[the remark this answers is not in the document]
innermatrix: I was mislead 😅

```
     * It's anticipated this will be expanded via declaration merging, e.g. if
     * your server is Koa then a `koaCtx` might be added.
```

grafast/grafserv/src/servers/lambda/index.ts
```diff
@@ -0,0 +1,132 @@
...
+
+        case "bufferStream": {
+          const { statusCode, headers, bufferIterator } = response;
+
+          // TODO: implement low-latency path using awslambda.streamifyResponse
+          // https://aws.amazon.com/blogs/compute/introducing-aws-lambda-response-streaming/
+          return {
+            statusCode,
+            headers,
+            body: Buffer.concat(await toArray(bufferIterator)).toString(),
```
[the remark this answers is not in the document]
innermatrix: Idk does it even merit a function?

grafast/grafserv/src/servers/lambda/index.ts
innermatrix: Ya, leftover from developing this inside my own package before moving it out

grafast/grafserv/package.json
benjie: I meant that you already have `@types/aws-lambda` so you shouldn't _also_ need `aws-lambda` since we're only using the types?

grafast/grafserv/package.json
innermatrix: `@types/aws-lambda` is right and `aws-lambda` is a totally unrelated 3rd party package smh

https://docs.aws.amazon.com/lambda/latest/dg/typescript-handler.html

grafast/grafserv/src/servers/lambda/index.ts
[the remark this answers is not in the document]
innermatrix: https://github.com/graphile/crystal/pull/1783

innermatrix:
> I'd also like to see a cut down version of the lambda handler that _only_ serves GraphQL and drops the GraphiQL/event stream/etc stuff.

I'd be happy to do that, but it was not clear to me what the narrowest set of cases was that I needed to handle to support bare GraphQL. I think it might be a good idea to change `class GrafservBase` so that it has `processGraphQLRequest` and `processGraphiQLRequest` (and possibly some other narrow request-handling methods) whose return type is narrowed down accordingly, and then the lambda server could just call `processGraphQLRequest`, and let the compiler enforce a narrower `response.type` than what I get from `handleRequest`. But in absence of that I'd accept a list from you of which response types I can drop if all I want to support is bare GraphQL.

benjie:
Have a look at the Fastify adaptor; it adds separate route handlers for each middleware; here's the code to just do the GraphQL part (you can **not** pass the `this.graphiqlHandler` if you don't want the GraphQL endpoint to fall back to rendering GraphiQL):

https://github.com/graphile/crystal/blob/85dbd6d010dcba9bf6bb9f1f91c40c49cfcadcfa/grafast/grafserv/src/servers/fastify/v4/index.ts#L186-L192

> I think it might be a good idea to change class GrafservBase so that it has processGraphQLRequest and processGraphiQLRequest (and possibly some other narrow request-handling methods)

Indeed, it has `.graphqlHandler(...)`, etc already :+1: 

https://github.com/graphile/crystal/blob/85dbd6d010dcba9bf6bb9f1f91c40c49cfcadcfa/grafast/grafserv/src/core/base.ts#L43-L44

All `processRequest` does is a big try/catch and some path matching logic before handing over to the relevant handler (if any). The handlers don't do path checking - so you have to be sure that you want it to run the handler beforehand.

grafast/grafserv/src/servers/lambda/index.ts
```diff
@@ -0,0 +1,132 @@
...
+      context: LambdaContext,
+    ): Promise<APIGatewayProxyResult> {
+      const version = event.requestContext.protocol.match(
+        "^HTTP/(?<major>[0-9]+).(?<minor>[0-9]+)$",
+      );
+
+      const request: RequestDigest = {
+        httpVersionMajor: parseInt(version?.groups?.major ?? "1"),
+        httpVersionMinor: parseInt(version?.groups?.minor ?? "0"),
+        isSecure: event.headers["X-Forwarded-Proto"] === "https",
```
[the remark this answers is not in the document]
innermatrix: So this is an interesting question.

Can you always trust `X-Forwarded-Proto`? No. Can you always trust `X-Forwarded-Proto` in a lambda? Maybe, but I can't find a definitive proof of it. In part, this is because there are several different ways to connect HTTP(S) to a lambda (AWS API Gateway, AWS Application Load Balancer, Amazon CloudFront, to name just the ones I am aware of), and as a result this becomes a question of "how many different teams at Amazon, probably at completely different points in time, implemented this header, and what security guarantees were they willing to make".

But let me turn this around for a second and ask a different question: why does grafserv need to know? Sure, I can see an argument that someone using grafserv with a specific server implementation (like express) might be interested in this flag, but that doesn't necessarily mean that grafserv should be in the business of exposing that flag, because I am not convinced that grafserv will ever be able to make a security guarantee about this flag that works in all server implementations. So maybe the right answer here is that the http/https flag should be exposed in the server-specific section of the request digest, rather than in the server-agnostic section, and its semantics should be left up to the specific server in question.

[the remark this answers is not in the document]
benjie: Grafserv needs to know because it needs to point the client at the correct URLs for assets/etc when rendering Ruru, for example (essentially it's actually asking "is this HTTP or HTTPS in the URL?"); but we might use it for other purposes in future (e.g. deciding whether or not to send cookies, or whatever). Generally servers like Express, Fastify, etc have a setting something like "trustProxy" that can be set either to "true" or to a list of trusted endpoints. I'd be very concerned merging this without _some_ guarantee. But also, we definitely need it, so maybe if we flag it up strong enough that'll be sufficient. Is there no other way on lambda to know if it's HTTPS or not?

benjie:
Are you able to address the CI issues (looks like TypeScript isn't happy), or do you need guidance?

grafast/grafserv/src/servers/lambda/index.ts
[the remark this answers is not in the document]
innermatrix: The issue is that there is no such thing as "HTTPS (or not) lambda". The lambda itself receives requests on a protected backplane; this is similar to any kind of load-balancing proxy situation, in which the client only interacts with the proxy, and only the proxy interacts with the real backend.

Exposing a lambda to the outside world involves wiring up some kind of an HTTP-to-lambda bridge (of which there are at least three, as I mentioned). All of the HTTP-to-lambda bridges that I am aware of will us `X-Forwarded-Proto`, but idk that any of them have it documented as having security critical semantics. And certainly I don't feel confident enough that I know every corner of the sprawling AWS behemoth to promise that the header has robust security semantics in all possible paths from HTTP to lambda.

I think the issue here is that we are conflating two things: one is serving a web app (like Ruru) that is expecting to directly interact with the end-user-facing parts of the browser and consequently expecting to get sensible answers to browser-centric questions like "should I send cookies over" and "where are my assets" etc; the other is a microservice that serves up a node function (being the GraphQL query executor) over XHR, and doesn't need to deal with user-centric concerns at all. 

So maybe the secure flag should only be applicable to GraphiQL request handling, not to GraphQL request handling?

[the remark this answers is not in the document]
innermatrix: For what it's worth, `serverless-express` enumerates 13 (!!) different ways to get HTTP into a lambda (https://github.com/vendia/serverless-express/tree/75e9de327972ad372a89cd7ae586b53c73c048f0/src/event-sources/aws) and it doesn't set the `req.secure` flag in any of them. 

(And yes, there's a non-zero chance that this code doesn't handle at least one of those 13 ways of getting HTTP into a lambda, and I am fine with that at this stage.)

innermatrix:
> Are you able to address the CI issues (looks like TypeScript isn't happy), or do you need guidance?

What I need is lunch 🙂

innermatrix:
> Have a look at the Fastify adaptor; it adds separate route handlers for each middleware; here's the code to just do the GraphQL part:

I gave up on trying to do this for now. There's too much going on between `processRequest` and `graphqlHandler` that I don't want to refactor in this PR. For example:

* `processRequest` calls `graphqlHandler` and passes its output to `convertHandlerResultToResult`, but the parts of `convertHandlerResultToResult` that are specific to graphql are not separately available, so I'd have to call `convertHandlerResultToResult` myself — at which point I might have as well just called `processRequest`
* `processRequest` calls `handleGraphQLHandlerError`, which isn't exported

Given the structure of the code right now, it's far easier to keep calling `processRequest` and only implement the response types that are relevant to graphql, letting the others fall into a default error case.

grafast/grafserv/src/servers/lambda/index.ts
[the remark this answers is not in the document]
benjie: If we're not sure it's true, let's either:

1. set it false,
2. add a config option so the user can force it true or false, or
3. add a callback so the user can look at the incoming event and determine if it's secure or not.

Option 3 seems best; in fact it seems like the entire `const request: RequestDigest = ...` block should be provided by the _default_ implementation of this new function (as-is) and that function should be exported so that users can call it and then modify and return the result from their own callbacks. This would also enable them to do any custom tweaks for whichever of these 13 different ways they are using 😉

benjie:
> Given the structure of the code right now, it's far easier to keep calling processRequest and only implement the response types that are relevant to graphql, letting the others fall into a default error case.

Now's the time to refactor this code; before we reach 5.0.0 - if it's hard to do right now, that hints that it needs to be reworked. As you can probably tell it's already been reworked quickly a few times, and the result is a bit of a mess - really someone needs to cleanly separate the concerns in the newest pattern and then rewrite everything to fit that. My plan was to add as many different adaptors as we can to determine what these shared needs are, and then refactor based on that so that each adaptor can get smaller and smaller.

But I totally understand if you don't have time to take on that work! Just saying you're welcome to do it if you're inclined - just keep me abreast of any large changes since they're harder to review (we may end up splitting some of them into separate PRs).

innermatrix:
> But I totally understand if you don't have time to take on that work! Just saying you're welcome to do it if you're inclined - just keep me abreast of any large changes since they're harder to review (we may end up splitting some of them into separate PRs).

Yeah, I'm not saying I don't want to do it, I am saying I don't want to fold it into this PR 🙂 This PR is on the critical path to adopting v5 at work, and the refactor is not.

grafast/grafserv/src/servers/lambda/v1/index.ts
```diff
@@ -0,0 +1,128 @@
...
+        return {
+          statusCode,
+          headers: { ...headers, "Content-Type": "text/plain" },
+          body: error.message,
+        };
+      }
+
+      case "buffer": {
+        const { statusCode, headers, buffer } = response;
+        return { statusCode, headers, body: buffer.toString() };
```
benjie, line 81:
```suggestion
        return { statusCode, headers, body: buffer.toString("utf8") };
```

grafast/grafserv/src/servers/lambda/v1/index.ts
```diff
@@ -0,0 +1,128 @@
...
+      case "json": {
+        const { statusCode, headers, json } = response;
+        return { statusCode, headers, body: JSON.stringify(json) };
+      }
+
+      default: {
+        console.log("Unhandled:");
+        console.dir(response);
+        return {
+          statusCode: 503,
```
benjie, line 93: 501 is not implemented; if you copied this from elsewhere I should fix it there too.

```suggestion
          statusCode: 501,
```

grafast/grafserv/src/servers/lambda/v1/index.ts
```diff
@@ -0,0 +1,128 @@
...
+  protected processLambdaRequest(
+    _event: APIGatewayEvent,
+    _context: LambdaContext,
+    request: RequestDigest,
+  ) {
+    return this.processRequest(request);
+  }
+}
+
+export function grafserv(config: GrafservConfig) {
```
benjie, line 126:
```suggestion
/** @experimental */
export function grafserv(config: GrafservConfig) {
```

grafast/grafserv/src/servers/lambda/v1/index.ts
```diff
@@ -0,0 +1,128 @@
...
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace Grafast {
+    interface RequestContext {
+      lambdav1: { event: APIGatewayProxyEvent; context: LambdaContext };
+    }
+  }
+}
+
+export class LambdaGrafserv extends GrafservBase {
```
benjie, line 25:
```suggestion
/** @experimental */
export class LambdaGrafserv extends GrafservBase {
```

grafast/website/grafserv/servers/lambda.md
```diff
@@ -1,15 +1,24 @@
 # Lambda
 
-**TODO: actually implement this!**
+Grafserv supports the following AWS lambda configurations:
```
benjie, line 3:
```suggestion
**THIS INTEGRATION IS EXPERIMENTAL**. PRs improving it are welcome.

Grafserv supports the following AWS lambda configurations:
```

benjie (changes requested):

benjie:
I'm happy with this. Please:

1. Add a changeset by running `yarn changeset` in the root folder and following the instructions to write a description for the feature
1. Format the code by running `yarn lint:fix` in the root folder

Then we're good to go I think?

.changeset/twelve-apes-try.md
```diff
@@ -0,0 +1,5 @@
+---
+"grafserv": minor
```
benjie, line 2: During the pre-release stage, everything is a patch change.

```suggestion
"grafserv": patch
```

innermatrix:
@benjie lgtm ty

benjie:
I'll aim to release this tomorrow, but no promises.

PR #1784, opened 2023-10-04
Feat/types

### LINKED ISSUE #1755: Broken type helper (GetPgResourceRelations)

wesselvdv:
## Description

fixes #1755 

## Checklist


wesselvdv:
@benjie I think I found a way to delay the inference without causing `any` to fly around the code constantly:

```typescript
export interface AnyPgCodec extends PgCodec {} // <-- allows use of default generic parameters, and prevents "circular default detected" errors 
/**
 * A codec for a Postgres type, tells us how to convert to-and-from Postgres
 * (including changes to the SQL statement itself). Also includes metadata
 * about the type, for example any of the attributes it has.
 */
export interface PgCodec<
  TName extends string = string,
  TCodecAttributes extends
    ReadonlyArray<PgCodecAttribute> = ReadonlyArray<PgCodecAttribute>,
  TFromPostgres = any,
  TFromJavaScript = TFromPostgres,
  TArrayItemCodec extends AnyPgCodec = AnyPgCodec, // <-- using just PgCodec here causes circular default parameter issues
  TDomainItemCodec extends AnyPgCodec = AnyPgCodec,
  TRangeItemCodec extends PgRangeItemCodec = PgRangeItemCodec,
> {
// snip
}
```

I think I might've been able to get `PgResource.resolveVia` to type check without casting using the above method. I haven't done all of it yet, but it seems promising.

benjie:
Exciting! This reminds me of the kind of trick you need to do to build a JSONValue type, I should have thought about that - the difference between eager evaluated and lazily evaluated types... I'm not yet good enough at TypeScript to figure these things out for myself. Keep up the great work!

wesselvdv:
> Exciting! This reminds me of the kind of trick you need to do to build a JSONValue type, I should have thought about that - the difference between eager evaluated and lazily evaluated types... I'm not yet good enough at TypeScript to figure these things out for myself. Keep up the great work!

Thanks, I've got a basic working example schema again. There are still loads of mismatches, and biggest one is the `PgUnionAllStep` atm. I think you chose wisely to not try and type the associated pgresources there, but I am gonna try.

graphile-build/graphile-build-pg/src/plugins/PgAttributesPlugin.ts
```diff
@@ -42,7 +41,7 @@ declare global {
       _attributeName(
         this: GraphileBuild.Inflection,
         details: {
-          codec: PgCodecWithAttributes;
+          codec: DefaultPgCodec;
```
benjie, line 44: Should this (and the related changes below) have been `DefaultPgCodecWithAttributes` or similar?

.gitignore
```diff
@@ -43,4 +43,4 @@ grafast/dataplan-pg/__tests__/**/*.mermaid.png
 /Session.vim
 /WHAT_ARE_YOU_DOING.md
 contrib
-/env
+/env
```
benjie, line 46: I think you've removed the newline from the end of `.gitignore` - perhaps when you removed the nix stuff?

wesselvdv: I'll put it back, :-)

graphile-build/graphile-build-pg/src/plugins/PgAttributesPlugin.ts
wesselvdv: Yes, but this gave me a lot of errors, so I opted to widen the constraint for now, as I felt it wasn't necessarily a requirement for the purpose of this PR.

benjie:
I've merged the latest and I've renamed `AnyPg*` to `_AnyPg*` to make it clear it's internal, and I've renamed `DefaultPg*` to `GenericPg*` to show that e.g. GenericPgCodec represents a generic (arbitrary) PgCodec. 

`DefaultPgCodec` I found quite confusing until I realised it meant `PgCodecWithDefaultGenericParameters`; I had interpretted it a different way. I discussed a number of possible options with @jemgillam and in the end `GenericPgCodec` won out, even though ironically it doesn't have any generics!

grafast/dataplan-pg/src/codecs.ts
```diff
@@ -112,7 +161,7 @@ export interface PgCodecAttribute<
    * these are all plural relationships. So identicalVia is generally one-way
    * (except in 1-to-1 relationships).
    */
-  identicalVia?: PgCodecAttributeVia;
+  identicalVia?: _AnyPgCodecAttributeVia;
```
benjie, line 164: If I change `via` and `identicalVia` to `GenericPgCodecAttributeVia` then I get weird errors in `exampleSchema.ts`. However, I know that if `via` is set then it will always be `string | { relation: string, attribute: string }` (or a narrower form of this) so I don't really understand why this needs to use `any`s?

benjie:
I've also fixed the linting issues, but I'm a bit lost in understanding what we're actually achieving here - seems that we're hitting `any` pretty fast in internal code, which gives me pause.

benjie:
(To be clear: the previous comment was stating that I'm lost because of my poor understanding of TypeScript, not because the PR is bad. It could be that we were hitting `any` pretty soon already, and I just didn't realise it - I find it quite hard confirming TypeScript types sometimes, especially when the hover hints are so huge. Or it could be that the any's internally are desired in order to reduce the `any`'s that consumers receive; in which case that's a trade-off I'm willing to make but would require additional testing (or casting). I just started at the top of the PR and got lost pretty quickly (see comment) and am looking for some guidance/understanding.)

benjie:
(@wesselvdv and I had a long call where we went through this, and unfortunately though it solves some issues it also creates others, in particular making the codebase harder to work on. Currently the `_Any*` and `Generic*` types don't adhere to the rules that we'd like (library consumers use `Generic*` and internally we use `_Any*`), and we can get into some tricky situations; we're hoping that we can figure out some better rules and apply them to a clean checkout of the codebase. Leaving this open for context and as a reminder.)

PR #1803, opened 2023-10-10
Make `defaultBehavior: "-*"` work as expected

benjie:
## Description

It would make sense that `defaultBehavior: "-*"` would disable all behaviors; but currently it doesn't. This is due to the order that behaviors are added and to some legacy approaches for adding behaviors (e.g. adding directly to tags). This PR sets out to address this problem, but is a work in progress.

## Performance impact

Unknown.

## Security impact

None known.


benjie:
Related: https://github.com/graphile/crystal/issues/1762

PR #1823, opened 2023-10-19
Fix 'create' to 'insert' in behavior example

erichosick:
With example, get the error "Error: 'create' behavior is forbidden; did you mean 'insert'?" Using insert worked.

## Description

documentation

## Security impact

none

## Checklist


benjie (approved):
Thanks for catching this!

PR #1830, opened 2023-10-27
fix(ruru-components): use the headers while introspecting

JoviDeCroock:
## Description

fixes https://github.com/graphile/crystal/issues/1828

This fix feels a bit less idiomatic but we need to get access to the `EditorContext` so we can fill in the `headers` and use them for the schema. I see the reason for the custom `useSchema` is the syncing stream so I figured this might be an acceptable solutions to get to the best of both worlds case

## Performance impact

unknown

## Security impact

unknown

## Checklist

- [ ] I've added tests for the new feature, and `yarn test` passes. I could not find tests, feel free to point me in the right direction

### DIFF
--- a/.changeset/warm-rabbits-provide.md
+++ b/.changeset/warm-rabbits-provide.md
@@ -1,0 +1,6 @@
+---
+"ruru-components": patch
+---
+
+use the headers provided in the GraphiQL headers editor while performing
+introspection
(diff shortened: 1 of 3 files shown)


benjie:
Thanks @JoviDeCroock! Looking through this I realised that we don't actually need `useSchema` at all - we copied it from V4 so that we could control refreshing of the schema, but we can do that with GraphiQL hooks now. So I've moved more responsibility over to GraphiQL itself, and have raised #1833 as an alternative fix :raised_hands:

JoviDeCroock:
Awesome! I'll close this one out then

benjie:
Should be fixed in `ruru@2.0.0-beta.8` (available now)

PR #1868, opened 2023-11-22
doc: fix typo in loadOne doc and use object instead of list

smmoosavi:


benjie (approved):
Good catch!

benjie:
Please be aware that the ioEquivalence object only works when the keys and values are the same; this is fixed in https://github.com/graphile/crystal/pull/1870

PR #1875, opened 2023-11-30
Consider attribute's `isNotNull` flag for polymorphic types

### LINKED ISSUE #1872: Polymorphic types non-nullable attributes not working
Extending on the example from the [docs](https://postgraphile.org/postgraphile/next/polymorphism#interface-modesingle) to make the `title` attribute required with the exclamation mark syntax:

```
@type TOPIC name:SingleTableTopic attributes:title!
```

does not work. The `title` attribute stays nullable in the resulting GraphQL schema.

enisdenjo:
Closes #1872


benjie (approved):
Hooray! Gotta love a nice simple fix! :raised_hands:

Issue #1851, opened 2023-11-08
V5 totalCount failing on polymorphic 


dinodeSimon:
### Summary

When using relational polymorphic tables and including totalCount in a query an error is returned

### Steps to reproduce

Using the example at https://postgraphile.org/postgraphile/next/polymorphism#interface-moderelational and running the below query returns an error

query MyQuery {
  relationalItems {
    totalCount
  }
}

### Expected results

The totalCount value populated

### Actual results

"error: column \"__relational_items__.type\" must appear in the GROUP BY clause or be used in an aggregate function"

### Additional context

Using V5 beta 16

benjie:
Fixed in #1895

PR #1895, opened 2023-12-21
Fix aggregates over polymorphic connections

enisdenjo:
Showcasing #1851


benjie (approved):
Thanks for the reproduction @enisdenjo; this should now be resolved :+1:

PR #1914, opened 2024-01-18
Ensure response body contains expected error object when grafast context function throws an error

donalmacanri:
## Description

Currently, when you provide a function for the `grafast.context` option to populate context, and an error is thrown in the function (eg when parsing an auth header into context, but the auth header is malformed), execution bypasses the 'maskError' function, resulting in an 'Unknown error occurred' error in the response body, 500 status code, accompanied by a complete stack trace in the response.

https://discord.com/channels/489127045289476126/498852330754801666/1197168131756531732

I believe this fix also addresses the root cause of:
https://github.com/graphile/crystal/issues/1847

## Performance impact

Unknown

## Security impact

_Should_ improve security by not disclosing stack trace information if errors are thrown in the context function.

## Checklist

### DIFF
--- a/grafast/grafserv/__tests__/errors.test.ts
+++ b/grafast/grafserv/__tests__/errors.test.ts
@@ -1,0 +1,40 @@
+import { fetch } from "@whatwg-node/fetch";
+
+import { makeExampleServer } from "./exampleServer.js";
+
+let server: Awaited<ReturnType<typeof makeExampleServer>> | null = null;
+
+afterAll(() => {
+  server?.release();
+});
+
+test("response body contains expected error object when function provided as grafast context option throws an error", async () => {
+  const maskError = jest.fn((error) => error);
+  server = await makeExampleServer({
+    grafserv: {
+      graphqlOverGET: true,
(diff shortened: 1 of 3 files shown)


grafast/grafserv/__tests__/errors.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    grafserv: {
+      graphqlOverGET: true,
+      graphqlPath: "/graphql",
+      dangerouslyAllowAllCORSRequests: true,
+      maskError,
+    },
+    grafast: {
+      context: () => {
+        throw new Error("a particular error");
+      },
```
benjie, line 23: If you're looking to get another PR merged... Adding another test where context sleeps 5ms and then throws an error would be good - to differentiate a synchronous throw vs an async reject.

benjie (approved):
What a disgustingly simple solution :laughing: 

I can't help wondering if that `try` should actually be moved to the very beginning of the function. It would make sharing `args` with the `catch` more problematic though... So let's stick with this for now.

Thanks!

PR #1940, opened 2024-02-06
Grafserv Node Server - Skip processing request if response is already finished

diesal11:
Grafserv's node server has a bug where it attempts to handle responses that have already been sent by prior middleware, causing the classic `Error [ERR_HTTP_HEADERS_SENT]: Cannot write headers after they are sent to the client`

Example server:
```ts
const serv = pgl.createServ(grafserv);
const server = createServer();
server.on('error', console.error);

server.on('request', (req, res) => {
  // Set CORS headers
	res.setHeader('Access-Control-Allow-Origin', req.headers.origin || '*');
	res.setHeader('Access-Control-Allow-Methods', 'OPTIONS, POST');
	res.setHeader('Access-Control-Allow-Headers', '*');

  // Respond to CORS preflight requests
  if (req.method === 'OPTIONS') {
    res.writeHead(200);
    res.end();
  }
});

serv.addTo(server).catch((e) => {
  console.error(e);
});
```

### DIFF
--- a/grafast/grafserv/src/servers/node/index.ts
+++ b/grafast/grafserv/src/servers/node/index.ts
@@ -109,6 +109,16 @@ export class NodeGrafservBase extends GrafservBase {
     const dynamicOptions = this.dynamicOptions;
     return async (req, res, next) => {
       try {
+        // If middleware prior to this has already finished the response,
+        // then skip processing the request.
+        if (res.writableFinished) {
+          if (typeof next === "function") {
+            next();
+          } 
+
+          return;
+        }
+
         const request = this.getDigest(dynamicOptions, req, res, isHTTPS);
         const result = await this.processRequest(request);
 


benjie:
Thanks for the suggestion, but it's not safe to attempt to add multiple "middlewares" to the Node HTTP server since it does not support middleware. If the middleware that you listed above was `async` and had a small delay then the two could end up conflicting in interesting ways. You should either wrap the PostGraphile middleware such that yours is called first and then you hand over to PostGraphile's middleware from yours (e.g. via a `next()` callback), or, better, you should use a server such as Express, Koa, Fastify, etc that supports middleware/routing to get the behavior you desire.

I won't be merging this pull request as it enables a pattern that is liable to be a footgun; I'd rather users discover earlier that there are issues (like the above) and fix them in their code.

diesal11:
Appreciate you for clarifying my mistake here, i've just clicked to some valuable knowledge on EventEmitters thanks to that.

Yes, was attempting to avoid existing frameworks as we will run on Lambda & this was just for local, but perhaps was trying too hard. Express it is :)

PR #1948, opened 2024-02-07
postgraphile website: Example Gallery rejig

jemgillam:
## Description

Moves the example gallery on postgraphile.org from top level to within the versioned docs (4.12 only). The actual examples are all pre-existing so this should be easy to merge. This also allows our example gallery to be versioned going forwards. There is no navigation to `/next/examples` any more, but if you end up there you get the same page as before directing you to the graphile.org site. 

`/current/examples` 

Screenshot 2024-02-07 at 14 17 39

`/next/examples`

Screenshot 2024-02-07 at 14 18 04


jemgillam:
If you want to hold off merging, I'll add the function gallery too (https://www.graphile.org/postgraphile/function-gallery/) I'm sure I did this already but I can find no evidence of it!

benjie (approved):

PR #1956, opened 2024-02-10
Add example of maxRequestLength

### LINKED ISSUE #1918: Set bodySizeLimit for graphile v5
### Summary
I'm using v5 and faced a need to increase bodySizeLimit, wasn't able to do so using a library

```
const preset: GraphileConfig.Preset = {
	extends: [PostGraphileAmberPreset],
	pgServices: [
		makePgService({
			connectionString,
			schemas: ['app_public']
		})
	],
	grafast: {
		context(requestContext, args) {
			return {
				pgSettings: {
					...args.contextValue?.pgSettings,
				}
			};
		}
	},
	schema: {}
};

export const pgl = postgraphile(preset);
const serv = pgl.createServ(grafserv);
```

I would expect a way to specify a http body limit, so my queries would pass thru 

I've found an option at grafast : _maxRequestLength_ but wasn't able to find the way hot to incorporate it in code above

XanderG2:
## Description

I've checked that this option works, it is already documented in the postgraphile documentation. This adds it to the grafserv documentation

- Fixes #1918

## Performance impact

Documentation only

## Security impact

Documentation only

## Checklist

### DIFF
--- a/grafast/website/grafserv/index.md
+++ b/grafast/website/grafserv/index.md
@@ -36,6 +36,7 @@ const preset = {
     graphiql: true,
     graphiqlPath: "/",
     websockets: true,
+    maxRequestLength: 100000,
   },
 };
 ```


benjie (approved):
Thanks!

PR #1962, opened 2024-02-12
export PgEntity type in pg-introspection

DanielFGray:
## Description

this should add `PgEntity` type to the package exports. it is referenced in other types but not directly importable

I have not done any additional testing but I assume this would fix it

## Performance impact

since it's only a small type change it should have no impact on runtime performance

## Security impact

I am unaware of any security impacts

## Checklist


benjie (approved):
LGTM!

PR #1978, opened 2024-03-01
Change `connection()` signature and allow associating extra data with edges

benjie:
This feature was effectively requested via a discussion on Discord:
https://discord.com/channels/489127045289476126/498852330754801666/1213043884033314836

Example usage:

```ts
import { connection, lambda, object } from "postgraphile/grafast";
import { gql, makeExtendSchemaPlugin } from "postgraphile/utils";
export const plugin = makeExtendSchemaPlugin((build) => {
  const { left_arm } = build.input.pgRegistry.pgResources;
  return {
    typeDefs: gql`
      extend type Person {
        allArms: PersonRelatedArmConnection
      }
      type PersonRelatedArmConnection {
        edges: [PersonRelatedArmEdge]
      }
      type PersonRelatedArmEdge {
        node: LeftArm
        isTheirs: Boolean
      }
    `,
    plans: {
      Person: {
        allArms($person) {
          const $arms = left_arm.find();

          return connection($arms, {
            edgeDataPlan($arm) {
              return object({ arm: $arm, person: $person });
            },
          });
        },
      },
      PersonRelatedArmEdge: {
        isTheirs($edge) {
          const $obj = $edge.data();
          const $arm = $obj.get("arm");
          const $armPersonId = $arm.get("person_id");
          const $person = $obj.get("person");
          const $personId = $person.get("id");
          return lambda(
            [$personId, $armPersonId],
            ([personId, armPersonId]) => personId === armPersonId,
          );
        },
      },
    },
  };
});
```

with query:

```graphql
{
  allPeopleList {
    id
    allArms {
      edges {
        isTheirs
        node {
          id
          lengthInMetres
          personByPersonId {
            id
          }
        }
      }
    }
  }
}
```

This demonstrate paginating over `leftArms` but getting edge-associated data indicating if the given arm belongs to the person it was selected on or not.

### DIFF
--- a/.changeset/fluffy-suits-tease.md
+++ b/.changeset/fluffy-suits-tease.md
@@ -5,10 +5,33 @@
 ---
 
 Breaking: `connection()` step now accepts configuration object in place of 2nd
-argument onwards.
+argument onwards:
+
+```diff
+-return connection($list, nodePlan, cursorPlan);
++return connection($list, { nodePlan, cursorPlan });
+```
 
 Feature: `edgeDataPlan` can be specified as part of this configuration object,
-allowing you to associate edge data with your connection edges.
+allowing you to associate edge data with your connection edges:
+
+```ts
+return connection($list, {
+  edgeDataPlan($item) {
+    return object({ item: $item, otherThing: $otherThing });
+  },
+});
+
+// ...
+
+const plans = {
+  FooEdge: {
+    otherThing($edge) {
+      return $edge.data().get("otherThing");
+    },
+  },
+};
+```
 
 Feature: `ConnectionStep` and `EdgeStep` gain `get()` methods, so
 `*Connection.edges`, `*Connection.nodes`, `*Connection.pageInfo`, `*Edge.node`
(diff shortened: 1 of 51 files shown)


benjie:
Like `each()`, this functionality is a little cheeky in that it uses a callback which can allow the user to "smuggle" dependencies from one layer in the graph to another. This potentially allows violation of the GraphQL spec, and instead should be accomplished via _also_ adding these dependencies to the parent (i.e. adding them to `each()` itself, or in this case to `connection()` itself). Usage of this pattern should be managed very carefully indeed - ideally we should have lint rules that forbid the callbacks from referencing steps from their closure... but if we add that we'll need a way of passing additional dependencies into the `each` callback; something like: `each($things, [$otherThing], ($thing, [$otherThing]) => object({ thing: $thing, otherThing: $otherThing }))`

I've thought about this a fair amount (hence the delay in merging this) and discussed it with @jemgillam  on a few occasions... but ultimately we've decided that we'll allow this current pattern because it's more efficient (no need to pass useless dependencies to parent steps) and it's more ergonomic (users don't have to worry about additional dependencies). We may revisit this decision if it turns out that we've overlooked something major.

PR #1985, opened 2024-03-07
Make adaptor in PgServiceConfiguration the concrete adaptor instance, not import path

### LINKED ISSUE #1826: Bundling error when used with NextJS
### Summary

Unable to import postgraphile (v5) or grafast in an next js project.

Attempting to setup a route that includes postgraphile v5 but next refuses to bundle

```
 ⚠ ./node_modules/graphile-build-pg/dist/pgServices.js
Critical dependency: the request of a dependency is an expression        

Import trace for requested module:
./node_modules/graphile-build-pg/dist/pgServices.js
./node_modules/graphile-build-pg/dist/index.js
./node_modules/postgraphile/dist/index.js
./src/lib/graphile/pgl.ts
./src/app/graphql/route.ts
```

```
//pgl.ts
import preset from './graphile.config'
import postgraphile from 'postgraphile'

export const pgl = postgraphile(preset)
```

### Steps to reproduce

Create a new next js app using `npx create-next-app@latest `

install  postgrahile@beta I was using `^5.0.0-beta.15`

Attempt to setup a route which imports postgraphile

### Expected results

Next js able to complile / bundle

### Actual results

A bundling error

### Additional context

@benjie thinks this is a bundling error 

> It’s a bundling issue; I’m afraid I’ve not put time into supporting that yet. Please file an issue with the above details; it sounds like I’ll need to change the configuration format to avoid a dynamic require

https://discord.com/channels/489127045289476126/1165551144299401317/1165574782675329056

image

hannesj:
## Description

Currently the exact adaptor to be used is passed in as an import specifier, rather than the concrete module instance. This changes it so that the concrete adaptor is passed in the config, making it possible to bundle the library eg for usage on function as a service platforms.

Fixes #1826 as per suggestion. 

Technically this is a breaking change as the configuration format changes, but I'm not sure how widely it is currently used today.

## Performance impact

Should not affect

## Security impact

Should not affect

## Checklist

### DIFF
--- a/grafast/dataplan-pg/src/index.ts
+++ b/grafast/dataplan-pg/src/index.ts
@@ -481,7 +481,18 @@ declare global {
     }
 
     interface PgAdaptors {
-      /* Add your own via declaration merging */
+      /*
+       * Add your adaptor configurations via declaration merging; they should
+       * conform to this:
+       *
+       * ```
+       * [moduleName: string]: {
+       *   adaptorSettings: { ... } | undefined;
+       *   makePgServiceOptions: MakePgServiceOptions & { ... };
+       *   client: PgClient & MyPgClientStuff;
+       * };
+       * ```
+       */
     }
   }
   namespace DataplanPg {
--- a/grafast/dataplan-pg/src/index.ts
+++ b/grafast/dataplan-pg/src/index.ts
@@ -1,7 +1,6 @@
 import type { GrafastSubscriber } from "grafast";
 import { exportAsMany } from "grafast";
 
-import type { PgAdaptorOptions } from "./adaptors/pg.js";
 import {
   domainOfCodec,
   enumCodec,
@@ -427,12 +426,12 @@ export { version } from "./version.js";
 
 declare global {
   namespace GraphileConfig {
-    interface PgServiceConfiguration<TAdaptorOptions = PgAdaptorOptions> {
+    interface PgServiceConfiguration<TAdaptorSettings> {
       name: string;
       schemas?: string[];
 
-      adaptor: PgAdaptor<TAdaptorOptions>;
-      adaptorSettings?: TAdaptorOptions;
+      adaptor: PgAdaptor<TAdaptorSettings>;
+      adaptorSettings?: TAdaptorSettings;
 
       /** The key on 'context' where the withPgClient function will be sourced */
       withPgClientKey: KeysOfType<Grafast.Context & object, WithPgClient>;
@@ -468,7 +467,7 @@ declare global {
     }
 
     interface Preset {
-      pgServices?: ReadonlyArray<PgServiceConfiguration>;
+      pgServices?: ReadonlyArray<PgServiceConfiguration<unknown>>;
     }
   }
   namespace DataplanPg {
--- a/grafast/dataplan-pg/src/index.ts
+++ b/grafast/dataplan-pg/src/index.ts
@@ -1,6 +1,7 @@
 import type { GrafastSubscriber } from "grafast";
 import { exportAsMany } from "grafast";
 
+import type { NodePostgresPgClient, PgAdaptorSettings } from "./adaptors/pg.js";
 import {
   domainOfCodec,
   enumCodec,
@@ -426,15 +427,21 @@ export { version } from "./version.js";
 
 declare global {
   namespace GraphileConfig {
-    interface PgServiceConfiguration<TAdaptorSettings> {
+    interface PgServiceConfiguration<
+      TAdaptor extends
+        keyof GraphileConfig.PgAdaptors = keyof GraphileConfig.PgAdaptors,
+    > {
       name: string;
       schemas?: string[];
 
-      adaptor: PgAdaptor<TAdaptorSettings>;
-      adaptorSettings?: TAdaptorSettings;
+      adaptor: GraphileConfig.PgAdaptors[TAdaptor]["adaptor"];
+      adaptorSettings?: GraphileConfig.PgAdaptors[TAdaptor]["adaptorSettings"];
 
       /** The key on 'context' where the withPgClient function will be sourced */
-      withPgClientKey: KeysOfType<Grafast.Context & object, WithPgClient>;
+      withPgClientKey: KeysOfType<
+        Grafast.Context & object,
+        WithPgClient<GraphileConfig.PgAdaptors[TAdaptor]["client"]>
+      >;
 
       /** Return settings to set in the session */
       pgSettings?: (
@@ -467,7 +474,16 @@ declare global {
     }
 
     interface Preset {
-      pgServices?: ReadonlyArray<PgServiceConfiguration<unknown>>;
+      pgServices?: ReadonlyArray<PgServiceConfiguration>;
+    }
+
+    interface PgAdaptors {
+      "@dataplan/pg/adaptors/pg": {
+        adaptor: PgAdaptor;
+        adaptorSettings: PgAdaptorSettings;
+        client: NodePostgresPgClient;
+      };
+      /* Add your own via declaration merging */
     }
   }
   namespace DataplanPg {
--- a/grafast/dataplan-pg/src/index.ts
+++ b/grafast/dataplan-pg/src/index.ts
@@ -478,11 +478,6 @@ declare global {
     }
 
     interface PgAdaptors {
-      "@dataplan/pg/adaptors/pg": {
-        adaptor: PgAdaptor;
-        adaptorSettings: PgAdaptorSettings;
-        client: NodePostgresPgClient;
-      };
       /* Add your own via declaration merging */
     }
   }
--- a/grafast/dataplan-pg/src/index.ts
+++ b/grafast/dataplan-pg/src/index.ts
@@ -1,7 +1,6 @@
 import type { GrafastSubscriber } from "grafast";
 import { exportAsMany } from "grafast";
 
-import type { NodePostgresPgClient, PgAdaptorSettings } from "./adaptors/pg.js";
 import {
   domainOfCodec,
   enumCodec,
@@ -434,7 +433,7 @@ declare global {
       name: string;
       schemas?: string[];
 
-      adaptor: GraphileConfig.PgAdaptors[TAdaptor]["adaptor"];
+      adaptor: PgAdaptor<TAdaptor>;
       adaptorSettings?: GraphileConfig.PgAdaptors[TAdaptor]["adaptorSettings"];
 
       /** The key on 'context' where the withPgClient function will be sourced */
@@ -474,7 +473,11 @@ declare global {
     }
 
     interface Preset {
-      pgServices?: ReadonlyArray<PgServiceConfiguration>;
+      pgServices?: ReadonlyArray<
+        {
+          [Key in keyof GraphileConfig.PgAdaptors]: PgServiceConfiguration<Key>;
+        }[keyof GraphileConfig.PgAdaptors]
+      >;
     }
 
     interface PgAdaptors {
--- a/grafast/dataplan-pg/src/index.ts
+++ b/grafast/dataplan-pg/src/index.ts
@@ -427,15 +427,12 @@ export { version } from "./version.js";
 
 declare global {
   namespace GraphileConfig {
-    interface PgServiceConfiguration<
-      TAdaptor extends
-        keyof GraphileConfig.PgDatabaseAdaptorOptions = keyof GraphileConfig.PgDatabaseAdaptorOptions,
-    > {
+    interface PgServiceConfiguration<TAdaptorOptions = PgAdaptorOptions> {
       name: string;
       schemas?: string[];
 
-      adaptor: TAdaptor;
-      adaptorSettings?: GraphileConfig.PgDatabaseAdaptorOptions[TAdaptor];
+      adaptor: PgAdaptor<TAdaptorOptions>;
+      adaptorSettings?: TAdaptorOptions;
 
       /** The key on 'context' where the withPgClient function will be sourced */
       withPgClientKey: KeysOfType<Grafast.Context & object, WithPgClient>;
@@ -472,11 +469,6 @@ declare global {
 
     interface Preset {
       pgServices?: ReadonlyArray<PgServiceConfiguration>;
-    }
-
-    interface PgDatabaseAdaptorOptions {
-      "@dataplan/pg/adaptors/pg": PgAdaptorOptions;
-      /* Add your own via declaration merging */
     }
   }
   namespace DataplanPg {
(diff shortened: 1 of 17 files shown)


benjie (commented):
From what I can tell from a quick scan over this (and without trying it out) I think we're losing one of the most powerful features of adaptors (and the primary reason that they exist): the ability to extend your adaptor with additional properties/methods specific to its capabilities. For example, some people might like a `pg-promise` adaptor which allows them to also use the `pg-promise` instance directly in their mutations, but though this would still be compatible with `PgAdaptor`, it would not be an exposed interface in the way that this is modelled. The previous string-based method allowed registering your own adaptors with additional properties/methods by name.

I wonder if we can achieve both approaches in a cunning way, such as:

```ts
adaptor: {
  name: '@dataplan/pg/adaptors/pg',
  module: await import('@dataplan/pg/adaptors/pg')
}
```

WDYT?

hannesj:
> it would not be an exposed interface in the way that this is modelled.
It shouldn't need to be? It only needs to implement the `PgAdaptor` interface, that we any way need to expose if we want the user to be able to supply their own imported module? I was under the impression that supplying the "correct" adaptorSettings as part of the config was the Typescript trickery mentioned, as that is the only part that can be overridden by declaration merging and for which the name of the module was used.

benjie:
Ah you're right, we never did make `WithPgClient` generic, so it does not pass through the adaptor currently and thus the PgClient you get doesn't explicitly support additional methods (though you can cast it to do so, I guess). I'm guessing I gave up trying to get it to work due to lack of time :man_shrugging: 

I'll revisit this PR next week, thanks again :+1:

grafast/dataplan-pg/src/index.ts
```diff
@@ -427,15 +427,12 @@ export { version } from "./version.js";
 
 declare global {
   namespace GraphileConfig {
-    interface PgServiceConfiguration<
-      TAdaptor extends
-        keyof GraphileConfig.PgDatabaseAdaptorOptions = keyof GraphileConfig.PgDatabaseAdaptorOptions,
-    > {
+    interface PgServiceConfiguration<TAdaptorOptions = PgAdaptorOptions> {
```
benjie, line 430: We should not have this; defaulting to using the adaptor options for the 'pg' adaptor doesn't make sense for the majority of other adaptors, and would be unsafe to use across the codebase.

Removing this default causes cascading failures. I can't push to your branch so I've pushed up a commit here https://github.com/graphile/crystal/commit/8d57b63cea2e0750506402e0779c177dcdc0c64b to address the beginnings of these, but I can't figure out how to get it to pass with solid typings without resorting to `any`. Please feel free to pull my changes into your branch and continue iterating.

benjie:
Regarding what I was discussing before, specifically I was referring to this pattern:

https://github.com/graphile/crystal/blob/05265a2d5576430256f796f17cebebf030206afe/grafast/dataplan-pg/src/adaptors/pg.ts#L377-L380

Note that we don't just return a generic PgClient, we return a NodePostgresPgClient. My hope was to hook this up via TypeScript such that users could conveniently use the additional methods on their PgClient of choice (in this case the `rawPgClient` property) in their plans/etc.

https://github.com/graphile/crystal/blob/05265a2d5576430256f796f17cebebf030206afe/grafast/dataplan-pg/src/adaptors/pg.ts#L64-L66

The plan was to do more declaration merging, e.g. something like:

```ts
declare global {
  namespace DataplanPg {
    interface PgClientByAdaptor {
      "@dataplan/pg/adaptors/pg": NodePostgresPgClient
      // Add more here via declaration merging
    }
  }
}
```

then you can use the adaptor name to look up the client type. I'm not sure it would be possible to do the same using a TAdaptorSettings generic?

benjie:
I've marked this as a draft; please ping me when you're ready for another review. Thanks for your work on this! :raised_hands:

hannesj:
Sorry for taking some time with this. I now preserved the approach of having a string-keyed object containing all the different adaptors, and using that string to select which on will be used in a config.

grafast/dataplan-pg/src/steps/withPgClient.ts
```diff
@@ -3,17 +3,19 @@ import { constant, ExecutableStep } from "grafast";
...
 import type { PgClient, PgExecutor, WithPgClient } from "../executor";
 
-export type WithPgClientStepCallback<TData, TResult> = (
-  client: PgClient,
-  data: TData,
-) => Promise<TResult>;
+export type WithPgClientStepCallback<
+  TPgClient extends PgClient,
+  TData,
+  TResult,
```
benjie, line 9: Here's roughly [the status quo on TypeScript Playground](https://www.typescriptlang.org/play/?#code/C4TwDgpgBACg5gUQB4QMYFdgHsBOUC8UA3hAFxTA7oQC+A3AFCiRTJqYCGARgDYQDKwCGAA8AFQB8BYgGchYcmJoMAlgDshOAGYdU0eAGEeKiBuIMoUAI7UcIABQcccGeSIUISYOTk51cKBoASnIYHCwAWxUZCBEONRAJBmUGTzBcYApwaAB1FWAAC0NjU2BBYQMOHh4uXQBrcQARDmAOABooMQAlCBl0HmApQnsLKFQSjVC4IxMNNtGAExaORWbW+aCCKTDI6Nju3v7BxjSMqC10NVRgFSw1KAB3fKLpieAm5Y6DvoGJEctPOxsDgpmwMMD5pYACRLVqkUaWAA+rBQ4O4fHKojEaw4SUsSKg9mxyygniEagWMigan6PCgyMuCwgWnUEAWUAA-NTafSoIzmaz2eQ1BAAG4QHBBSFjKo1erkPKFYqzMrySrVWqoBrE9adHo-QYbchgzi8ATycT6o5SIijHAQYDoHD3GnVKAcKnxECMFKoO5yUmozC4aSuuke90Jd1U+Am4GMP1qANQlYooHo83CES+fxDWTycgAIiEckL9AYiYDWmkTyVrxV9kB4NwHRTHQ9ICuhPGKvbm3wUg4QUYAHoR-ioAA9DkMIA). Note `TData` and `TResult` can be inferred from usage.

Adding a new generic at the beginning seems like it's a breaking change for no specific reason; e.g. if you call `withPgClient<SuperPgClient>(...)` [you'll get the error](https://www.typescriptlang.org/play/?#code/C4TwDgpgBACg5gUQB4QMYFdgHsBOUC8UA3hAFxTA7oQC+A3AFCiRTJqYCGARgDYQDKwCGAA8AFQB8BYgGchYcmJoMAlgDshOAGYdU0eAGEeKiBuIMoUAI7UcIABQcccGeSIUISYOTk51cKBoASnIYHCwAWxUZCBEONRAJBmVVDQhtXWh+dEgcQ2NTYChPITUAExlYOCMTMyILSygy+L17EKg1dAiudMYUzzBcIuZoAHUVYAALfNrgQWEDDh4eLl0Aa3EZwuKvUwqqmsKAGigxABEOYA4TsQAlCBl0HmApQnsG1AKNRS2NI4bmldFBcrv8ggQpGFItFYncHk8XowBkMoFp0GpUMAVFg1FAAO4TabVL7ATbE2Y7Ur7X7AG4g66ne6PZ4Sd6WTzsbA4UKIFAYLn-SwAEkBHFIDUsAB9WHzOLwBPJxPSko0oNL7OdLhxKXtKp1lmqoOiyhAtOoIGUoAB+DpPHiG42m82W8hqCAAN3SQUFUFQSxW63I4ymNPmYEWy1WqA2YhpdK1NyZCIkYPIbH53D4YfESZZ5ksOAgwHQOFx+vtHEq8RAfQYDFQOLkO05uGk5aglY7CQ7lXg6cwuEYDbUTaFYplnMzCuEIl8-lesnk5AAREI5Mv6PXG0UtNICSHyYURNlcjTWRz+bgTmOTpWQBioPZPrNb+D8FJ3Bw8RwJr6SQA6Gx0gcIghC8cgOGCOgoELYtS0sSCgkYAB6ZDVQAPStBggA) that `withPgClient` requires 3 generics. So if we're going to add this I think we should do so via the last position:

```suggestion
export type WithPgClientStepCallback<
  TData,
  TResult,
  TPgClient extends PgClient = PgClient,
```

Unless there's a good reason not to?

Further I think we should do a ridiculous TypeScript hack for typing and insert a function that does (essentially) nothing:

```ts
const $result = withCustomPgClient<SuperPgClient>()(
  executor,
  $data,
  async (pgClient, data) => {
    //   ^? SuperPgClient
  }
);
```

This allows you to pass the type for the client whilst still having the other two generics inferred.

https://www.typescriptlang.org/play/?#code/C4TwDgpgBACg5gUQB4QMYFdgHsBOUC8UA3lBAFxTA7rQC+A3AFCiRTJqYCGARgDYQBlYBDAAeACoA+AsSgBnYWArioDRgEsAdsJwAzTqmjwAwr3URtxRlCgBHGjhAAKTjjhyKJYUmAUFOLThVAEoKGBwsAFt1OQhRTk0QSSZaRg1tCD0DaAF0SBwTMwtgUh8LABM5WDhTc0siayhyhMMnUKhNdEjuTJS0iCQwXBKWaAB1dWAAC0K64CERY05eXm4DAGtRRvEAEU5gTgAabYAlCDl0XmBjm3FZ4tLhTUrq2uLj6UInVCLtZXvtIcmvtOMo9gdggRpOEojE4uIzhcrslGAMhjgSrp0JpUMB1FhNFAAO6TGY1X7ACQAkoDJ4valA3YgxmIy7ASRORoDDjYHBhRAoDC8m5QAAkzQOZEaNgAPmxBVw+IJFBJwZxJNKoHKnEyDo8KlVOistVBseUILotBBylAAPwdS68E1mi1Wm0UTQQABumWCItQy1WGwoE2m1IWYCWKzWqE2uqOUAR5zZjOpkmO7XYQp4-AjElZyKsNhwEGA6BwhKNTs4VQSID6jFQBIUpR5uBkVagNa7iS7VXgWcwuCYTc0LdFoPlPJzypEon8gU+sgUIgoACJhAo16oR83MTISWHyXMnNyhbggROgTWQDioN8KdfIfhpA0bD85kwbAB6b9QAB6tqNJwRKcJMUAfsUAB09iZM4Xi0hQnAhF+UAlmWFZdikwS7mOJQAGIyLoTC-jYZFkYB-SDMMUBYjieIEsSpLGOgChRNSVLHg8tIGq8FIcu0Wy3GqIpJki1yMByXIKry-KDsKjTiiCUrkXK8kznm8Yaqp97xvqzyGo6zrPK6no2vanZyi6lpmVAHrer6-qBjG6whqS4aKFGQaxqqzKJgW1yJmmGZQlO2ZKppAWvo06HloSh5km82gNqOLZBIQCUsWxkQcbk+Rpm0nI2GeQ44CKV7AXIt6oPekGAl2z7ReRdXAKhP5-pR5EgWBJQtTBDjwZQiFdihmqxZhnCobQGa4S2ADiMhwCRf7kRRQFAA


That said... It feels like it's the `executor` itself that should contain information on what the PgClient type is (since that's where the plan that gets it from `context()` resides). We might need TypeScript code generation for that to work reliably though.

postgraphile/postgraphile/__tests__/helpers.ts
```diff
@@ -255,15 +257,15 @@ export async function runTestQuery(
...
-            ? {
+            ? () => ({
                 search_path,
-              }
+              })
             : undefined,
         schemas: schemas,
         adaptorSettings: {
           connectionString,
         },
```
benjie, line 267: You still have `adaptorSettings` here, but removed it in other places?

postgraphile/postgraphile/src/cli.ts
```diff
@@ -186,16 +185,10 @@ export async function run(args: ArgsFromOptions<typeof options>) {
...
-    const importSpecifier = adaptor.match(/^([a-z]:|\.\/|\/)/i)
-      ? pathToFileURL(adaptor).href
-      : adaptor;
-
-    const mod = await import(importSpecifier);
-    const makePgService = (mod.makePgService ?? mod.default?.makePgService) as (
-      options: MakePgServiceOptions,
-    ) => GraphileConfig.PgServiceConfiguration;
+    const makePgService = adaptor.makePgService;
     if (typeof makePgService !== "function") {
```
benjie, line 192: Please re-add the ESM shenanigans avoidance code, otherwise my support burden will increase :sweat_smile: 

```ts
let tmp;
const adaptor =
      preset.pgServices?.[0]?.adaptor ??
      ((tmp = await import("@dataplan/pg/adaptors/pg")) && tmp.makePgService ? tmp : tmp.default);
```

postgraphile/website/postgraphile/config.mdx
```diff
@@ -309,8 +309,8 @@ nitty-gritty: each entry in the list is an object with the following keys (only
 - `name: string` - an arbitrary unique name for this config; please keep it
   alphanumeric! Don't set this unless you have more than one pgService; see
   warning below.
-- `adaptor: string` - the name of the module to use as the postgres adaptor;
-  e.g. `@dataplan/pg/adaptors/pg` for the `pg` module
+- `adaptor: PgAdaptor` - the module to use as the postgres adaptor; e.g. the
+  result of `import(@dataplan/pg/adaptors/pg)` for the `pg` module
```
benjie, line 313:
```suggestion
  result of `import("@dataplan/pg/adaptors/pg")` for the `pg` module
```

benjie (changes requested):
Sorry for the delay in reviewing, I've been head down in rewriting some complex Grafast internals!

~~If the aim of this is that you can pass your custom client type via a generic then we should keep that in mind but~~ make sure that we don't a) break existing code and b) make it harder for people who don't want to replace their PgClient. I think we should move the TPgClient generic to the end, and add a method so you can pass just the client as a generic and the other generics get inferred automatically.

~~However, if the aim is that using `withPgClient` should automatically infer the PgClient interface then we should explore other options (we'll still need the generics added):~~

~~Option 1: make the client a generic parameter to the executor, and infer from there automatically.~~

~~Option 2: _guess_ that we're reading from the `context.withPgClient` property, and thus read the type from `Grafast.Context['withPgClient']`. I'm not keen on this option because it's inflexible, and assumes you only have one client.~~

~~Non-breaking addition of the generics should be able to go through anyway, so we don't need to decide Option 1 or 2 now - we can merge without that.~~

benjie:
Dear Benjie, the main motivation for this is to improve _bundling_ as outlined in #1826. TypeScript is a secondary concern. As such, I think we should follow your suggestion above:

> we should move the `TPgClient` generic to the end, and add a method so you can pass just the client as a generic and the other generics get inferred automatically.

benjie:
Rebased on the latest main; did a rough comparison of the diff to see that the rebase went well.

benjie:
Massively overhauled; I've removed the adaptor from the types itself because it became a bit circular; instead now you just define the argument types and the client types and we imply everything else from that. I've also moved all the new generics to the end and given them all defaults, this makes the change minimally breaking. Looks good to me; thanks for your work on this! :raised_hands:

benjie (approved):

PR #1990, opened 2024-03-13
Allow loadOne/loadMany to support unary steps

benjie:
## Description

Following [this discussion in Discord](https://discord.com/channels/489127045289476126/498852330754801666/1217421837206552689) I felt rather motivated to add support for unary steps to loadOne/loadMany.

### Before

```ts
const $executorContext = executor.context();
const $id = ...;
const $result = loadOne(list([$executorContext, $id]), async (tuples) => {
  // Batch the values by their executor context; probably there will only be one batch entry. (Highly technical note: unary values have been added to `main` recently but not yet shipped; we may be able to get rid of this later by changing the signature of loadOne and declaring $executorContext as a unary value so that you know there will be exactly one value of it - then you will just have `executorContext` and `ids` without this batching overhead.)
  const idAndIndexesByExecutorContext = new Map();
  for (let i = 0; i < tuples.length; i++) {
    const [executorContext, id] = tuples[i];
    const list = idAndIndexesByExecutorContext.get(executorContext) || [];
    if (list.length === 0) {
      idAndIndexesByExecutorContext.set(executorContext, list);
    }
    list.push([id, i]);
  }

  // Prepare the results array
  const results = new Array(tuples.length);

  // Now execute these batches, and write to the results array
  for (const [executorContext, idsAndIndexes] of idAndIndexesByExecutorContext) {
    const ids = idsAndIndexes.map(t => t[0]);
    const { withPgClient, pgSettings } = executorContext;
    const data = await withPgClient(pgSettings, (client) => getTheRecordsByIdsOrWhatever(client, ids));
    // Write the results back to the results array
    for (const entry of data) {
      // Find the index of this result from the DB
      const idx = ids.indexOf(entry.id);
      if (idx >= 0) {
        // Find the corresponding index in the input tuples list
        const resultIdx = idAndIndexes[idx][1];
        // Store the result in this index
        results[resultIdx] = entry;
      }
    }
  }

  // Finally return the results
  return results;
});
```

^ Note that we pass executorContext and id through as a tuple, so we get a list of tuples as input and then we must loop over these, group them by executorContext, then execute each group, and finally figure out where in the results to write the result for each `id` back to.

### After

```ts
const $executorContext = executor.context();
const $id = ...;
const $result = loadOne($id, $executorContext, async (ids, { unary: executorContext }) => {
  const { withPgClient, pgSettings } = executorContext;
  const data = await withPgClient(pgSettings, (client) => getTheRecordsByIdsOrWhatever(client, ids));
  return ids.map(id => data.find(d => d.id === id));
});
```

^ Note that here we pass `$executorContext` as a separate, unary, step. Because it's in the unary position we know it will have exactly one value, so we do not need to group by it. Removing the grouping code **massively** simplifies the remaining code, and we end up with the above much shorter plan resolver which does effectively the same thing.

## Performance impact

Unknown.

## Security impact

None known.

## Checklist

### DIFF
--- a/.changeset/stale-schools-press.md
+++ b/.changeset/stale-schools-press.md
@@ -1,0 +1,6 @@
+---
+"grafast": patch
+---
+
+Add unary-step enhanced loadOne and loadMany usage to avoid manual grouping in
+callbacks.
--- a/grafast/grafast/src/steps/load.ts
+++ b/grafast/grafast/src/steps/load.ts
@@ -317,7 +317,7 @@ export class LoadStep<
       meta.cache = cache;
     }
     const batch = new Map<TSpec, number[]>();
-    const unary = values1 && values1.isBatch ? values1.value : undefined;
+    const unary = values1?.isBatch === false ? values1.value : undefined;
 
     const results: Array<PromiseOrDirect<Maybe<TData>>> = [];
     for (let i = 0; i < count; i++) {
--- a/grafast/grafast/src/steps/load.ts
+++ b/grafast/grafast/src/steps/load.ts
@@ -657,7 +657,7 @@ export function loadOne<
       $spec,
       null,
       loadCallbackOrIoEquivalenceOrUnarySpec,
-      loadCallbackOrIoEquivalenceOrUnarySpec as LoadOneCallback<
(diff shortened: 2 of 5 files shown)


benjie:
This needs a test before it can be merged.

PR #2000, opened 2024-03-21
Add example global behavior to disable everything by default

jcgsville:
## Description

This is a small addition to the documentation for something that took me several minutes to figure out. It is something I suspect will be useful for others who want to take the same disabled-by-default approach that I am taking with postgraphile

## Performance impact

None - just docs

## Security impact

None - just docs

## Checklist

@benjie, I imagine we don't need release notes for documentation tweaks?

### DIFF
--- a/postgraphile/website/postgraphile/behavior.md
+++ b/postgraphile/website/postgraphile/behavior.md
@@ -76,6 +76,11 @@ These global defaults can still be overridden by each entity, so they're a
 good way of making wide ranging "default" behaviors without locking yourself in
 too hard.
 
+For example, if you want to run postgraphile with all behaviors disabled by default,
+you can start your defaultBehavior string with `-*`. This would allow you to take the
+approach of requiring intentional engineer decisions on how database schema additions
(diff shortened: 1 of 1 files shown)


jcgsville:
It turns out I'm still getting familiar with the behavior system. Not even sure if I can figure out how to selectively enable the things I want to after using `-*`, so maybe take this suggestion with a large grain of salt

jcgsville:
Only after messing with this for an hour or two did I realize that this list is not exhaustive

https://postgraphile.org/postgraphile/next/behavior#core-behaviors

Should probably specify that

benjie:
This isn’t disabled yet because it’s buggy:
- https://github.com/graphile/crystal/pull/1803

jcgsville:
> This isn’t documented yet because it’s buggy:
> 
> * [Make `defaultBehavior: "-*"` work as expected #1803](https://github.com/graphile/crystal/pull/1803)

Makes sense. I definitely should have skimmed the other open issues first instead of relying on search

PR #2050, opened 2024-05-07
Various major fixes to schema exporting

benjie:
## Description

This PR sets up testing of the exported schemas by running the integration tests again, this time by generating the schema (as before) then exporting the schema, importing it, and using the imported version for running the tests. It's really inefficient right now, making the tests really slow, but worth it for peace of mind. In future this will need to be redeveloped to find all the schema configurations and then write out those schemas in advance and use them to run the tests against.

Then, of course, there were loads of hard to figure out errors/discrepancies which I had to track down. Roughly:

- `grafast`'s `makeGrafastSchema` has been rebuilt to follow better patterns (fixing issues with enums in the process)
- `grafast`'s `makeGrafastSchema` now supports `__inputPlan(){}` on an input object's plans
- `graphile-export` now exports the input object input plan in `typeDefs` mode
- `graphile-export` now detects functions with additional properties (e.g. `isSyncAndSafe: true`) and makes sure these properties are represented in the export
- `graphile-export` now outputs the `serialize`, `parseValue` and `parseLiteral` methods on custom scalars in `typeDefs` mode
- Marks a number of other things as `EXPORTABLE` to fix exports
- Removes `EXPORTABLE` from a couple of things to fix exports
- Fixes an issue with our linting of the exported schema where it might be ignored sometimes
- `ConstantStep` now supports `.get()` and `.at()` patterns
- Fix `enableDeferStream` detection in schema export by looking at directives in schema being exported

Fixes https://github.com/graphile/crystal/issues/2043
Fixes https://github.com/graphile/crystal/issues/2049

## Performance impact

Import/export only.

## Security impact

Improvement: more accurate exports are safer.

## Checklist


postgraphile/postgraphile/__tests__/schema/v4/defaultOptions.1.export.mjs
```diff
@@ -12373,6 +12389,15 @@ const makeArgs64 = (args, path = []) => {
   return selectArgs;
 };
 const resource_compound_type_computed_fieldPgResource = registry.pgResources["compound_type_computed_field"];
+function UUIDSerialize(value) {
+  return "" + value;
+}
```
benjie, line 12394: As an optimization we should have this export as `toString` (the source name) rather than being called `UUIDSerialize` everywhere. But no big deal.

postgraphile/postgraphile/__tests__/schema/v4/defaultOptions.1.export.mjs
```diff
@@ -33133,6 +33440,9 @@ export const plans = {
     inclusive: undefined
   },
   NestedCompoundTypeInput: {
+    "__inputPlan": function NestedCompoundTypeInput_inputPlan() {
+      return object(Object.create(null));
+    },
```
benjie, line 33445: Would be nice to optimize this down to a method syntax.

PR #2073, opened 2024-05-28
feat: call next() from koa middleware for downstream middlewares

aryascripts:
## Description

Koa middlewares run sequentially, and each middleware calls `next()` function after using or updating the `ctx` (KoaContext). The issue with the grafast middleware that is added here ([koa/v2/index.ts](https://github.com/graphile/crystal/blob/91e87ab6516490a4cc7b7fc6400efb7623fbd331/grafast/grafserv/src/servers/koa/v2/index.ts#L70)) is that it only calls `next()` in some scenarios, and not all. In theory, all scenarios of a middleware should call next, unless it's a catastrophic error.

Once this is enabled, it allows the overall koa app to process something _after_ grafast. This could be things like collecting metrics or logging.

## Performance impact
There is no performance impact to graphile or grafast. `next()` will be the next middleware in the chain of middlewares in the Koa app.

## Security impact

unknown, or none.

## Testing
There were no tests here.

~~I need some help with the tests, because I tried to create a new test file for this.Currently it's not committed because I can't verify that it runs.~~
~~When I try running the tests, I get a few different errors. Are there any docs on the set up required for this repository? What databases are needed? What commands to run to run the tests? (`yarn test` doesn't work right after yarn install)~~

I was able to add a test file to test this change.

## Checklist

- [x] If this is a breaking change I've explained why (not breaking)

### DIFF
--- a/.changeset/chatty-geese-shout.md
+++ b/.changeset/chatty-geese-shout.md
@@ -1,0 +1,6 @@
+---
+"grafserv": patch
+---
+
+Koa middleware created in KoaGrafserv class now calls next function in all use
+cases.
--- a/grafast/grafserv/__tests__/koa.test.ts
+++ b/grafast/grafserv/__tests__/koa.test.ts
@@ -1,0 +1,50 @@
+import { constant, error, makeGrafastSchema } from "grafast";
+import Koa, { Context } from "koa";
+import { createMockContext } from "@shopify/jest-koa-mocks";
+
+import { KoaGrafserv } from "./../src/servers/koa/v2";
+
+const schema = makeGrafastSchema({
+  typeDefs: /* GraphQL */ `
+    type Query {
+      hello: String!
+      throwAnError: String
+    }
+  `,
+  plans: {
+    Query: {
+      hello() {
+        return constant("world");
+      },
+      throwAnError() {
+        return error(new Error("You asked for an error... Here it is."));
+      },
+    },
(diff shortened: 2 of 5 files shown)


benjie:
> In theory, all scenarios of a middleware should call next, unless it's a catastrophic error.

That does not seem to align with the Koa documentation:

> Middleware that decide to respond to a request and wish to bypass downstream middleware may simply omit next(). Typically this will be in routing middleware, but this can be performed by any. 
> -- https://github.com/koajs/koa/blob/master/docs/guide.md#response-middleware

My understanding is that "terminal middleware" (or as Koa calls them, response middleware) need not call `next()`; in fact doing so could cause bugs in people's code by having multiple response handlers execute unexpectedly. Koa's nested middleware pattern allows each early middleware to perform actions both before and after any later middleware, for example logging and metrics.

For example:

```ts
app.use(async function (ctx, next) {
  console.log('>> one');
  await next();
  console.log('<< one');
});

app.use(async function (ctx, next) {
  console.log('>> two');
  await next();
  console.log('<< two');
});

app.use(async function (ctx, next) {
  ctx.body = 'hello';
});
```

this would log `>> one`, `>> two`, then it would set the response, then it would log `<< two` and then `<< one`.

You must make sure any logging/metrics middleware is registered _before_ any terminal middlewares that they wish to instrument.

PR #2084, opened 2024-06-05
Make adapter in config the concrete adapter instance, not import path

benjie:
Replaces https://github.com/graphile/crystal/pull/1985 since I can't push to that branch.

This rebases on `main` and will apply the suggestions made during review.


benjie:
Rebased so needed force push :man_facepalming: I should really read error messages more carefully.

PR #2101, opened 2024-06-19
Allow for re-set of a Type as long as we are in the init phase

gitrojones:
## Description

Currently it's possible to alias a custom scalar type with a primitive type (Int, String, etc..) in the init phase with `build.setGraphQLTypeForPgCodec` out of the box with an init hook. However, if you want to alias one scalar type over another it's not currently possible because those scalars don't exist until after `PgCodecsPlugin` runs and once set they cannot be re-set. 

It seems reasonable that a hook set to run after `PgCodecsPlugin` but within the init phase should be able to re-set any of those automatically generated types.

## Performance impact

Unknown

## Security impact

Unknown

## Checklist

### DIFF
--- a/graphile-build/graphile-build-pg/src/plugins/PgBasicsPlugin.ts
+++ b/graphile-build/graphile-build-pg/src/plugins/PgBasicsPlugin.ts
@@ -325,8 +325,9 @@ export const PgBasicsPlugin: GraphileConfig.Plugin = {
 
             const situations_ = Array.isArray(variants) ? variants : [variants];
             for (const situation of situations_) {
-              if (meta.typeNameBySituation[situation] != null) {
-                // TODO: allow this?
+              if ((!build.status.isBuildPhaseComplete || build.status.isInitPhaseComplete) &&
+                  meta.typeNameBySituation[situation] != null) {
+                // TODO: allow this in more phases?
                 throw new Error("Type already set");
               }
               meta.typeNameBySituation[situation] = typeName;


benjie:
This seems to already be possible by setting the type before PgCodecsPlugin executes.

```ts
const makeSetCodecTypePlugin = (
  databaseTypeName: string,
  inputGraphqlTypeName: string,
  outputGraphqlTypeName: string = inputGraphqlTypeName,
): GraphileConfig.Plugin => ({
  name: `SetCodecTypePlugin_${databaseTypeName}_${inputGraphqlTypeName}_${outputGraphqlTypeName}`,
  version: "0.0.0",

  schema: {
    hooks: {
      init: {
        before: ["PgCodecsPlugin"],
        callback(_, build) {
          for (const codec of build.allPgCodecs) {
            console.log(codec.extensions?.pg?.name, databaseTypeName);
            if (codec.extensions?.pg?.name === databaseTypeName) {
              build.setGraphQLTypeForPgCodec(
                codec,
                "input",
                inputGraphqlTypeName,
              );
              build.setGraphQLTypeForPgCodec(
                codec,
                "output",
                outputGraphqlTypeName,
              );
            }
          }
          return _;
        },
      },
    },
  },
});
```

> However, if you want to alias one scalar type over another it's not currently possible because those scalars don't exist until after PgCodecsPlugin runs

You only reference them by name, so this shouldn't matter. That's one of the major reasons that the new system works the way it does: all of the potential types are registered up front, but none are actually created until later.

If this doesn't address your need, please create a reproduction of your issue. I'm not keen to encourage overwriting of things like codecs or GraphQL types; overwriting things makes debugging hard.

PR #2128, opened 2024-07-17
Interim performance optimization for large input arg types

adamni21:
## Description

This PR aims to improve the performance for large input arg types.
Eg. a filter input with many elements in the and/or field.

Discussed in discord: https://discord.com/channels/489127045289476126/498852330754801666/1261317211604123718

Only serves as an interim solution, while `eval*` is being removed/the filter plugin gets rewritten to be evaluated at execution time instead of planning time.

## Performance impact

Vast improvement for the mentioned use case, >500x for a filter with 20 "or filters" on a filter type with ~250 fields.
Maybe a slight performance decrease, for general usage, due to increased operation plan splitting.

## Security impact

Unknown (likely none)

## Checklist

- [ x] My code matches the project's code style and `yarn lint:fix` passes.
- [ X] If this is a breaking change I've explained why.

I didn't add tests, but it should already be covered by `grafast/dataplan-pg/__tests__/queries/conditions/complex-filter.test.graphql`


grafast/grafast/src/steps/__inputObject.ts
```diff
@@ -173,6 +173,31 @@ export class __InputObjectStep<
...
+      throw new Error("evalKeys must only be called for object types");
+    }
+
+    const keys = new Array<
+      (keyof TInputType & string) | `${keyof TInputType & number}`
+    >();
+    const inputFieldKeys = Object.keys(this.inputFields);
+    for (let i = 0; i < inputFieldKeys.length; i++) {
+      const inputFieldPlan = this.inputFields[inputFieldKeys[i]].step;
+      if (inputFieldPlan.eval() !== undefined) {
```
benjie, line 190: `.eval()` will split the plan for every individual value. Use `evalIs(undefined)` instead, then it will at most have two branches: one for `undefined`, and one for everything else.

```suggestion
      if (!inputFieldPlan.evalIs(undefined)) {
```

grafast/grafast/src/steps/__inputObject.ts
```diff
@@ -173,6 +173,31 @@ export class __InputObjectStep<
...
+    (keyof TInputType & string) | `${keyof TInputType & number}`
+  > {
+    if (this.inputValues?.kind !== "ObjectValue") {
+      throw new Error("evalKeys must only be called for object types");
+    }
+
+    const keys = new Array<
+      (keyof TInputType & string) | `${keyof TInputType & number}`
+    >();
+    const inputFieldKeys = Object.keys(this.inputFields);
```
benjie, line 187: We can safely assume this will always just be strings, since no other types are permitted.

```suggestion
    const inputFieldKeys = Object.keys(this.inputFields) as string[];
```

This should then mean we don't need the awkward cast in the `keys.push(...)` below.

grafast/grafast/src/steps/__inputObject.ts
benjie: Note to self: this is required. With `__inputObject` we know that it's an explicit input object (not variable) in the GraphQL document, but the values of each key may still be undefined if they're a variable that isn't supplied and has no default. In these cases we do not wish to return these keys (since the input object is not seen as having those keys set), but that will differ on an operation-to-operation basis, and thus we must evaluate whether or not they are undefined. Note that this implicitly adds constraints for these values; we do not need to explicitly add any constraint for the object itself because the document itself guarantees it will always be present.

Note to self: potential optimization: we should not need to `.evalIs(undefined)` for any input field that is declared as non-nullable, I think?

grafast/grafast/src/steps/__trackedValue.ts
```diff
@@ -366,6 +366,42 @@ export class __TrackedValueStep<
...
+   *
+   * **WARNING**: avoid using this where possible, it causes OpPlans to split.
+   */
+  evalKeys(): Array<(keyof TData & string) | `${keyof TData & number}`> {
+    const { value, path } = this;
+
+    const keys = new Array<
+      (keyof TData & string) | `${keyof TData & number}`
+    >();
+    if (!isInputObjectType(this.nullableGraphQLType) || !value) {
```
benjie, line 382:
```suggestion
    if (!isInputObjectType(this.nullableGraphQLType)) {
```

grafast/grafast/src/steps/__trackedValue.ts
```diff
@@ -366,6 +366,42 @@ export class __TrackedValueStep<
...
+      throw new Error("evalKeys must only be called for object types");
+    }
+    const keysOfType = Object.keys(this.nullableGraphQLType.getFields() ?? {});
```
benjie, line 385:
```suggestion
    const inputObjectType: GraphQLInputObjectType = this.nullableGraphQLType;
    const keysOfType = Object.keys(inputObjectType.getFields());
```

grafast/grafast/src/steps/__trackedValue.ts
```diff
@@ -366,6 +366,42 @@ export class __TrackedValueStep<
...
 
+  /**
+   * Evaluates the keys of the current value, and adds a
+   * constraint to the OpPlan to ensure that all future evaluations of this
+   * check will always return the same result.
+   *
+   * **WARNING**: avoid using this where possible, it causes OpPlans to split.
+   */
+  evalKeys(): Array<(keyof TData & string) | `${keyof TData & number}`> {
+    const { value, path } = this;
```
benjie, line 377:
```suggestion
    const { value, path } = this;
    if (value == null) {
      this.constraints.push({
        type: "keys",
        path,
        keys: null,
      });
      return null;
    }
```

grafast/grafast/src/steps/__trackedValue.ts
```diff
@@ -366,6 +366,42 @@ export class __TrackedValueStep<
...
   }
 
+  /**
+   * Evaluates the keys of the current value, and adds a
+   * constraint to the OpPlan to ensure that all future evaluations of this
+   * check will always return the same result.
+   *
+   * **WARNING**: avoid using this where possible, it causes OpPlans to split.
+   */
+  evalKeys(): Array<(keyof TData & string) | `${keyof TData & number}`> {
```
benjie, line 376:
```suggestion
  evalKeys(): Array<(keyof TData & string) | `${keyof TData & number}`> | null {
```

benjie (changes requested):
This is awesome! Here's some initial changes from a very quick review to help us align on the right direction. In particular, just like with arrays, you need to handle the value not being set - in which case `evalKeys()` should return `null`.

benjie (approved):
Excellent work; thanks for this! I've optimized it a bit, handled more edge cases, and added another test to check handling of partial variables (and undefined variables). I've also fixed some existing bugs I noticed at the same time which somehow have flown under the radar until now.

adamni21:
> Excellent work; thanks for this! I've optimized it a bit, handled more edge cases, and added another test to check handling of partial variables (and undefined variables). I've also fixed some existing bugs I noticed at the same time which somehow have flown under the radar until now.
Lovely.
Happy to help 😀

PR #2166, opened 2024-09-02
Update exporting-schema docs to reference graphile-export docs

mrjackdavis:
## Description

Simply updates docs to reference graphile-export. More documentation updates to come there


mrjackdavis:
Closed in favour of #2167

PR #2167, opened 2024-09-02
Improve graphile export docs

mrjackdavis:
## Description

Raising this PR early for visibility. Open to feedback as I go


utils/website/graphile-export/index.md
```diff
@@ -4,73 +4,24 @@ sidebar_position: 1
...
+- Much faster cold starts for Lambda
+- And probably more
 
-```js
-> (function add(a, b) { return a + b }).toString()
-'function add(a, b) { return a + b }'
-```
+Previously, in Postgraphile 4, export took the form of encoding the Postgres
+schema to JSON. Postgraphile 5 takes this further and generates runtime code,
+further cutting startup times.
```
benjie, line 20: This should not be part of the graphile-export documentation; add it to the PostGraphile pages if you feel it warrants detail (though I'd personally keep it to the V4 migration guide).

benjie: Same with paragraph below

benjie (changes requested):
Thanks for your work on this! It is a good start; but I put in a lot of effort to make sure that the solutions we build are generally useful where they can be - i.e. someone using any other schema generation framework should be able to export their schema as executable using `graphile-export` without even having to know PostGraphile exists. Thus any PostGraphile-specific documentation should remain within PostGraphile's documentation.

Similarly, a tool like `graphile-export`'s documentation should detail fully how to use the tool no matter what type of consumer you are; however PostGraphile's documentation should cover specifically how to use the tool with PostGraphile in a simplified form - for example, if we wanted to demonstrate a PostGraphile plugin to automatically export the schema once it's built ([like this](https://github.com/graphile/crystal/blob/47098fc69468726f4d212fcf1782fbdad880adea/postgraphile/postgraphile/graphile.config.ts#L168-L187)), that would be in the PostGraphile documentation, not the graphile-export docs.

I hope that makes sense!

(Sorry again that it's taken a while to get around to looking at this.)

benjie (approved):

PR #2180, opened 2024-09-20
Remove default filter from WellKnownMap

nckswt:
## Description

I noticed an issue where some modules weren't being exported properly. [Full discord discussion here](https://discord.com/channels/489127045289476126/498852330754801666/1286380009379856465).

### Problem
My source code has:

```ts
import CityTimezones from 'city-timezones';
import snakeCase from 'lodash/snakeCase';

...
export const pgInsert = EXPORTABLE((snakeCase) => async (pgClient, table, data) => {
  const columns = Object.keys(data).map(snakeCase).join(', ');
  ...
}, [snakeCase]);
```

and I'm exporting modules like:

```ts
import * as CityTimezones from 'city-timezones';
import * as lodashSnakeCase from 'lodash/snakeCase';

...
  await exportSchema(schema, exportFileLocation, {
    mode: "typeDefs",
    modules: {
      'city-timezones': CityTimezones,
      'lodash/snakeCase': lodashSnakeCase,
    }
  });
```

But my exported schema has `snakeCase` undefined:

```js
const pgInsert2 = async (pgClient, table, data) => {
  const columns = Object.keys(data).map(undefined).join(", ");
  // !! Note how snakeCase has become undefined /\
```

### Resolution
After [this line](https://github.com/graphile/crystal/blob/d9d175b23a0abaed2aa620440cc7959f4c1e8837/utils/graphile-export/src/wellKnown.ts#L24) I added the following:

```ts
  console.log({ moduleName, exportName, exp: obj[exportName] });
  console.log(exportName !== "default");
  console.log(!wellKnownMap.has(obj[exportName]));
```

and noticed

```ts
{
  moduleName: 'city-timezones',
  exportName: 'lookupViaCity',
  exp: [Function: lookupViaCity]
}
true
true
{
  moduleName: 'lodash/snakeCase',
  exportName: 'default',
  exp: [Function (anonymous)]
}
false
true
```

So, for now, I'm simply removing the `exportName !== "default"` conditional and my schema seems to be exported as expected, but I don't know why that was there in the first place or the other downstream effects.

## Performance impact

unknown, likely none.

## Security impact

unknown, likely none.

## Checklist

### DIFF
--- a/.changeset/quiet-items-divide.md
+++ b/.changeset/quiet-items-divide.md
@@ -1,0 +1,7 @@
+---
+"graphile-export": patch
+---
+
+Fix bug in graphile-export handing modules where default export
+(`import mod from 'mod'`) differed from wildcard export
+(`import * as mod from 'mod'`).
--- a/utils/graphile-export/src/wellKnown.ts
+++ b/utils/graphile-export/src/wellKnown.ts
@@ -21,7 +21,7 @@ function makeWellKnownFromOptions(options: ExportOptions) {
     preferViaDefault = false,
   ) {
     for (const exportName of Object.keys(obj)) {
-      if (exportName !== "default" && !wellKnownMap.has(obj[exportName])) {
+      if (!wellKnownMap.has(obj[exportName])) {
         /**
          * ESM is still a bit flaky, so though `import { foo } from 'bar';` may
          * work in some contexts, in raw Node it's often required to do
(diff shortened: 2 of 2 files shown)


benjie (approved):
Thanks! Hopefully this is a more thorough fix and cleans up some existing code too :+1:

benjie (approved):

PR #2204, opened 2024-10-04
Support ws subscriptions for nuxt and h3 (with doc)

Dodobibi:
## Description

Add the support of ws subscriptions for nuxt and h3

## Performance impact

unknown (not)

## Security impact

unknown (not)

## Checklist


benjie (commented):
Awesome; so happy to see this being updated to a more stable implementation! Some early comments; my main question is around the path to the websocket handler, and why `serv` can't handle that internally rather than having the user write additional code. If there's a good reason behind this it's fine, I just want to know what that reason is (as an outsider unfamiliar with H3/Nuxt).

grafast/grafserv/src/servers/h3/v1/index.ts
```diff
@@ -1,8 +1,8 @@
-import type { IncomingMessage, Server as HTTPServer } from "node:http";
-import type { Server as HTTPSServer } from "node:https";
-import type { Duplex } from "node:stream";
 import { PassThrough } from "node:stream";
 
+//@ts-expect-error type imports.
+import type { Hooks, Peer } from 'crossws';
```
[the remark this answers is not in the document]
Dodobibi: Yes, the lib is typed, but it seems types ar exported as module by default, and grafserv is CommonJS.

I have this error else : 
`The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("crossws")' call instead.
  To convert this file to an ECMAScript module, change its file extension to '.mts' or create a local package.json file with '{ "type": "module" }'.`
Maybe a wrong configuration of my dev environment somewhere ??
  
Because we import only types, this can be ignored.

grafast/website/grafserv/servers/h3.md
```diff
@@ -26,31 +31,9 @@ serv.addTo(app).catch((e) => {
   process.exit(1);
 });
 
+// If you need websocket subscriptions (need h3@^1.13.0);
+app.use("/path_to_ws_handler", defineWebSocketHandler(serv.makeWsHandler()));
```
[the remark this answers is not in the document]
Dodobibi: I do this because there is no option in grafserv for pass the ws endpoint (which isn't necessarily the same, but it can).
Yes, we can delegate this in `addTo` function.
I'll make theses changes.

grafast/grafserv/src/servers/h3/v1/index.ts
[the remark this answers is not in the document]
benjie: Okay, please use this for now then:

```ts
type Hooks = import('crossws').Hooks;
type Peer = import('crossws').Peer;
```

grafast/website/grafserv/servers/h3.md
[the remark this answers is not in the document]
benjie: Hmmm; if that's a useful feature then lets add it here:

https://github.com/graphile/crystal/blob/44ca3dadfa05a86da50875326009fa45662b5aa1/grafast/grafserv/src/index.ts#L109-L112

Something like:

```ts
      /** The path at which the GraphQL websocket would be made available; defaults to `graphqlPath` */
      websocketsPath?: string;
```

Then we should adjust the relevant places to support this option.

Alternatively for now just require that it matches `graphqlPath` and we can change it later.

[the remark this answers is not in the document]
benjie: (I think the latter is what you have done?)

grafast/grafserv/src/servers/h3/v1/index.ts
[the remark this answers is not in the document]
Dodobibi: Same error with your proposal (in french):
image

grafast/website/grafserv/servers/h3.md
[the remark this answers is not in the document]
Dodobibi: As ruru does not yet support a custom ws endpoint: [/grafast/grafserv/src/middleware/graphiql.ts#L37-L39](https://github.com/graphile/crystal/blob/10ed7e83767e0552c3132aeb73d27e66d7d5fc7b/grafast/grafserv/src/middleware/graphiql.ts#L37-L39), it's relevant for now to require it to match graphqlPath and we can change it later 👍 

> (I think the latter is what you have done?)

Yes

grafast/grafserv/src/servers/h3/v1/index.ts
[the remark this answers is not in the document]
benjie: Okay; no worries.

grafast/website/grafserv/servers/h3.md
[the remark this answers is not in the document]
benjie: We'll save the other option for a later PR then.

grafast/grafserv/src/servers/h3/v1/index.ts
```diff
@@ -227,6 +228,13 @@ export class H3Grafserv extends GrafservBase {
         : ["post"],
     );
 
+    if (this.resolvedPreset.grafserv?.websockets) {
+      app.use(
+        this.dynamicOptions.graphqlPath,
+        defineWebSocketHandler(this.makeWsHandler()),
+      );
+    }
```
benjie, line 236: I've modified this so it only kicks in when websockets are enabled - is this the only place to make that change?

grafast/grafserv/src/servers/h3/v1/index.ts
```diff
@@ -242,58 +250,49 @@ export class H3Grafserv extends GrafservBase {
...
+      message(peer, message) {
+        clients.get(peer)?.handleMessage?.(message.text());
+      },
+      close(peer, details) {
+        clients.get(peer)?.closed?.(details.code, details.reason);
+        clients.delete(peer);
+      },
+      error(peer, error) {
+        // clients.delete(peer)?
+      },
```
benjie, line 294: When an `error` occurs, does `close` also get called? I noticed that `clients` is a `Map` but it looked like it grew indefinitely so I added `clients.delete(peer)` to the `close()` event so that we wouldn't leak memory.

grafast/website/grafserv/servers/nuxt.md
```diff
@@ -40,101 +40,25 @@ import { serv } from "@/server/grafserv/serv";
...
-      err
-        ? console.warn("Runtime server wrongly closed", err)
-        : console.log("Runtime server stopped"),
-    );
-  });
-  // Attaching ws to server
-  serv.attachWebsocketsToServer_experimental(server);
-});
+// Create and export the `/api/graphql` websocket handler
+export default defineWebSocketHandler(serv.makeWsHandler());
```
benjie, line 50: Is `defineWebSocketHandler(cb)` effectively the same as `eventHandler({websocket:cb})`?

grafast/website/grafserv/servers/nuxt.md
```diff
@@ -40,101 +40,25 @@ import { serv } from "@/server/grafserv/serv";
...
-export default defineNitroPlugin(async (nitroApp) => {
-  // this hook will be called only once at first http request
-  nitroApp.hooks.hookOnce("request", (event: H3Event) => {
-    const server = event.node.req.socket.server;
-    if (server) {
-      // Attaching ws to server
-      serv.attachWebsocketsToServer_experimental(server);
-    }
-  });
+export default eventHandler({
```
benjie, line 58: Please ensure we have the relevant imports.

benjie: Also this is by far the more common case: that the URL is the same for websockets and not, so I would move it above the `graphql-ws` one - ideally right next to the `/graphql` endpoint itself.

benjie (changes requested):
Looking good! I've done some refactoring, and have some questions

grafast/grafserv/src/servers/h3/v1/index.ts
Dodobibi: Yes, this is the only place.
`addTo` is used only for h3 standalone server.
using with nuxt, we don't call `addTo` but define handlers directly in filesystem router (`server/api/graphql.ts`).

grafast/grafserv/src/servers/h3/v1/index.ts
Dodobibi: clients is a WeakMap, normally, the garbage collector do it's job later, but we can delete peers manually for sure.
Thus, you're right, we need to delete peer's reference too on error.

grafast/website/grafserv/servers/nuxt.md
Dodobibi: > Please ensure we have the relevant imports.

Nuxt has an automatic import system, no need to import `eventHandler`, `defineEventHandler`, `defineWebSocketHandler` into a nuxt application. But we can import them explicitly for clarity.

> Also this is by far the more common case: that the URL is the same for websockets and not, so I would move it above the `graphql-ws` one - ideally right next to the `/graphql` endpoint itself.

Agree with this 👍

grafast/website/grafserv/servers/nuxt.md
Dodobibi: Not quite : `eventHandler({handler:handler, websocket:cb})` is not explicitly documented (maybe not up to date)
see : [https://h3.unjs.io/guide/websocket](https://h3.unjs.io/guide/websocket)

see code on h3 repo:
```ts
/**
 * Define WebSocket event handler.
 *
 * @see https://h3.unjs.io/guide/websocket
 */
export function defineWebSocketHandler(hooks: Partial<WSHooks>) {
  return defineEventHandler({
    handler() {
      throw createError({
        statusCode: 426,
        statusMessage: "Upgrade Required",
      });
    },
    websocket: hooks,
  });
}
```
`defineWebSocketHandler` must be used when only ws on this handler, instead of `eventHandler` (or `defineEventHandler`) when there is at least a standard handler.

Also, it's right to call `defineWebSocketHandler ` in `addTo` because it push this handler **after** the graphql/graphiql handler in the stack, and the _error_ handler will be called only there is no previous handler in the stack (which is not the case because we set graphql/graphiql handler previously).

grafast/grafserv/src/servers/h3/v1/index.ts
benjie: Oh I didn't notice it was a WeakMap. I'd much prefer it was a `Map` and we cleaned up properly than relying on `WeakMap` which can result in spiky memory usage.

grafast/website/grafserv/servers/nuxt.md
benjie: If users of nuxt won't be confused by the lack of imports it's fine - again, I'm unfamiliar with that system. I'm personally not a fan of magic/ambient imports - I like to see where the variables/functions I'm referencing come from.

grafast/grafserv/src/servers/h3/v1/index.ts
Dodobibi: I will make these changes

grafast/website/grafserv/servers/nuxt.md
Dodobibi: I will make these changes too

grafast/website/grafserv/servers/nuxt.md
```diff
@@ -17,124 +17,50 @@ import schema from "./schema.mjs";
 export const serv = grafserv({ schema, preset });
 ```
 
-and the API routes :
+## API routes
+
+### Graphql endpoint
```
benjie, line 22:
```suggestion
### GraphQL endpoint
```

benjie (approved):
Looks good to me; thanks for your work on this.

Note to self: needs a changelog.

benjie (approved):

PR #2218, opened 2024-10-22
docs: documentation for at() and get() methods

### LINKED ISSUE #467: Expected id of elements of field `nodes` to be strings.
Common database id field is not allowed in Relay modern:
https://github.com/facebook/relay/issues/1682

I suggest to generating id as _id.

jemgillam:
## Description

Needs review! 

- detail on implementing `at` and `get` added under the header "Access methods" 
- added missing detail about the methods a ListStep can accept
- Saw a missing link while reviewing the methods for a Lambda step

Closes https://github.com/graphile/crystal-pre-merge/issues/497

### DIFF
--- a/grafast/website/grafast/step-classes.md
+++ b/grafast/website/grafast/step-classes.md
@@ -159,6 +159,54 @@ console.log("$a = " + $a.toString());
 You may override this to add additional data to the `toString` method (the data
 that would occur between the triangular brackets).
 
+## Access methods
+
+When writing a step class, you should implement either `.at()` or `.get()` depending on
+if the step represents a list, array, or an objet.
+
+### at
+
+Implement `.at()` if your step represents a list or an array. It should accept a single argument, an
+integer, which represents the index within the list-like value which should be accessed.
+
+Usage:
+
+```ts
+const $abTuple = list([$a, $b]);
+
+// abTuple.at(0) returns $a
+```
+
+### get
+
+Implement `.get()` if your step represents an object. It should accept a single argument, a
+string, which represents an attribute to access an object-like value.
+
+### access
+
+If optimization isn't a concern, you can defer to `access` while implementing `get` or `at`. This
+means you can refactor and optimize at a later stage without affecting code using that step class.
+
+```ts
+class MyStep extends ExecutableStep {
+  // ...
+
+  // For steps representing objects:
+  get(key) { return access(this, key); }
+
+  // For steps representing lists:
+  at(index) { return access(this, index); }
+```
+
+:::caution
+
+If your step implements `.at()` or `.get()`, make sure it conforms to the expectations of
+those methods: ie it correctly accepts a single argument of either an integer or string.
+&ZeroWidthSpace;<grafast /> relies on these assumptions; unanticipated behaviours may result
+from steps which don't adhere to these expectations.
+
+:::
+
 ## Lifecycle methods
 
 ### execute
--- a/grafast/website/grafast/step-classes.md
+++ b/grafast/website/grafast/step-classes.md
@@ -50,6 +50,72 @@ inherit directly from `ExecutableStep`.
 
 :::
 
+## Conventions
+
+Your step may implement any additional methods that it needs; however certain methods
+have special meaning. For example, if your step represents an object then it should
+implement the `.get(key)` method; and if the step represents an array/list then it
+should implement the `.at(index)` method.
+
+These conventions are still evolving, and more may be added as common usage patterns are
+detected. Functions that have special meanings/expectations can be found below:
+
+### at
+
+Implement `.at()` if your step represents a list or an array. It should accept a single argument, an
+integer, which represents the index within the list-like value which should be accessed.
+
+Usage:
+
+```ts
+import { access } from "grafast";
+
+class MyListStep extends ExecutableStep {
+  // ...
+
+  at(index) {
+    // Your step may implement a more optimized solution here.
+    return access(this, index);
+  }
+}
+```
+
+:::caution
+
+If your step implements `.at()`, make sure it meets the expectations:
+ie it correctly accepts a single argument an integer.
+&ZeroWidthSpace;<grafast /> relies on this assumption; unanticipated behaviours may result
+from steps which don't adhere to these expectations.
+
+:::
+
+### get
+
+Implement `.get()` if your step represents an object. It should accept a single argument, a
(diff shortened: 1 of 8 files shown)


grafast/website/grafast/step-classes.md
```diff
@@ -159,6 +159,54 @@ console.log("$a = " + $a.toString());
 You may override this to add additional data to the `toString` method (the data
 that would occur between the triangular brackets).
 
+## Access methods
+
+When writing a step class, you should implement either `.at()` or `.get()` depending on
+if the step represents a list, array, or an objet.
```
benjie, line 165:
```suggestion
## Conventions

Your step may implement any additional methods that it needs; however certain methods 
have special meaning. For example, if your step represents an object then it should
implement the `.get(key)` method; and if the step represents an array/list then it
should implement the `.at(index)` method.

These conventions are still evolving, and more may be added as common usage patterns are
detected. Functions that have special meanings/expectations can be found below:
```

grafast/website/grafast/step-classes.md
```diff
@@ -159,6 +159,54 @@ console.log("$a = " + $a.toString());
...
+
+Implement `.at()` if your step represents a list or an array. It should accept a single argument, an
+integer, which represents the index within the list-like value which should be accessed.
+
+Usage:
+
+```ts
+const $abTuple = list([$a, $b]);
+
+// abTuple.at(0) returns $a
```
benjie, line 177: Maybe this; what do you think?

```suggestion
import { access } from 'grafast';

class MyListStep extends ExecutableStep {
  // ...

  at(index) {
    // Your step may implement a more optimized solution here.
    return access(this, index);
  }
}
```

grafast/website/grafast/step-classes.md
```diff
@@ -159,6 +159,54 @@ console.log("$a = " + $a.toString());
...
+```
+
+### get
+
+Implement `.get()` if your step represents an object. It should accept a single argument, a
+string, which represents an attribute to access an object-like value.
```
benjie, line 183:
````suggestion
string, which represents an attribute to access an object-like value.

```ts
import { access } from 'grafast';

class MyObjectStep extends ExecutableStep {
  // ...

  get(key) {
    // Your step may implement a more optimized solution here.
    return access(this, key);
  }
}
```
````

File: grafast/website/grafast/step-classes.md [hunk 4]
```diff
@@ -159,6 +159,54 @@ console.log("$a = " + $a.toString());
...
+```ts
+class MyStep extends ExecutableStep {
+  // ...
+
+  // For steps representing objects:
+  get(key) { return access(this, key); }
+
+  // For steps representing lists:
+  at(index) { return access(this, index); }
+```
```
benjie, line 199:
```suggestion
```

grafast/website/grafast/step-classes.md
```diff
@@ -159,6 +159,54 @@ console.log("$a = " + $a.toString());
...
+
+:::caution
+
+If your step implements `.at()` or `.get()`, make sure it conforms to the expectations of
+those methods: ie it correctly accepts a single argument of either an integer or string.
+&ZeroWidthSpace;<grafast /> relies on these assumptions; unanticipated behaviours may result
+from steps which don't adhere to these expectations.
+
+:::
```
benjie, line 208: Please add this under each of "get" and "at" and customize the wording to be relevant to that method.

grafast/website/grafast/step-library/standard-steps/list.md
```diff
@@ -8,3 +8,9 @@ Usage:
 ```ts
 const $abTuple = list([$a, $b]);
 ```
+
+A ListStep has the following methods:
+
+- `.at(index)` - gets the value at `index` (`abTuple.at(0)` returns `$a`).
+- `.first()` - returns the first value in the list (`abTuple.first` returns `$a`).
+- `.last()` - returns the last value in the list (`abTuple.last()` returns `$b`).
```
benjie, line 16:
```suggestion
- `.at(index)` - gets the value at `index` (`$abTuple.at(1)` returns `$b`).
- `.first()` - returns the first value in the list (`$abTuple.first()` returns `$a`).
- `.last()` - returns the last value in the list (`$abTuple.last()` returns `$c`).
```

grafast/website/grafast/step-library/standard-steps/list.md
```diff
@@ -8,3 +8,9 @@ Usage:
 ```ts
 const $abTuple = list([$a, $b]);
```
benjie, line 9:
```suggestion
const $abTuple = list([$a, $b, $c]);
```

benjie (changes requested):
Really good improvements! Here's some suggested edits :raised_hands:

grafast/website/grafast/step-library/standard-steps/list.md
jemgillam: is it still a tuple?

grafast/website/grafast/step-classes.md
jemgillam: Should this then be moved earlier / later in the document? 
At the moment it sits between "Built in methods" and "Lifecycle methods"

grafast/website/grafast/step-library/standard-steps/list.md
benjie: [Indeed](https://en.wikipedia.org/wiki/Tuple). To be more specific we could call it a 3-tuple (or n-tuple).

benjie: But it should probably be `$abcTuple` now.

grafast/website/grafast/step-classes.md
benjie: It's repetitive if you read the whole page. If you jump right to the relevant part of the docs (e.g. for the `get` method) then it would be easy to miss if it were shared at top or bottom.

You can set the title of the warnings:

```
:::caution Any `get` method on a step must conform!
```

grafast/website/grafast/step-classes.md
jemgillam: Moved it higher.

jemgillam:
I've reorgainsed the sections on the step-classes page as discussed

I've added a blurb to the https://grafast.org/grafast/step-library/standard-steps/ page as it is linked to directly from the step-classes page, I wrote a small summary taken from the step-library category page

PR #2225, opened 2024-11-01
[draft] fix: constraints unit test

aryascripts:
## Description
This is only an example to explain the issue in #2226

Screenshot 2024-11-01 at 3 04 48 PM


benjie:
I appreciate your work on this, but before changing the behavior of constraint matching I'd like to know what it is that's causing this constraint to exist in the first place, because whatever code that is, is likely wrong. See this comment for more context:

https://github.com/graphile/crystal/issues/2226#issuecomment-2454199650

benjie:
Seems the underlying issue was already solved by #2129

PR #2262, opened 2024-12-02
website: Migrate over plugin gallery

jemgillam:
## Description

Migrating over the plugin gallery from https://www.graphile.org/postgraphile/plugin-gallery/ I'm not sure how this one got overlooked. 

It only exists on the Version 4 version of the docs

It mirrors the other galleries, where it is over a number of pages within a subcategory

## Checklist


benjie (approved):
Thanks! We should do a final pass to see if we can now redirect graphile.org/postgraphile to postgraphile.org :raised_hands:

PR #2266, opened 2024-12-03
Reconnect eventSource after connection is closed

hos:
## Description

This fixes an issue when ruru SSE is connected to the server using protocol later then `HTTP/1.1`. When using through `localhost` you probably won't have any issue as it is uses `HTTP/1.1`, but when running using caddy or some tunnel, some of them use `HTTP/2` or `HTTP/3`. The header `connection: keep-alive` is not supported on protocols later then `HTTP/1.1` and the behavior of reconnect is different. In my case, on server restart as soon as client (ruru) loses the connection it throws an error `throw new Error("Lifecycle management error for EventSource");` and the react render crashes (white screen). After that, when trying to reconnect, if the server is still not available (5xx error), the client stops retrying.

1. Fixed by replacing `throw new Error` with `console.log` combined with `setError`.

2. Can be fixed by passing `{reconnectInterval: 2000, maxReconnectAttempts: 10}` to the newly added `eventSourceInit` option on the `preset.ruru`.

 

## Performance impact

Not expected to have any.

## Security impact

Not introducing anything new.

## Checklist


benjie (approved):
This is great; thanks for tracking this down @hos it's been bothering me for a while but I've not had the time to look into it!

Issue #2060, opened 2024-05-10
Epic: removing eval


benjie:
I've never liked the `eval` system, but it seemed a requirement in the way that Grafast was architected.

As it turns out, a lot of the need for eval should be solved by the "global dependencies" feature we released in beta 22 (i.e. unary steps #1973).

Getting rid of eval entirely would be a massive step in the right direction for Grafast, it would take us closer to "one operation has one plan" rather than "one operation has at least one and potentially infinite plans" which is the current status quo.

However, this will be a significant undertaking because we need to rewrite a whole bunch of stuff to do with PgSelectStep and other steps that rely on eval currently. We may also need to factor in alternative approaches to plan resolvers. And even more so, we may need to factor it into query planning itself, since query planning performs evals to find out if certain paths are required or not - hopefully those can be addressed by "early exit" in combination with conditional OutputPlan fields.

Here's a rough plan I outlined for this:

---

Move connections/collections over to using "unary" steps rather than eval; we need this so that we can inhibit zeroes (but also because eval is evil):

* [ ] Make the limit (in the SQL) a placeholder rather than a literal
    * [ ] Can we actually have LIMIT 7 as a placeholder? This would be ideal because if the argument is null we want to omit LIMIT entirely rather than adding LIMIT 9999999999 or whatever
* [ ] At execute, get the associated unary step’s value, and look it up in an LRU cache
* [ ] If no match, build SQL using this new placeholder and write it to the LRU cache
* [ ] Execute as normal.

---

Change @skip/@include from using eval to instead using unary steps and early exit:

* [ ] At a skip/include boundary, all future steps should be STOPPED (like for polymorphism) if the skip/include doesn’t match.
* [ ] Output plans should factor in the skip/include into their logic; they can do this inline without the need to create duplicate output plans.
* [ ] As an optimization: we could perform the skip/include in OutputPlans using a bitmask; this would mean we could have at most 32 skip/include variables but that should be enough for most people. Failing that, we can do it as an array of ints and then go arbitrarily large.

---

Make it so argument "undefined" doesn’t use eval either...

* [ ] This one is going to be a challenge!

---

With unary steps and early exit, perhaps we can get rid of the need for eval entirely?

* [ ] Have eval log an error the first time it is used in a new position (development only!)
* [ ] Try and remove all other internal usages.
* [ ] Update documentation to remove references to eval.

---

Note, if we achieve this then we should close #1950.

benjie:
[...]

PR #2267, opened 2024-12-03
Change from json_agg to array()

benjie:
This is an optimization stub... also a compatibility stub since we shouldn't be using the JSON type since it won't be compatible with other adaptors...

Will return to this at a later time.


benjie:
Managed to accomplish this in the #2060 work, and the SQL queries are generally faster, but Node.js handling of results was slower so I had to also do #2397. But anyway... Solved!

PR #2268, opened 2024-12-04
funding.json

jemgillam:
## Description

I have begun a `funding.json` file. I have kept the comments in it for now, but they will need removing before the final merge into the codebase. Some `well-known` links will need to be created across our websites. 

I chose Graphile Worker and PostGraphile as our projects, we can add more, but I feel Grafast might be too new. The whole thing links to Graphile as an organisation any way. 

I put in our plans from $100/mo and above.

See https://floss.fund/funding-manifest/

This is a new initiative from https://zerodha.tech/ 
They launched it last month, the first funding cycle from them is going to start in the new year. 

If this initiative gets momentum, then it is worth having a `funding.json` file anyway, even if FLOSS/fund don't grant us any sponsorship. 

Here is a directory of others who have already set this up: https://dir.floss.fund/browse/projects

Here are the tags we can choose from for our projects: https://floss.fund/static/project-tags.txt


jemgillam:
Moved to https://github.com/graphile/graphile.github.io/pull/424

PR #2271, opened 2024-12-05
websites: funding manifests

jemgillam:
## Description

Funding manifests at `.well-known` URLs for each of our websites, to work in tandem with https://github.com/graphile/graphile.github.io/pull/424


benjie (changes requested):
If you run `yarn start` in the relevant website folder can you browse to these files on the local server? I suspect not - I _think_ you need to move them inside the `public` folder?

PR #2274, opened 2024-12-07
Update `usage-schema.md` example

hos:
## Description

The documentation was using deprecated interfaces. With this update

1. It uses new `hookArgs`, where you can pass `resolvedPreset`,  `requestContext` with the args.
2. It uses `TypedDocumentNode` from `@graphql-typed-document-node/core`, to support fully typed queries when using graphql-codegen's `client-preset`.

Note:

The return type of `execute` function is `PromiseOrValue<ExecutionResult | AsyncGenerator<AsyncExecutionResult, void, undefined>>`, but I didn't found `AsyncExecutionResult` so the type in the current example supports only `ExecutionResult`.

## Checklist

### DIFF
--- a/postgraphile/website/postgraphile/usage-schema.md
+++ b/postgraphile/website/postgraphile/usage-schema.md
@@ -104,31 +104,37 @@ call `hookArgs()`.
 Here's a full example:
 
 ```ts
-import { makeSchema } from "postgraphile";
-import { parse, validate } from "postgraphile/graphql";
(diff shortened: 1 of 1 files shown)


benjie:
Thanks! I extracted the refactor from `makeSchema()` to `postgraphile()` into #2275, and then re-implemented the TypedDocumentNode example via a separate PR #2277 - let me know what you think!

PR #2279, opened 2024-12-10
Introduce step cache to reduce load on deduplicate

benjie:
## Description

Deduplicate is expensive; this new step cache enables the duplicate steps to not be created in the first place, reducing the load on deduplicate. By carefully controlling the way in which steps can be cached (and when these caches need to be cleared) we can enable this functionality safely without risking the complexities that would occur with more naive approaches.

## Performance impact

Should reduce planning time for most queries, however the act of writing to the cache does incur some performance penalty so it's certainly possible that some queries will be slower than before. I expect them to be faster in the vast majority of cases.

## Security impact

Reduced memory usage (and time usage) during planning should reduce security threats slightly.

## Checklist

### DIFF
--- a/.changeset/thick-mugs-trade.md
+++ b/.changeset/thick-mugs-trade.md
@@ -1,0 +1,8 @@
+---
+"postgraphile": patch
+"@dataplan/pg": patch
+"grafast": patch
+---
+
+Introduce step caching to reduce deduplication workload safely, thereby reducing
+planning time for many larger queries.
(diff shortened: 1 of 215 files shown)


benjie:
This reduces the planning time of [@slaskis' reproduction](https://github.com/graphile/crystal/issues/2238#issuecomment-2531691192) from ~6.25s to ~1.26s - i.e. it's now 5x as fast.

benjie:
(The actual solution to @slaskis' issue is #2060; but this fix is welcome nonetheless.)

PR #2288, opened 2024-12-23
Added whatwg-node-server adapter for grafserv

kzlar:
## Description
@whatwg-node/server is an adapter that helps create generic servers across runtimes (https://www.npmjs.com/package/@whatwg-node/server). It could, at least in theory, replace all of the servers currently implemented, while supporting several more. If this all works out it should help integrations while having to maintain less code. 

Note: This doesn't support web-sockets, I'm still looking for a way to do that in a generic fashion, but I see some of the other adapters don't support web-sockets either so I believe this is good enough for now.

## Performance impact

unknown, but this PR won't affect anything by itself, it's just adding the adapter.

## Security impact

None

## Checklist
- [X] I've added tests for the new feature, and `yarn test` passes (Well, sort of - I ran the grafserv tests with it and it passed)

### DIFF
--- a/grafast/grafserv/__tests__/exampleServer.ts
+++ b/grafast/grafserv/__tests__/exampleServer.ts
@@ -4,7 +4,7 @@ import type { AddressInfo } from "node:net";
 import { constant, error, makeGrafastSchema } from "grafast";
 import { resolvePreset } from "graphile-config";
 
-import { grafserv } from "../src/servers/node/index.js";
+import { grafserv } from "../src/servers/whtatwg-node-server";
 
 export async function makeExampleServer(
   preset: GraphileConfig.Preset = {
@@ -36,8 +36,7 @@ export async function makeExampleServer(
   });
 
   const serv = grafserv({ schema, preset });
-  const server = createServer();
-  serv.addTo(server);
+  const server = createServer(serv.createHandler());
   const promise = new Promise<void>((resolve, reject) => {
     server.on("listening", () => {
       server.off("error", reject);
--- a/grafast/grafserv/package.json
+++ b/grafast/grafserv/package.json
@@ -115,6 +115,7 @@
     "@types/koa": "^2.13.8",
     "@types/koa-bodyparser": "^4.3.10",
     "@whatwg-node/fetch": "^0.9.10",
+    "@whatwg-node/server": "^0.9.64",
     "express": "^4.20.0",
     "fastify": "^4.22.1",
     "grafast": "workspace:^",
--- a/grafast/grafserv/src/servers/whtatwg-node-server/index.ts
+++ b/grafast/grafserv/src/servers/whtatwg-node-server/index.ts
@@ -1,0 +1,115 @@
+import { createServerAdapter } from '@whatwg-node/server'
+
+import { GrafservBase } from "../../core/base.js";
+import type {
+  GrafservConfig,
+  RequestDigest,
+  Result,
+} from "../../interfaces.js";
+
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace Grafast {
+    interface RequestContext {
+        whatwg: {
+            request: Request
+        }
+    }
+  }
+}
+
+/** @experimental */
+export class WhatwgGrafserv extends GrafservBase {
+  protected whatwgRequestToGrafserv(
+    request: Request
+  ): RequestDigest {
+    const url = new URL(request.url);
+    return {
+      httpVersionMajor: 1,
+      httpVersionMinor: 0,
+      isSecure: url.protocol === 'https://',
+      method:  request.method,
+      path: url.pathname,
+      headers: this.processHeaders(request.headers),
+      getQueryParams() {
+        return Object.fromEntries(url.searchParams.entries()) as Record<string, string>;
+      },
+      async getBody() {
+        const text = await request.text()
+        return {
+          type: "text",
+          text,
+        };
+      },
+      requestContext: {
+        whatwg: {request}
+      },
+      preferJSON: true,
+    };
+  }
+
+  protected processHeaders(headers: Headers): Record<string, string> {
+    const headerDigest: Record<string, string> = Object.create(null);
+    headers.forEach((v,k)=> {
+      headerDigest[k]= v
+    })
+    return headerDigest
+  }
+
+  protected grafservResponseToWhatwg(response: Result | null): Response {
+    if (response === null) {
+      return new Response("¯\\_(ツ)_/¯", {status: 404, headers: new Headers({"Content-Type": "text/plain"})})
+    }
+
+    switch (response.type) {
+      case "error": {
+        const { statusCode, headers, error } = response;
+        const respHeaders = new Headers(headers)
+        respHeaders.append("Content-Type", "text/plain")
+        return new Response(error.message, {status: statusCode, headers:respHeaders})
+      }
+
+      case "buffer": {
+        const { statusCode, headers, buffer } = response;
+        const respHeaders = new Headers(headers)
+        return new Response(buffer.toString('utf8'), {status: statusCode, headers:respHeaders})
+      }
+
+      case "json": {
+        const { statusCode, headers, json } = response;
+        const respHeaders = new Headers(headers)
+        return new Response(JSON.stringify(json), {status: statusCode, headers:respHeaders})
+      }
+
+      default: {
+        console.log("Unhandled:");
+        console.dir(response);
+        return new Response("Server hasn't implemented this yet", {status: 501, headers: new Headers({"Content-Type": "text/plain"})})
+      }
+    }
+  }
+
+  createHandler() {
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    return createServerAdapter(async (request: Request): Promise<Response> => {
+      return this.grafservResponseToWhatwg(
+        await this.processWhatwgRequest(
+          request,
+          this.whatwgRequestToGrafserv(request),
+        ),
+      );
+    })
+  }
+
+  protected processWhatwgRequest(
+    _request: Request,
+    request: RequestDigest,
+  ) {
+    return this.processRequest(request);
+  }
+}
+
+/** @experimental */
+export function grafserv(config: GrafservConfig) {
+  return new WhatwgGrafserv(config);
+}


grafast/grafserv/__tests__/exampleServer.ts
```diff
@@ -4,7 +4,7 @@ import type { AddressInfo } from "node:net";
 import { constant, error, makeGrafastSchema } from "grafast";
 import { resolvePreset } from "graphile-config";
 
-import { grafserv } from "../src/servers/node/index.js";
+import { grafserv } from "../src/servers/whtatwg-node-server";
```
benjie, line 7: Typo? Should be `whatwg`?

Also please don't update the example server; the current one does not add additional dependencies (just uses `node`) and thus is preferred.

grafast/grafserv/__tests__/exampleServer.ts
```diff
@@ -36,8 +36,7 @@ export async function makeExampleServer(
   });
 
   const serv = grafserv({ schema, preset });
-  const server = createServer();
-  serv.addTo(server);
+  const server = createServer(serv.createHandler());
```
benjie, line 39: Please don't make this change, it does not handle websockets and other concerns. It means that `serv` is not passed an instance of `server` and thus cannot add any event listeners.

grafast/grafserv/package.json
```diff
@@ -115,6 +115,7 @@
     "@types/koa": "^2.13.8",
     "@types/koa-bodyparser": "^4.3.10",
     "@whatwg-node/fetch": "^0.9.10",
+    "@whatwg-node/server": "^0.9.64",
```
benjie, line 118: Looking at the code, this should also be an optional peer dependency, otherwise `createServerAdaptor` cannot be imported? (The other adaptors do not require anything but types be imported.)

grafast/grafserv/src/servers/whtatwg-node-server/index.ts
```diff
@@ -0,0 +1,115 @@
...
+
+/** @experimental */
+export class WhatwgGrafserv extends GrafservBase {
+  protected whatwgRequestToGrafserv(
+    request: Request
+  ): RequestDigest {
+    const url = new URL(request.url);
+    return {
+      httpVersionMajor: 1,
+      httpVersionMinor: 0,
```
benjie, line 29: 1.0 seems unlikely, that's ancient! Are we not able to determine the HTTP version from a WHATWG server?

grafast/grafserv/src/servers/whtatwg-node-server/index.ts
```diff
@@ -0,0 +1,115 @@
...
+      isSecure: url.protocol === 'https://',
```
benjie, line 30: Are you sure this is correct? The protocol is actually `https:`, the `//` is not part of the protocol itself (see for example `location.protocol` in a browser).

https://developer.mozilla.org/en-US/docs/Web/API/URL/protocol

grafast/grafserv/src/servers/whtatwg-node-server/index.ts
```diff
@@ -0,0 +1,115 @@
...
+      headers: this.processHeaders(request.headers),
+      getQueryParams() {
+        return Object.fromEntries(url.searchParams.entries()) as Record<string, string>;
+      },
+      async getBody() {
+        const text = await request.text()
+        return {
+          type: "text",
+          text,
+        };
```
benjie, line 42: Are we certain the body will always be text?

grafast/grafserv/src/servers/whtatwg-node-server/index.ts
```diff
@@ -0,0 +1,115 @@
...
+      },
+      requestContext: {
+        whatwg: {request}
```
benjie, line 45: This should include a version number: `whatwgv1` or similar.

grafast/grafserv/src/servers/whtatwg-node-server/index.ts
```diff
@@ -0,0 +1,115 @@
...
+      },
+      preferJSON: true,
```
benjie, line 47: Disappointing, this opts out of some performance enhancements.

grafast/grafserv/src/servers/whtatwg-node-server/index.ts
```diff
@@ -0,0 +1,115 @@
...
+        const { statusCode, headers, error } = response;
+        const respHeaders = new Headers(headers)
+        respHeaders.append("Content-Type", "text/plain")
+        return new Response(error.message, {status: statusCode, headers:respHeaders})
+      }
+
+      case "buffer": {
+        const { statusCode, headers, buffer } = response;
+        const respHeaders = new Headers(headers)
+        return new Response(buffer.toString('utf8'), {status: statusCode, headers:respHeaders})
```
benjie, line 75: There's no guarantee that the buffer can safely be represented as a utf8 string; is it not possible to return binary data directly to the response?

grafast/grafserv/src/servers/whtatwg-node-server/index.ts
```diff
@@ -0,0 +1,115 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    return createServerAdapter(async (request: Request): Promise<Response> => {
+      return this.grafservResponseToWhatwg(
+        await this.processWhatwgRequest(
+          request,
+          this.whatwgRequestToGrafserv(request),
+        ),
+      );
+    })
+  }
```
benjie, line 102: Please also implement an `addTo` method to follow convention; e.g.:

https://github.com/graphile/crystal/blob/36c41e80bfe312f584af42545688fb78d3adaf3f/grafast/grafserv/src/servers/node/index.ts#L290-L298

benjie (changes requested):
Thanks for starting work on this; it would be great to add @whatwg-node/server as a supported server framework!

Please be sure to format your code; you can run `yarn lint:fix` in the root. May be worth installing a `prettier` extension into your editor of choice and configuring it so that it happens automatically on save.

We will also need some tests before this can be merged. Probably makes sense for `makeExampleServer()` in this file:

https://github.com/graphile/crystal/blob/36c41e80bfe312f584af42545688fb78d3adaf3f/grafast/grafserv/__tests__/graphql-http.test.ts

to accept an argument choosing the server framework to use - then we can test it on each different adaptor easily using the same test suite.

grafast/grafserv/__tests__/exampleServer.ts
kzlar: Correct, fixed

grafast/grafserv/package.json
kzlar: Done

grafast/grafserv/src/servers/whtatwg-node-server/index.ts
kzlar: Yeah I didn't like it either but couldn't find a way to get the HTTP version from a WHATWG Request object. Changed it to 1.1 but still hardcoded unfortunately

grafast/grafserv/src/servers/whtatwg-node-server/index.ts
kzlar: Correct, fixed.

grafast/grafserv/src/servers/whtatwg-node-server/index.ts
kzlar: Done

grafast/grafserv/src/servers/whtatwg-node-server/index.ts
kzlar: I'm honestly not sure about this. How can I tell if this can be set to false?

grafast/grafserv/src/servers/whtatwg-node-server/index.ts
kzlar: Good catch, returning `buffer` as is.

grafast/grafserv/__tests__/exampleServer.ts
kzlar: I'm not sure this would work, wouldn't that make the WHATWG adapter reliant on node types?
I also see that the lambda adapter does not have an addTo implementation.

For now I changed the `makeExampleServer` function to receive a type parameter to revert the node behavior to addTo while keeping the different behavior for whatwg. Let me know if that works.

grafast/grafserv/src/servers/whtatwg-node-server/index.ts
kzlar: See this comment https://github.com/graphile/crystal/pull/2288#discussion_r1896784441

grafast/grafserv/src/servers/whtatwg-node-server/index.ts
kzlar: Tried changing this based on the other adapters, let me know if this looks better.

grafast/grafserv/__tests__/exampleServer.ts
benjie: > I also see that the lambda adapter does not have an addTo implementation.

There is no lambda adaptor. There's a README for it, but that was written before the `addTo` pattern was established; every adaptor currently implements `addTo` I think?

> wouldn't that make the WHATWG adapter reliant on node types?

Are there alternatives to `createServer()` you'd use here? Perhaps you can expand the examples to show why this would be a concern?

grafast/grafserv/src/servers/whtatwg-node-server/index.ts
benjie: That'll do for now. May be a bridge we have to cross at some point though, because e.g. chunked encoding is not supported in HTTP/2+

grafast/grafserv/src/servers/whtatwg-node-server/index.ts
benjie: Is there a common interface that all the servers support that would allow for this? I'd rather consistency if possible, but if we have to do it a different way for this one adaptor that's fine. Another option would be to have `addTo` only support Node-like servers, and other servers would need to use a lower level API.

grafast/grafserv/src/servers/whatwg-node-server/index.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+  });
+}
+
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace Grafast {
+    interface RequestContext {
+        whatwg: {
+            version:string
```
benjie, line 52: Sorry; I meant that the version should be part of the scope key like with the other servers; then when a new version is supported we can add support for that without breaking existing users. See the Koa, Express, Fastify, Lambda and H3 adaptors for example.

```suggestion
        whatwgv0_9: {
```

grafast/grafserv/src/servers/whatwg-node-server/index.ts
```diff
@@ -0,0 +1,154 @@
...
+      headers: this.processHeaders(request.headers),
+      getQueryParams() {
+        return Object.fromEntries(url.searchParams.entries()) as Record<string, string>;
+      },
+      async getBody() {
+        return getBodyFromRequest(request, dynamicOptions.maxRequestLength)
+      },
+      requestContext: {
+        whatwg: {
+          version:'whatwgv1',
```
benjie, line 81:
```suggestion
        whatwgv0_9: {
```

benjie (commented):
Please use `prettier` to format the code.

grafast/grafserv/src/servers/whatwg-node-server/index.ts
benjie: This should also be factored into the file name; e.g. `src/servers/whatwg-node-server/v0_9/...` - see the same list of other adaptors for examples.

benjie (approved):
Awesome; thanks for getting us started on this! :raised_hands:

PR #2290, opened 2024-12-27
adding a callout to graphile config docs re merging exists-but-undefined options

jcgsville:
## Description

Addresses https://github.com/graphile/crystal/issues/2286

When reading through the middleware code, I also noticed one variable was somewhat confusingly named. I recommend changing it to activityName. Happy to do it in another PR or not do it if you'd rather not mix docs with this nitpick code change

Note that running `yarn prettier 'utils/website/graphile-config/preset.md' --write` with the `proseWrap: "preserve",` commented out also seemed to adjust a few newlines.

## Performance impact

Just docs

## Security impact

None

## Checklist

### DIFF
--- a/utils/website/graphile-config/preset.md
+++ b/utils/website/graphile-config/preset.md
@@ -194,15 +194,15 @@ below.
 
 :::note Merging exists-but-undefined options
 
-When extending presets, Graphile Config merges options scopes with
-`Object.assign({}, baseScope, extendingScope)`. This has the following
-implications:
+When extending presets, Graphile Config merges options scopes with logic
+equivalent to `Object.assign({}, baseScope, extendingScope)`. This has the
+following implications:
 
-1. When an option is set to `undefined` or `null`, it will overwrite the value
-   in the extended preset. Depending on the library behavior, this may result in
-   the library default being used for the option.
-2. When an option is not set in a scope, the resolved value will be that of the
-   extended preset.
+1. When an option is set to `undefined` or `null` in the extending preset, it
+   will override the value in the base preset. Depending on the library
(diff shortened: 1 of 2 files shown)


utils/website/graphile-config/preset.md
```diff
@@ -191,10 +192,24 @@ below.
       `extendingScope` actually exist.
 1. Return `mergedPreset`.
 
+:::note Merging exists-but-undefined options
+
+When extending presets, Graphile Config merges options scopes with
+`Object.assign({}, baseScope, extendingScope)`. This has the following
+implications:
```
benjie, line 199:
```suggestion
When extending presets, Graphile Config merges options scopes with logic
equivalent to `Object.assign({}, baseScope, extendingScope)`. This has the
following implications:
```

utils/website/graphile-config/preset.md
```diff
@@ -191,10 +192,24 @@ below.
...
+
+1. When an option is set to `undefined` or `null`, it will overwrite the value
+   in the extended preset. Depending on the library behavior, this may result in
+   the library default being used for the option.
+2. When an option is not set in a scope, the resolved value will be that of the
+   extended preset.
```
benjie, line 205:
```suggestion
1. When an option is set to `undefined` or `null` in the extending preset, it
   will override the value in the base preset. Depending on the library
   behavior, this may result in the library default being used for the option.
2. When an option is _not present_ in the extending scope, the resolved value
   will be that of the option in the base scope (if any).
```

benjie (changes requested):
I've tweaked the wording a little, WDYT?

PR #2294, opened 2025-01-02
makeExtendSchemaPlugin: Improve typeDefs-related error messages

the-sun-will-rise-tomorrow:
## Description

- Make it clear that the `gql` helper needs to be used to construct `typeDefs`, not the entire object.

- If `typeDefs` was missing entirely, produce an appropriate error message.

(Tried to start out with a "minimal" plugin which does nothing and got this confusing error message.)

## Performance impact

None?

## Security impact

None?

## Checklist


benjie (changes requested):
Awesome; thanks!

benjie (dismissed):

benjie (approved):

PR #2297, opened 2025-01-03
Initital Hono grafserv adapter

olup:
## Description

This PR adds a initial Hono adapter to grafserv. I won't handle bufferStream or websockets yet.

## Performance impact

n/a

## Security impact

unknown

## Checklist

- [?] I have added this feature to 'Pending' in the `RELEASE_NOTES.md` file (if one exists).


grafast/grafserv/src/servers/hono/v4/index.ts
```diff
@@ -0,0 +1,254 @@
...
+  Result,
+} from "../../../interfaces.js";
+import { GRAPHQL_TRANSPORT_WS_PROTOCOL, makeServer } from "graphql-ws";
+import { UpgradeWebSocket } from "hono/ws";
+import { web } from "webpack";
+
+declare global {
+  namespace Grafast {
+    interface RequestContext {
+      Hono: {
```
benjie, line 25: This should include a version number and I think the others are lower case so it should match that too:

```suggestion
      honov4: {
```

grafast/grafserv/src/servers/hono/v4/index.ts
```diff
@@ -0,0 +1,254 @@
...
+        honoContext: HonoContext;
```
benjie, line 26: Is it necessary to name this `honoContext`? In the Koa adaptor we just call it `ctx` because that's how the Koa docs refers to it.

grafast/grafserv/src/servers/hono/v4/index.ts
```diff
@@ -0,0 +1,254 @@
...
+  }
+}
+
+function getDigest(honoContext: HonoContext): RequestDigest {
+  const req = honoContext.req;
+  const res = honoContext.res;
+  return {
+    httpVersionMajor: 1, // Hono uses Fetch API, which doesn't expose HTTP version
+    httpVersionMinor: 1,
+    isSecure: req.url.startsWith("https"),
```
benjie, line 38:
```suggestion
    isSecure: req.url.startsWith("https:"),
```

grafast/grafserv/src/servers/hono/v4/index.ts
```diff
@@ -0,0 +1,254 @@
...
+      if (!json) {
+        throw new Error("Failed to retrieve body from hono");
+      }
+      return {
+        type: "json",
+        json,
+      } as GrafservBodyJSON;
+    },
+    requestContext: {
+      hono: {
```
benjie, line 56: The case you had here didn't match what you added in the types (`Hono`)?

```suggestion
      honov4: {
```

grafast/grafserv/src/servers/hono/v4/index.ts
```diff
@@ -0,0 +1,254 @@
...
+        context: honoContext,
```
benjie, line 57: You called this `honoContext` above, but `context` here?

benjie (commented):
Happy to see work being done on this! Suggest that you find other people using hono to validate the approach - feel free to ask around in #watercooler on the Discord.

grafast/grafserv/src/servers/hono/v4/index.ts
olup: Most of the hono doc uses just "c" but isn't it too obscure? ctx seems a good middleground indeed.

benjie: I would use whatever hono docs readers would expect, so c is fine.

Sofianel5:
Would love to see this merged!

benjie (approved):

benjie:
I've added some work to this, notably I removed the `node: { req, res }` fields from the context and removed the deprecated `handleEvent` method (why introduce a deprecated method?). Please review the commits that I did above one-by-one and let me know if you're happy with the changes. If so, we can get this merged! Thanks @olup for your work on this, and @JendaT for your review :+1:

grafast/grafserv/examples/example-hono.mjs
```diff
@@ -0,0 +1,23 @@
...
+// Create a Node HTTP server
+const app = new Hono();
+
+// Create a Grafserv instance
+// the second argument is an optional websocket upgrade handler
+// see https://hono.dev/docs/helpers/websocket
+const serv = grafserv({ schema, preset });
+
+// Mount the request handler into a new HTTP server
+serv.addTo(server).catch((e) => {
```
hobbescodes, line 17: First of all love the work here! Out of curiosity, is this supposed to mount to the previously defined Hono HTTP server (app)? I would think this line should be:

```suggestion
serv.addTo(app).catch((e) => {
```

Just was looking over the code changes and noticed this discrepancy (in comparison to the added test file logic).

benjie: Good catch!

benjie (approved):

PR #2309, opened 2025-01-13
website: Graphile Engine 4.0 docs migrated to build.graphile.org

jemgillam:
## Description

I noticed we have yet to migrate over the Graphile Engine docs. This is needed before we refresh graphile.org. 

The docs on build.graphile.org are now versioned. "Current" is 4.x and "Next" is the new version. This reflects how postgraphile.org is currently. 

Almost all of the content is a straight copy from https://www.graphile.org/graphile-build/ with exceptions:

- The introduction has a bit at the top taken from https://github.com/graphile/graphile-engine/tree/v4/packages/graphile-build with some new copy explaining the name change to Graphile Build

- graphile-build-pg still has its own section, so has its own place within the codebase for versioning. The wording in the overview has been taken from https://github.com/graphile/graphile-engine/tree/v4/packages/graphile-build-pg

- I have sneaked in some tweaks to the dark mode contrast

## To fix

There's two things I think would be nice to fix, but this can be reviewed and merged without: 

1. There are now two dropdowns for the versioning. One for each project (graphile-build and graphile-build-pg). I think it would be nice to put in vertical dividers in the nav bar between the items, but apparently this isn't possible with docusaurus as it only allows specific items in the navbar. Perhaps it can be done with CSS borders. Another hacky option which I think is too overkill is detailed here: https://stackoverflow.com/questions/69173994/docusaurus-2-how-to-add-custom-react-component-in-navbar

Screenshot 2025-01-13 at 17 17 27

2. Our old site has a neat solution to tidy up the table of contents where arguments in brackets are replaced with `...`. It would be nice to replicate this: 

OLD
Screenshot 2025-01-13 at 17 18 11

NEW
Screenshot 2025-01-13 at 17 17 59

3. I've noticed postgraphile.org has nice styling for the highlighted line in a code block, I'd like to replicate that here.


benjie:
Looks great; just needs the dropdown issue fixed (as discussed in person) then we can merge! This will break existing links, so may be worth searching the rest of the codebase for `build.graphile.org` links and updating them.

benjie:
I've used the solution outlined in https://github.com/facebook/docusaurus/issues/3930#issuecomment-819520796 to avoid the version dropdowns appearing until you're in the docs :+1:

PR #2326, opened 2025-01-18
Move pagination data requirements to runtime (not plantime)

benjie:
## Description

Related to:

- #2060 

Follow up to:

- #2325 

This PR removes `$step.eval*()` usage from pagination-related logic, moving `first`, `last`, `before`, `after`, and `offset` evaluation to runtime. (Future PRs will deal with `orderBy`, `condition`, etc.)

## Performance impact

Unknown; hopefully an improvement of worst-case performance.

## Security impact

Unknown; hopefully an improvement of worst-case performance.

## Checklist


benjie:
This seems to have added some empty SQL fetches (presumably for pagination info?) to stream; but since the results are still valid and we'll be replacing stream/defer with the updated version soon I don't see much point looking into it.

benjie:
The above issue exists elsewhere too; I've created its own issue:

- https://github.com/graphile/crystal/issues/2327

PR #2342, opened 2025-01-29
Remove TODO from make-add-pg-table-condition-plugin.md

jcgsville:
I tested it by creating a nearly identical plugin in my Postgraphile project. It works. TODO no longer necessary unless you wanted some more rigorous or automated testing or something


benjie (approved):
Awesome; thanks!

PR #2359, opened 2025-02-05
postgraphile website: Rename 4.x to 4 and remove /current/ from url

jemgillam:
## Description

Docs restructure, removes `current` from the URL 
Utilizes symlinks so we can have permalinking to the current docs


jemgillam:
@benjie This is ready to go 👍

benjie (approved):

PR #2375, opened 2025-02-19
Add CI check that links don't end in slash

XanderG2:
## Description

adds a test to check if internal links end in slashes, as slashes at the end of links break navigation.

## Performance impact

N/A

## Security impact

N/A

## Checklist


benjie:
@jemgillam this looks good to me, but as you can see from the check being added to CI quite a lot of our links are already failing with it. I think you have work in progress to fix these, let's get that merged :+1:

benjie (approved):
Excellent; thanks! 🙌

PR #2378, opened 2025-02-19
Vale github actions

XanderG2:
## Description

Vale github action. Gives warnings and errors for spelling and grammar. Only impacts files changed during pull request.

## Checklist


.github/workflows/vale.yml
```diff
@@ -0,0 +1,13 @@
...
+jobs:
+  vale:
+    name: runner / vale
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: errata-ai/vale-action@v2.1.1
+        with:
+          files: '["postgraphile/website", "grafast/website",
+            "graphile-build/website", "utils/website"]'
```
benjie, line 13: @jemgillam wanna look into giving the blogs different rules?

.github/workflows/vale.yml
```diff
@@ -0,0 +1,13 @@
+name: reviewdog
+on: [pull_request]
+
+jobs:
+  vale:
+    name: runner / vale
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: errata-ai/vale-action@v2.1.1
```
benjie:
```suggestion
      - uses: vale-cli/vale-action@d89dee975228ae261d22c15adcd03578634d429c # v2.1.1
```

utils/website/pg-sql2/index.md
```diff
@@ -69,5 +69,5 @@ This is a replacement for [@calebmer's pg-sql](https://www.npmjs.com/package/pg-
...
-- Adds `sql.literal` which is similar to `sql.value` but when used with simple values can write the valid direct to the SQL statement. **USE WITH CAUTION**. The purpose for this is if you are using _trusted_ values (e.g. for the keys to [`json_build_object(...)`](https://www.postgresql.org/docs/9.6/static/functions-json.html)) then debugging your SQL becomes a lot easier because fewer placeholders are used.
+- Uses a symbol-key on the query nodes to protect against an object accidentally being inserted verbatim and being treated as valid (because every Symbol is unique an attacker would need control of the code to get a reference to the Symbol in order to set it on an object (it cannot be serialised/deserialised via JSON or any other medium), and if the attacker has control of the code then you've already lost)
```
benjie:
```suggestion
- Uses a symbol-key on the query nodes to protect against an object accidentally being inserted verbatim and being treated as valid (because every Symbol is unique an attacker would need control of the code to get a reference to the Symbol in order to set it on an object (it can't be serialised/deserialised via JSON or any other medium), and if the attacker has control of the code then you've already lost)
```

utils/website/pg-sql2/index.md
```diff
@@ -69,5 +69,5 @@ This is a replacement for [@calebmer's pg-sql](https://www.npmjs.com/package/pg-
...
+- Adds `sql.literal` which is similar to `sql.value` but when used with simple values can write the valid direct to the SQL statement. **Use with caution**. The purpose for this is if you are using _trusted_ values (e.g. for the keys to [`json_build_object(...)`](https://www.postgresql.org/docs/9.6/static/functions-json.html)) then debugging your SQL becomes a lot easier because fewer placeholders are used.
```
benjie:
```suggestion
- Adds `sql.literal` which is similar to `sql.value` but when used with simple values can write the valid direct to the SQL statement. **Use with caution**. The purpose for this is if you are using _trusted_ values (for example the keys to [`json_build_object(...)`](https://www.postgresql.org/docs/9.6/static/functions-json.html)) then debugging your SQL becomes a lot easier because fewer placeholders are used.
```

.vale.ini
```diff
@@ -0,0 +1,19 @@
+StylesPath = .github/styles
+
+Vocab = Graphile
+
+MinAlertLevel = suggestion
+
+Packages = Microsoft, MDX
+
+[*.{md}]
```
benjie, line 9: We probably need to add mdx

benjie (changes requested):
Excellent; thanks!

benjie (approved):

PR #2379, opened 2025-02-19
Add CI check to ensure SQL is lowercase

XanderG2:
## Description

adds a ci check to ensure sql is in lowercase

## Performance impact

N/A

## Security impact

N/A

## Checklist


scripts/ci-check-to-ensure-sql-is-in-lowercase.mjs
```diff
@@ -0,0 +1,98 @@
+import fs from "fs";
+import path from "path";
+
+const regex = /```sql\n([\s\S]*?)\n```/g;
```
benjie, line 4: `m` enables "multiline" mode where `^` and `$` match beginning/end of line (rather than of string); we can then check that the line starts/ends with ```` ``` ````:

```suggestion
const regex = /^```sql\n([\s\S]*?)\n```$/gm;
```

scripts/ci-check-to-ensure-sql-is-in-lowercase.mjs
```diff
@@ -0,0 +1,98 @@
...
+  "HAVING",
+  "AS",
+  "VALUES",
+];
+
+const WEBSITE_FOLDERS = [
+  "../postgraphile/website",
+  "../grafast/website",
+  "../graphile-build/website",
+  "../utils/website",
```
benjie, line 24: We want this script to work no matter where you run it from, so let's make these absolute:

```suggestion
  `${__dirname}/../postgraphile/website`,
  `${__dirname}/../grafast/website`,
  `${__dirname}/../graphile-build/website`,
  `${__dirname}/../utils/website`,
```

scripts/ci-check-to-ensure-sql-is-in-lowercase.mjs
```diff
@@ -0,0 +1,98 @@
...
+  }
+  return result;
+}
+
+async function readFileAsync(filePath) {
+  try {
+    const data = await fs.promises.readFile(filePath, "utf-8");
+    return data;
+  } catch (error) {
+    console.error("Error reading file:", error);
```
benjie, line 47: This logs the error but continues anyway; would it be better to throw the error to ensure that any such issue is fixed?
```suggestion
    console.error(`Error reading file ${filePath}:`, error);
    throw error;
```

scripts/ci-check-to-ensure-sql-is-in-lowercase.mjs
```diff
@@ -0,0 +1,98 @@
...
+
+function wrongSQLCheck(segment) {
+  if (keywords.some((keyword) => segment.includes(keyword))) {
+    return true;
+  } else {
+    return false;
+  }
+}
+
+(async () => {
```
benjie, line 59: There's nothing wrong with this IIFE, but I prefer a more traditional layout:

```suggestion
async function main() {
```

scripts/ci-check-to-ensure-sql-is-in-lowercase.mjs
```diff
@@ -0,0 +1,98 @@
...
+
+  console.log(
+    Object.keys(wrongSQL),
+    `The above ${Object.keys(wrongSQL).length} files contain capital SQL.`,
+  );
+
+  if (wrongSQL.length !== 0) {
+    process.exitCode = 1;
+  }
+})();
```
benjie, line 98: continued from above

```suggestion
}

main().catch(e => {
  console.error(e);
  process.exit(1);
});
```

scripts/ci-check-to-ensure-sql-is-in-lowercase.mjs
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  console.log(
+    Object.keys(wrongSQL),
+    `The above ${Object.keys(wrongSQL).length} files contain capital SQL.`,
+  );
+
+  if (wrongSQL.length !== 0) {
+    process.exitCode = 1;
+  }
```
benjie, line 97: This just slightly improves the output:

```suggestion
  if (wrongSQL.length === 0) {
    console.log("All files pass checks");
  } else {
    console.log(
      `Found ${Object.keys(wrongSQL).length} files containing capitalized SQL; the Graphile style guide encourages the use of lower case SQL. Please lowercase the SQL in the following files:`,
    );
    console.dir(wrongSQL);
    process.exitCode = 1;
  }
```

benjie (changes requested):
This is great! The next step would be too hook it up to GitHub actions by creating a new workflow. It is likely that a very simple workflow should work since we don't even need to install any dependencies for this script. Great work!

benjie (dismissed):
Thanks; this will be a good way of maintaining consistency in the docs!

benjie (approved):

PR #2416, opened 2025-03-12
Expanded tests using variables

jemgillam:
## Description

Here's some tests added alongside the `order` test in dataplan-pg, using variables for the various arguments.


grafast/dataplan-pg/__tests__/queries/connections/order.variables.string.test.graphql
```diff
@@ -0,0 +1,34 @@
+## expect(queries).toHaveLength(2);
```
benjie, line 1:
```suggestion
## expect(errors).toBeFalsy();
## expect(queries).toHaveLength(2);
```

grafast/dataplan-pg/__tests__/queries/connections/order.variables.strings.test.graphql
```diff
@@ -0,0 +1,34 @@
+## expect(queries).toHaveLength(2);
```
benjie, line 1:
```suggestion
## expect(errors).toBeFalsy();
## expect(queries).toHaveLength(2);
```

benjie (changes requested):
Really good, but you missed `## expect(errors).toBeFalsy();` from a couple of the tests so it created `null` errors files. No problem, can fix from GitHub interface I think!

PR #2427, opened 2025-03-28
codecs: Add explicit array check

the-sun-will-rise-tomorrow:
## Description

Replaces an obscure error message such as "Cannot read properties of null (reading 'length')" with an explicit one.

## Performance impact

Should be negligible.

## Security impact

Should be none.

## Checklist

### DIFF
--- a/grafast/dataplan-pg/src/codecs.ts
+++ b/grafast/dataplan-pg/src/codecs.ts
@@ -656,6 +656,11 @@ export function listOfCodec<
         .flat(100)
         .map((v) => (v == null ? null : innerCodec.fromPg(v))) as any,
     toPg: (value) => {
+      if (!Array.isArray(value)) {
+        throw new Error(
+          `Do not know how to encode ${inspect(value)} to an array`,
+        );
+      }
       let result = "{";
       for (let i = 0, l = value.length; i < l; i++) {
         if (i > 0) {


benjie:
Thanks for the pull request! The `Array.isArray` check is not cheap when in hot loops, so I'd like to understand better the motivation for this.

> Replaces an obscure error message such as "Cannot read properties of null (reading 'length')" with an explicit one.

This is a bit of a strange issue, IIRC the codec's `toPg` shouldn't be called on `null` (since `null` should be handled in a common way for all codecs) - do you know how you were able to produce this error?

the-sun-will-rise-tomorrow:
Ah OK, perhaps this issue is entirely on my side.

I have a custom plugin which implements an upsert-like action. The code is based on the existing upsert plugin. It looks like this:

```ts
for (let attrName in resource.codec.attributes) {
  const attr = resource.codec.attributes[attrName];
  const gqlName = inflection.camelCase(attrName);
  if (patch[gqlName] === undefined) continue;
  sqlColumns.push(sql.identifier(attrName));
  sqlValues.push(sql.value(attr.codec.toPg(patch[gqlName])));
  updateArray.push(
    sql.fragment`${sql.identifier(attrName)} = ${sql.value(attr.codec.toPg(patch[gqlName]))}`,
  );
}
// ...
if (doInsert) {
  sqlQuery = sql.query`
      INSERT INTO ${resource.codec.sqlType}
      (${sql.join(sqlColumns, ", ")})
      VALUES (${sql.join(sqlValues, ", ")})
      RETURNING *
    `;
} else {
  sqlQuery = sql.query`
      UPDATE ${resource.codec.sqlType}
      SET ${sql.join(updateArray, ", ")}
      WHERE id = ${sql.value(id)} AND version_id = ${sql.value(draftId)}
      RETURNING *
    `;
}
```

In any case, feel free to close this if it's not useful.

benjie:
Ah thanks for sharing! Rather than

```ts
sql.value(attr.codec.toPg(patch[gqlName]))
```

do:

```ts
sqlValueWithCodec(patch[gqlName], attr.codec)
```

`sqlValueWithCodec` is new in this latest release and I've not got around to documenting it yet. (Doc PRs always welcome!)

PR #2506, opened 2025-05-24
Fix unnecessary duplicate calls to same plan resolver during polymorphic planning

benjie:
We were calling the `plan()` resolver function for each individual polymorphic path, but this isn't necessary. This change batches these calls so that we only need to call it again if something differs (for example, the arguments). It's based on the operation path, so fields with different aliases (e.g. multiple of the same mutation) will still be called for each alias.


benjie:
Split this up into:

- #2507 
- #2508 
- #2509  
- #2510

PR #2549, opened 2025-06-04
Beginning of codegen plugin

benjie:
This has only been tested with `Step<...>` and may not work with other step classes. This only works with hand-written schemas. This only works if you supply your own type configuration file containing a list of all the types and what they expect in terms of steps.

But it's a start. We use it to type [the DCC test schema](https://github.com/graphile/crystal/blob/main/grafast/grafast/__tests__/dcc/dcc-schema.ts)


benjie:
**DO NOT MERGE** until next release is out.

PR #2560, opened 2025-06-08
docs: add type inference with TypedDocument for variableValues

Sleepful:
## Description

Just some docs, I am not a fan of `as any` but that's what it takes for now, and allows `variableValues` to be typed-checked at `executeDocument` callsite

### DIFF
--- a/postgraphile/website/postgraphile/usage-schema.md
+++ b/postgraphile/website/postgraphile/usage-schema.md
@@ -212,7 +212,7 @@ automatically reflect the correct types thanks to graphql-codegen:
 
 ```ts
 import type { DocumentNode, ExecutionResult } from "postgraphile/graphql";
-import type { TypedDocumentNode } from "@graphql-typed-document-node/core";
+import type { TypedDocumentNode, VariablesOf } from "@graphql-typed-document-node/core";
 import { postgraphile } from "postgraphile";
 import { execute, hookArgs } from "postgraphile/grafast";
 import { validate } from "postgraphile/graphql";
@@ -223,7 +223,7 @@ const pgl = postgraphile(preset);
 export async function executeDocument<TData = any, TVariables = any>(
   requestContext: Partial<Grafast.RequestContext>,
   document: DocumentNode | TypedDocumentNode<TData, TVariables>,
-  variableValues?: Record<string, unknown> | null,
+  variableValues?: VariablesOf<TypedDocumentNode<TData, TVariables>> | null,
   operationName?: string,
 ): Promise<ExecutionResult<TData, TVariables>> {
   const { schema, resolvedPreset } = await pgl.getSchemaResult();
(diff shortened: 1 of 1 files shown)


Sleepful:
Added another commit, now it properly check for nested properties in the `variableValues` and reports an error if there are any additional properties that should not be there!

Example:

The query, just for completeness:
```ts
  const q = graphql(`
    mutation Patch($rowId: Int!, $sliderPatch: sliderPatch!) {
      updateSliderByRowId(
        input: {sliderPatch: $sliderPatch, rowId: $rowId}
      ) {
        clientMutationId
        slider {
          enDescription
          esDescription
          id
          image
          rowId
          url
        }
      }
    }
`)
```

Here is where it gets better reporting, ignoring the first argument of `requestContext` for this example:

```ts
  const res = await executeDocument({}, q, {
    sliderPatch:
      { enDescription, esDescription, image, url, a: "" },
                                    // highlights ^ `a` and reports error!
      rowId
    })
```

Error reported:

```ts
Object literal may only specify known properties, and 'a' does not exist in type '{ url?: string | null | undefined; esDescription?: string | null | undefined; enDescription?: string | null | undefined; image?: string | null | undefined; }'
```

postgraphile/website/postgraphile/usage-schema.md
```diff
@@ -212,20 +212,20 @@ automatically reflect the correct types thanks to graphql-codegen:
...
-export async function executeDocument<TData = any, TVariables = any>(
+export async function executeDocument<TDoc extends TypedDocumentNode<TData, TVariables>, TData = any, TVariables = any>(
   requestContext: Partial<Grafast.RequestContext>,
-  document: DocumentNode | TypedDocumentNode<TData, TVariables>,
-  variableValues?: Record<string, unknown> | null,
+  document: TDoc,
+  variableValues?: VariablesOf<typeof document>,
   operationName?: string,
-): Promise<ExecutionResult<TData, TVariables>> {
+): Promise<ExecutionResult<ResultOf<typeof document>, TVariables>> {
```
benjie: I don't think we need generics for TData/TVariables, maybe something like this would work:

```ts
export async function executeDocument<TDoc extends TypedDocumentNode | DocumentNode, TExtensions = ObjMap<unknown>>(
  requestContext: Partial<Grafast.RequestContext>,
  document: TDoc,
  variableValues?: VariablesOf<TDoc>,
  operationName?: string,
): Promise<ExecutionResult<ResultOf<TDoc>, TExtensions>> {
```

(Also the second parameter to ExecutionResult is TExtensions not TVariables - I recognize this was already incorrect.)

benjie:
Hi @Sleepful, gentle nudge to see if you're interested in the above change?

Sleepful:
Thanks for taking care of the diff and merging it 💙

PR #2561, opened 2025-06-09
website fix: fix links to graphile.org

jemgillam:
## Description

We changed the slug of the page on graphile.org but never updated the link from the other webpages!


jemgillam:
Ooops, old filenames

PR #2571, opened 2025-06-13
Add JSONPath scalar type and codec support

slaskis:
## Description

I have been patching postgraphile to add support for [`jsonpath`](https://www.postgresql.org/docs/current/datatype-json.html#DATATYPE-JSONPATH) and thought it might make sense to add upstream.

## Performance impact

unknown, negligible I would presume since it just adds a codec

## Checklist


grafast/dataplan-pg/src/codecs.ts
```diff
@@ -1042,6 +1042,7 @@ export const TYPES = {
     fromPg: jsonParse,
     toPg: jsonStringify,
   }),
+  jsonpath: t()("4072", "jsonpath", verbatim),
```
benjie: We can't safely treat jsonpath as text (unlike `citext`) because it doesn't have the relevant casts:

```
$ SELECT '$.foo.bar'::jsonpath = '$.foo.bar';
ERROR:  42883: operator does not exist: jsonpath = unknown
```

```suggestion
  jsonpath: t()("4072", "jsonpath"),
```

benjie (dismissed):
Looks good to me! Please:

1. Add a column to the b.types table: https://github.com/graphile/crystal/blob/main/postgraphile/postgraphile/__tests__/kitchen-sink-schema.sql#L275
2. Insert null into the column value for the first b.types row: https://github.com/graphile/crystal/blob/main/postgraphile/postgraphile/__tests__/kitchen-sink-data.sql#L158
3. Insert a sensible expression into the second b.types row: https://github.com/graphile/crystal/blob/main/postgraphile/postgraphile/__tests__/kitchen-sink-data.sql#L208
4. Add a column to the types mutation update and create operations: https://github.com/graphile/crystal/blob/dc0d44696d929f2ffcb8725e5bcf9951db937cb1/postgraphile/postgraphile/__tests__/mutations/v4/types.test.graphql#L81 and https://github.com/graphile/crystal/blob/dc0d44696d929f2ffcb8725e5bcf9951db937cb1/postgraphile/postgraphile/__tests__/mutations/v4/types.test.graphql#L138
5. Ensure that the value is queried in the response to the mutation and to the types query: https://github.com/graphile/crystal/blob/dc0d44696d929f2ffcb8725e5bcf9951db937cb1/postgraphile/postgraphile/__tests__/mutations/v4/types.test.graphql#L292 and https://github.com/graphile/crystal/blob/main/postgraphile/postgraphile/__tests__/queries/v4/types.test.graphql#L197
6. Run the full test suite with envvar `UPDATE_SNAPSHOTS=1`

Thanks!

slaskis:
Great review thanks! 🙏 i was unable to get it to build and test (unrelated to this PR) but i’ll have another go at it later this week.

slaskis:
All right, I managed to get the build and tests to run in an Ubuntu VM so it's looking a lot better now!

A few whitespace changes snuck in, if you want me to revert those let me know!

benjie (commented):
Looks good!

Might make sense for us to implement default behaviors so that orderBy/condition on jsonPath is disabled by default. You can leave that with me if you want because I can't remember how to do it off-hand.

benjie (dismissed):

PR #2586, opened 2025-06-19
Update acl.ts

ttfkam:
## Description

* Improve validation on ACL string: prevent "**", "*m"
* Grants can skip string search if the base permission is missing.

No new tests. Should only require regression testing.

## Performance impact

No appreciable speed regression on regex. Very minor speed improvement for ACL parsing logic in aggregate.

## Security impact

Excludes more invalid ACL input.

## Checklist


benjie:
Out of interest, what’s the motivation for this?

If you want to make it faster, using indexOf instead of includes would allow you to then check the next character trivially to see if it’s a * without needing another string search. Better still would be a single loop over the string and no regexps at all.

ttfkam:
Superseded by #2587 per your comments. Motivation: browsing through the code and saw a simple opportunity.

PR #2587, opened 2025-06-20
Strict ACL parsing without a regex

ttfkam:
## Description

Supersedes #2586 per maintainer comments.

Motivation: browsing through the code and saw a simple opportunity.

* Improve validation on ACL string: prevent "**", "*m", "z", "%", etc.

No new tests. Should only require regression testing.

## Performance impact

Minor speed improvement for ACL parsing logic in aggregate due to one less regex.

## Security impact

Excludes most invalid ACL input. (Doesn't stop duplicate ACL tokens.)

## Checklist


utils/pg-introspection/src/acl.ts
```diff
@@ -134,80 +134,67 @@ export type ResolvedPermissions = Omit<AclObject, "role" | "granter">;
...
  * Accepts an ACL string such as `foo=arwdDxt/bar` and converts it into
  * a parsed AclObject.
  */
 export function parseAcl(aclString: string): AclObject {
-  // https://www.postgresql.org/docs/current/ddl-priv.html#PRIVILEGE-ABBREVS-TABLE
-
-  const matches = aclString.match(/^([^=]*)=([rwadDxtXUCcTm*]*)\/([^=]+)$/);
-
-  if (!matches) {
+  const acl = Object.assign({}, NO_PERMISSIONS);
```
benjie:
```suggestion
  const acl: AclObject = { ...NO_PERMISSIONS };
```

utils/pg-introspection/src/acl.ts
```diff
@@ -134,80 +134,67 @@ export type ResolvedPermissions = Omit<AclObject, "role" | "granter">;
...
+  D: 'truncate',
+  x: 'references',
+  t: 'trigger',
+  X: 'execute',
+  U: 'usage',
+  C: 'create',
+  c: 'connect',
+  T: 'temporary',
+  m: 'maintain',
+};
```
benjie:
```suggestion
} as const;
```

utils/pg-introspection/src/acl.ts
```diff
@@ -134,80 +134,67 @@ export type ResolvedPermissions = Omit<AclObject, "role" | "granter">;
...
+  const acl = Object.assign({}, NO_PERMISSIONS);
+  const roleEndIndex = aclString.indexOf("=");
+  if (roleEndIndex === -1) {
     throw new Error(`Could not parse ACL string '${aclString}'`);
+  } else if (roleEndIndex > 0) {
+    acl.role = parseIdentifier(aclString.substring(0, roleEndIndex));
+  }
+  const aclLength = aclString.length;
+  const lastGrantableTokenIndex = aclLength - 1;
+  let currentPerm: string;
```
benjie: No need for this to be in a higher scope? In general I avoid `let` where possible.
```suggestion
```

utils/pg-introspection/src/acl.ts
```diff
@@ -134,80 +134,67 @@ export type ResolvedPermissions = Omit<AclObject, "role" | "granter">;
...
+  let i = roleEndIndex + 1; // skip past the "="
+  // Process the ACL tokens
+  while (i < aclString.length) {
+    const nextChar = aclString[i];
+    if (nextChar === "/") {
+      // granter begins
+      i++;   // skip past the "/" delimiter
+      break; // we're done with ACL tokens
+    }
+    currentPerm = ACL_MAP[nextChar];
```
benjie:
```suggestion
    const currentPerm = ACL_MAP[nextChar as keyof typeof ACL_MAP];
```

utils/pg-introspection/src/acl.ts
```diff
@@ -134,80 +134,67 @@ export type ResolvedPermissions = Omit<AclObject, "role" | "granter">;
...
+    if (currentPerm === undefined) {
+      throw new Error(`Could not parse ACL string '${aclString}'`);
+    }
+    acl[currentPerm] = true;
+    if (i < lastGrantableTokenIndex && aclString[i + 1] === '*') {
+      // permission + grant
+      i++; // skip past the "*" character
+      acl[currentPerm + "Grant"] = true;
```
benjie: TypeScript will recognize this as a constant string, which is one of the allowed keys on AclObject.

```suggestion
      acl[`${currentPerm}Grant`] = true;
```

utils/pg-introspection/src/acl.ts
```diff
@@ -134,80 +134,67 @@ export type ResolvedPermissions = Omit<AclObject, "role" | "granter">;
...
+    }
+    i++;
+  } // end token processing
+  if (i < aclLength) {
```
benjie: Previously the granter was required, and in general I think it will always be set. Let's just throw an error if it's the empty string.

benjie (changes requested):
This definitely looks more performant - a single loop over the string rather than at least 26! However, it's not as clearly correct as it was - the previous code was very obvious and this code is less so - so it will need some unit tests. Just simple assertions that a number of standard permissions strings with/without role, with/without grants parse correctly. Since pg-introspection doesn't seem to have tests already, please use `node:test` for them (or mocha/chai if you prefer).

In general we trust permission strings to be valid (we trust Postgres) so I'm not too concerned about malicious permission strings, _however_ Postgres does add new permissions from time to time so it's worth ensuring that the code correctly handles (via an error) a new permission character.

I've included a number of TypeScript improvements, all minor.

utils/pg-introspection/src/acl.ts
```diff
@@ -134,80 +134,66 @@ export type ResolvedPermissions = Omit<AclObject, "role" | "granter">;
...
+  t: 'trigger',
+  X: 'execute',
+  U: 'usage',
+  C: 'create',
+  c: 'connect',
+  T: 'temporary',
+  m: 'maintain',
+} as const;
+const NO_PERMISSIONS: AclObject = Object.values(ACL_MAP).reduce((acc, val) => {
+  acc[val] = acc[val + "Grant"] = false;
```
benjie:
```suggestion
  acc[val] = acc[`${val}Grant`] = false;
```

utils/pg-introspection/src/acl.ts
```diff
@@ -134,80 +134,66 @@ export type ResolvedPermissions = Omit<AclObject, "role" | "granter">;
...
+  U: 'usage',
+  C: 'create',
+  c: 'connect',
+  T: 'temporary',
+  m: 'maintain',
+} as const;
+const NO_PERMISSIONS: AclObject = Object.values(ACL_MAP).reduce((acc, val) => {
+  acc[val] = acc[`${val}Grant`] = false;
+  return acc;
+}, { role: "public", granter: "" });
```
benjie:
```suggestion
}, { role: "public", granter: "" } as Partial<AclObject>) as AclObject;
```

ttfkam:
Hey, sorry for dropping the ball here. I got caught up at work. Thank you for taking this past the finish line.

PR #2610, opened 2025-06-29
Unref the pgServices cache timer to prevent blocking process exit

### LINKED ISSUE #2609: getWithPgClientFromPgService prevents quick exit
### Summary

I'm integrating postgraphile with nuxt, and as part of that I would like it so I can generate the schema and types during a nuxt prepare step. I've gotten this to work pretty well, but it seems to hang for a long time after completing the schema export - even if `makeSchema` has completed.

### Steps to reproduce

I narrowed down the issue to not be related to schema export but rather in the gather phase. So if you run the following code with any database that has a "app" schema:
```js
async function main() {
  console.log('Starting Graphile build...')
  const input = await gather({
    extends: [PostGraphileAmberPreset],
    pgServices: [
      makePgService({
        connectionString: process.env.DATABASE_URL,
        schemas: ['app'],
      }),
    ],
  })
  console.log('Gathered Graphile build input:', input)
}

await main()
```

### Expected results

The code should print "Starting Graphile build..." followed by "Gathered Graphile build input" and then exit very shortly after.

### Actual results

The program will run for at least 10 seconds no matter what.
If we add `poolConfig: { allowExitOnIdle: true }` to the pgService config, then it gets better, but the program will still idle for at least 5 seconds before exiting.

### Additional context
```
node: v24.1.0
@dataplan/pg: 0.0.1-beta.32

### Possible Solution

<!--- Optionally suggest a fix or reason for the bug. -->

jsmnbom:
## Description

Fixes #2609 

## Performance impact

unknown - should be none
possibly positive in case user code anywhere is waiting for process exit?

## Security impact

unknown - should be none.

## Checklist

I have no idea how/if this would be tested - sorry, and it's such a tiny change so dunno about the others.

### DIFF
--- a/grafast/dataplan-pg/src/pgServices.ts
+++ b/grafast/dataplan-pg/src/pgServices.ts
@@ -84,7 +84,7 @@ export function getWithPgClientFromPgService<
         cachedValue.retainers--;
 
         // To allow for other promises to resolve and add/remove from the retaininers, check after a tick
-        setTimeout(
+        const tid = setTimeout(
           () => {
             if (cachedValue.retainers === 0 && !released) {
               released = true;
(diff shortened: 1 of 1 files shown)


benjie (approved):
Thanks!

PR #2622, opened 2025-07-11
Upgrade GraphiQL, fix PostGraphile Ruru, grafserv static support

benjie:
- Removes our patch to `graphiql` which is no longer needed
- Adds grafserv support for serving Ruru's static assets
- Fixes rendering of Ruru from PostGraphile
- Adds support in grafserv for "raw" responses allowing any content-type header


grafast/grafserv/src/options.ts
```diff
@@ -95,6 +95,7 @@ export function optionsFromConfig(config: GraphileConfig.ResolvedPreset) {
     graphiql = true,
     graphiqlOnGraphQLGET = true,
     graphiqlPath = "/",
+    graphiqlStaticPath = `/ruru-static/`,
```
benjie:
```suggestion
    graphiqlStaticPath = "/ruru-static/",
```

PR #2624, opened 2025-07-11
ruru: make the loading state match the expected sizes better

benjie:
Make it less "flicky" for @jemgillam


jemgillam:
@benjie Um Actually, it's for the Greater Good. _the Greater Good._

PR #2626, opened 2025-07-11
Upgrade more dependencies

benjie:


benjie:
We can't do this until we either ditch Jest or somehow convince it to support `require(esm)` inside of VM modules.

PR #2639, opened 2025-07-15
Custom node id codecs

### LINKED ISSUE #2332: Ability  to replace `nodeId` creation logic for specific or all types
### Feature description

It would be very helpful to be able to replace the logic how `nodeId` is generated for tables or any other autogenerated resources. This will allow us to encrypt or mask restricted content in the nodeId. Which will be especially useful for the projects that are migrating to postgraphile.

### Motivating example

Example: 

When a `resource.primary_key` is a restricted information, we want to return to the user the primary key, but encrypted, so only the server can decrypt the value, which will allow us to mask the data but the server will be able to understand what have been masked. In a system where there is already a large code designed around the idea that `resource.primary_key` is a restricted information, the only thing that is missing is to also hide it in the `nodeId`. It's possible to introduce a new resource with a primary key which is not related to the original resource and also add a relation between those 2 resources that only the server can access, but it can be easily avoided by allowing us to change the logic of generating and reading the `nodeId`.

### Breaking changes

I'm not await what this will break, but I guess that internally, postgraphile is using the content of the `nodeId`, so if we are adding an option to replace the logic, maybe we will also need an API which will return the data postgraphile needs? Like

```ts
{
  write: (resource) => `any string that the "read" function below understands`,
  read: (nodeId) => ({original: [Type, id], otherExtractedData: 'data' }),
  // The "original" field is what postgraphile needs
}
```

### Supporting development

I [tick all that apply]:

- [x] am interested in building this feature myself
- [ ] am interested in collaborating on building this feature
- [x] am willing to help testing this feature before it's released
- [ ] am willing to write a test-driven test suite for this feature (before it exists)
- [x] am a [Graphile sponsor](https://www.graphile.org/sponsor/) ❤️
- [ ] have an active [support or consultancy contract](https://www.graphile.org/support/) with Graphile

jsmnbom:
## Description
Fixes #2332

## Performance impact

None, unless the user custom a custom nodeIdCodec that does heavy work.

## Security impact

Unknown

## Checklist

https://github.com/graphile/crystal/blob/main/CONTRIBUTING.md doesn't explain how to add RELEASE_NOTES and such, so haven't done that.

I've added tests mostly for defaultNodeIdCodec, as I'm not sure how tests for smart comments that do not change the schema output would look or where they would go.

### DIFF
--- a/postgraphile/postgraphile/__tests__/queries/relay/defaultNodeIdCodec.Query.node.json5
+++ b/postgraphile/postgraphile/__tests__/queries/relay/defaultNodeIdCodec.Query.node.json5
@@ -1,0 +1,7 @@
+{
+  alice: {
+    id: "people|3",
+    firstName: "Alice",
+    lastName: "Alicesson",
+  },
+}
--- a/postgraphile/postgraphile/__tests__/queries/relay/defaultNodeIdCodec.Query.node.mermaid
+++ b/postgraphile/postgraphile/__tests__/queries/relay/defaultNodeIdCodec.Query.node.mermaid
@@ -1,0 +1,153 @@
+%%{init: {'themeVariables': { 'fontSize': '12px'}}}%%
+graph TD
+    classDef path fill:#eee,stroke:#000,color:#000
+    classDef plan fill:#fff,stroke-width:1px,color:#000
+    classDef itemplan fill:#fff,stroke-width:2px,color:#000
+    classDef unbatchedplan fill:#dff,stroke-width:1px,color:#000
+    classDef sideeffectplan fill:#fcc,stroke-width:2px,color:#000
+    classDef bucket fill:#f6f6f6,color:#000,stroke-width:2px,text-align:left
+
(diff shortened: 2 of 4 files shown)


graphile-build/graphile-build/src/plugins/NodePlugin.ts
```diff
@@ -31,6 +31,14 @@ declare global {
     interface ScopeObjectFieldsField {
       isRootNodeField?: boolean;
     }
+
+    interface SchemaOptions {
+      /**
+       * The default Node ID codec to use when not overridden by a smart tag.
+       * Use build.registerNodeIdCodec to register your own codecs.
```
benjie:
```suggestion
       * The default Node ID codec to use. May be overridden on a
       * table-by-table basis using the `@nodeIdCodec` smart tag.
       *
       * Use build.registerNodeIdCodec to register your own codecs.
       *
       * @defaultValue `"base64JSON"`
```

benjie (approved):

PR #2659, opened 2025-08-05
Overhaul connections, error handling, and loadOne/loadMany

benjie:
:sweat_smile:


jemgillam:
I am getting an interesting error on writing some tests 

```
LoadedRecordStep could not find the parent LoadManyStep; instead found Access{4}<27.item>[28]
```

PR #2660, opened 2025-08-05
connections: grafast connection tests

jemgillam:
## Description

Some tests for
#2659 

- Better error given for connections when using eg first & before


grafast/grafast/__tests__/dcc/dcc-schema.ts
```diff
@@ -394,6 +432,10 @@ export const makeBaseArgs = () => {
             const $items = get($crawler, "items");
             return lambda([$items, $first], applyLimit);
           },
+          itemsConnection($crawler, fieldArgs) {
+            const $items = get($crawler, "items");
+            return connection($items, { fieldArgs });
+          },
```
benjie, line 438: We need a plan resolver on the other types too

benjie (approved):

PR #2675, opened 2025-08-20
Fix `relatedNode=null` bug after polymorphism overhaul

slaskis:
## Description

As [discussed in Discord](https://discord.com/channels/489127045289476126/498852330754801666/1407381647254356018) I have an issue with subscriptions after upgrading from v5 beta 33 to beta 46 where they no longer include the relatedNode in the subscription output. 

This PR adds a test which showcase the issue based on a [simplified "ouch-my-finger" example](https://github.com/slaskis/ouch-my-finger/pull/1).

## Performance impact

unknown

## Security impact

unknown

## Checklist


postgraphile/postgraphile/__tests__/queries/v4/sub.json5
```diff
@@ -0,0 +1,20 @@
+[
+  {
+    data: {
+      listen: {
+        __typename: "ListenPayload",
+        relatedNodeId: "WyJYIiwxXQ==",
+        relatedNode: null,
```
slaskis: I believe these show the current issue, not sure how to write an expect for this in the test yet...

postgraphile/postgraphile/__tests__/queries/v4/sub.test.graphql
```diff
@@ -0,0 +1,22 @@
+#> schema: ["sub"]
+#> subscriptions: true
+#> simpleSubscriptions: true
+#> directPg: true
+#> checkErrorSnapshots: false
+#! const query = 'select sub.insert_x($1)';
+#! await pgClient.query(query, ["a"]);
+#! await waitFor(() => payloads.length >= 1, 15000);
+#! await pgClient.query(query, ["b"]);
+#! await waitFor(() => payloads.length >= 2, 15000);
```
benjie: I think you want an assertion like this; have a look at the incremental (stream/defer) tests to see what the format should be.

```suggestion
#! await waitFor(() => payloads.length >= 2, 15000);
## expect(payloads[0].listen.relatedNode).toBeTruthy()
```

postgraphile/postgraphile/__tests__/queries/v4/sub.test.graphql
```diff
@@ -0,0 +1,22 @@
+#> schema: ["sub"]
+#> subscriptions: true
+#> simpleSubscriptions: true
```
benjie: Please create a mutation test with the same configuration and perform:

```graphql
mutation {
  insertX(name: "Foo") {
    x {
      nodeId
      id
      name
    }
  }
}
```

This will confirm the nodeId that we're aiming for - they need to match.

postgraphile/postgraphile/__tests__/helpers.ts
```diff
@@ -575,7 +575,7 @@ export async function runTestQuery(
               payloads,
               errors,
               queries,
-              extensions: payloads[0].extensions,
+              extensions: payloads[0]?.extensions,
```
benjie, line 584: Ah; for a subscription with no events there'd be no payloads - got it. So you should also mod the block on line 569 too - we don't want to try and re-sort these. Something like:

```ts
const payloads =
  operationType === "subscription"
    ? payloads
    : ([
        originalPayloads[0],
        ...originalPayloads.slice(1).sort(sortPayloads),
      ] as AsyncExecutionResult[]);
```

postgraphile/postgraphile/__tests__/queries/v4/sub.test.graphql
slaskis: nice, that's much better than a pr comment :)

postgraphile/postgraphile/__tests__/queries/v4/sub.test.graphql
slaskis: added!

PR #2705, opened 2025-09-08
Don't send empty invalidation notifications in postgraphile_watch

zarybnicky:
Fixes an issue where the schema was rebuilding on every drop of pg_temp, namely after the end of every session with REFRESH MATERIALIZED VIEW CONCURRENTLY. That resulted in not one but two invalidation notification: a `type: ddl` with an empty payload, and a `type: drop` with payload of `["pg_temp"]`.

`pg_temp` is a special per-session schema which gets used in `CONCURRENTLY` commands and it is where `TEMPORARY` objects get created, and I'm confident these changes shouldn't invalidate the schema.

I'm less sure about not sending notifications in the case where `pg_event_trigger_ddl_commands()` returns zero rows. Looking into its behavior, it seems to me it should return zero rows only in the case of internal housekeeping operations (like this operation which is an effective `DROP SCHEMA pg_temp`), and I haven't found any such situation where postgraphile was actually supposed to invalidate.

(Should I add a changeset or change anything else? I looked for some guidance but likely in the wrong place :) )

### DIFF
--- a/graphile-build/graphile-build-pg/src/watchFixtures.ts
+++ b/graphile-build/graphile-build-pg/src/watchFixtures.ts
@@ -21,30 +21,37 @@ drop schema if exists postgraphile_watch cascade;
 create schema postgraphile_watch;
 
 create function postgraphile_watch.notify_watchers_ddl() returns event_trigger as $$
+declare
+  ddl_commands json;
 begin
-  perform pg_notify(
-    'postgraphile_watch',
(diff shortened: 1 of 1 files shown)


benjie:
Closing in favour of #2709

(Note: in future you should contribute via a branch on your fork, that way the maintainer can push fixes to your branch.)

PR #2711, opened 2025-09-14
Add graphile-build-pg pluins to its index export

JannisKoeksel:
## Description
The graphile-build-pg preset uses four plugins not exportet in graphile-build-pg index. I just added them.

## Performance impact

None

## Security impact

None

#


benjie (dismissed):
Thanks!

PR #2740, opened 2025-09-30
Docs for pg-sql2

benjie:
- Fixes #2312

The first round of this was AI-generated, and it made a complete mess.

I had it tidy up... That was a mess too.

I've now gone through and done my own editorial.

And @jemgillam is finally cleaning up the remainder.

PR #2746, opened 2025-10-01
website: pg-sql2 examples in docs

jemgillam:
## Description

Continuing work from 

- #2740 

Making sure the examples are consistent with each other, are easily copyable and include the required import lines. 

Fixes:
- #2754

### DIFF
--- a/grafast/dataplan-pg/src/steps/pgSelect.ts
+++ b/grafast/dataplan-pg/src/steps/pgSelect.ts
@@ -3843,7 +3843,10 @@ function buildQueryParts<TResource extends PgResource<any, any, any, any, any>>(
   const extraSelectIndexes = extraSelects.map((_, i) => i + l);
 
   return {
-    comment: sql.comment(`From ${info.sourceStepDescription}`, false),
+    comment: sql.comment(
+      `From ${info.sourceStepDescription}`,
+      process.env.DATAPLAN_PG_DEBUG === "1",
+    ),
     selects,
     from,
     joins: info.joins,


utils/website/pg-sql2/api/sql.md
```diff
@@ -39,19 +39,33 @@ Returns a `SQL` fragment that can be:
...
+/*
+  SELECT *
+  FROM users
+  WHERE id = 123
+*/
 ```
 
 ### Embedding identifiers and values
 
+This example uses [`sql.identifier`](./sql-identifier.md) and [`sql.value`](.sql-value),
```
benjie:
```suggestion
This example uses [`sql.identifier`](./sql-identifier.md) and [`sql.value`](./sql-value.md),
```

benjie (changes requested):
If you have a code block like:

```ts
const query = sql`...1...`;
// ...
const query = sql`...2...`;
```

the code block won't run as a whole, because you're not allowed to declare the same variable twice. There are three potential solutions to this:

1. Wrap each section in block:
   ```ts
   {
     const query = sql`...`;
     console.log(sql.compile(query));
   }
   {
     const query = sql`...`;
     console.log(sql.compile(query));
   }
   ```
   Con: many developers don't know this is legit, it will confuse them.
2. Use different variable names for each `query`.  
   Con: very easy to confuse the variables or forget to update them somewhere.
3. Make them separate code blocks
   Con: longer documentation page (more boilerplate)

Not sure which is best.

jemgillam:
I was veering towards using different query names but wasn't sure if `query1` `query2` etc would be annoying to read

I was wondering if they just copy & paste as it is they could fix it themselves, but that also isn't very helpful on our part. I'll see if renaming makes sense, or perhaps sometimes it's a rename and sometimes it's a separate code block.

jemgillam:
I've gone for a kind of hybrid. 

When an example has multiple `query` in the same code block, those have now been split up without the `import` line included, and often with a comment at the top showing the difference between it and the previous example. But when the examples are broken up by headings I've kept the `import` line.

utils/website/pg-sql2/api/sql-literal.md
```diff
@@ -31,17 +31,28 @@ Returns a `SQL` fragment with the value embedded directly in the SQL text.
...
-sql`LIMIT ${sql.literal(50)}`; // -> LIMIT 50
-sql`WHERE active = ${sql.literal(true)}`; // -> WHERE active = TRUE
+const query1 = sql`LIMIT ${sql.literal(50)}`;
+console.log(sql.compile(query1).text);
+// LIMIT 50
+
+const query2 = sql`WHERE active = ${sql.literal(true)}`;
+console.log(sql.compile(query2).text);
+// WHERE active = TRUE
+```
```
benjie, line 46:
````suggestion
```

```js
const query = sql`WHERE active = ${sql.literal(true)}`;
console.log(sql.compile(query).text);
// WHERE active = TRUE
```
````

File: utils/website/pg-sql2/api/sql.md [hunk 3]
```diff
@@ -62,20 +76,40 @@ const userQuery = sql`
...
   WHERE ${sql.identifier(tableName, columnName)} = ${sql.value(columnValue)}
 `;
+
+const { text, values } = sql.compile(userQuery);
+console.log(text, values);
+/*
+  SELECT *
+  FROM "users"
+  WHERE "users"."id" = $1
+*/
```
benjie, line 90: The code block here has the text but not the values.

benjie (approved):
Looks great; thanks!

PR #2777, opened 2025-10-15
Implement queue for watchGather/watchSchema/etc

benjie:
Need to test this before merging.

### DIFF
--- a/.changeset/ninety-hairs-wave.md
+++ b/.changeset/ninety-hairs-wave.md
@@ -1,0 +1,9 @@
+---
+"graphile-build": patch
+"postgraphile": patch
+"grafserv": patch
+---
+
+Implement queue for `watchGather`/`watchSchema`/etc and integrate with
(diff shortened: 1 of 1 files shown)


benjie:
Tested via:

```sh
(for I in `seq 1 1000`; do echo 'create table my_table(id serial primary key, name text); drop table my_table;'; done) | psql watchtest
```

With 

```ts
let schemaCounter = 0;
const DelaySetPresetPlugin = {
  name: "DelaySetPresetPlugin",
  grafserv: {
    middleware: {
      setPreset(next, event) {
        return sleep(1000).then(next);
      },
    },
  },

  schema: {
    hooks: {
      finalize(schema, build) {
        console.log(`Produced ${++schemaCounter}th schema`);
        return schema;
      },
    },
  },
};
```

Running PostGraphile hooked up to this schema created 24 schemas total, even though the above should have triggered 2000 schema build notifications. The 1 second delay was clearly respected.

PR #2792, opened 2025-10-27
Handle bufferStream in HonoGrafserv

malyzeli:
Implemented event streaming response in `hono` adapter to enable watch mode support.

Basically it's handler from `koa` adapter (`grafast/grafserv/src/servers/koa/v2/index.ts`) patched to use `hono` API, absolutely no innovation there.

Using my version locally and haven't see any problems so far, linter doesn't report any problems in changed file.

PS: This is my first contribution to open-source ever, so I'm not sure whether I need to do/add something else...

### DIFF
--- a/.changeset/smooth-readers-look.md
+++ b/.changeset/smooth-readers-look.md
@@ -2,4 +2,5 @@
 "grafserv": patch
 ---
 
-Add support for event streams (e.g. schema watch mode) to Hono Grafserv adaptor. (Thanks @malyzeli!)
+Add support for event streams (e.g. schema watch mode) to Hono Grafserv adaptor.
+(Thanks @malyzeli!)


benjie:
Seems good to me, but I don't know Hono; please run `git pull --rebase && yarn && yarn lint:fix && git commit -am"Lint" && git push`

What do you think @olup / @hobbescodes / @JendaT?

malyzeli:
@benjie Formatting fixed. 

I guess not many people are using postgraphile on Hono due to missing watch mode, having to constantly restart during development is kinda clumsy - at least that was the reason we kept using Koa on our previous project with @JendaT. :-D 

Finally I had some time to tinker with code so I migrated to Hono and tried to fix it. Not sure if that buffer stream response is used for anything else, because GraphQL API was working as expected even without it. Next step is to enable subscriptions over websocket and then I can send PR for docs update. :-)

benjie:
I don’t think you’d need to restart without this, just refresh schema in GraphiQL manually?

benjie:
This seems like a straightforward fix and it’s not likely to break anyone’s environment so I’m going to go ahead and merge it; thanks @malyzeli!

malyzeli:
> I don’t think you’d need to restart without this, just refresh schema in GraphiQL manually?

That's what I expected as well, but seems it doesn't work that way. When `grafserv.watch` is not enabled in the preset, refreshing schema in GraphiQL does nothing - to my surprise even browser refresh does not fetch updated schema. Didn't have courage for digging too deep into internals, but it looks like event stream is the only channel for schema updates, so it was easier to implement that. :-)

JendaT:
Actually the real reason to use koa over hono was erratic behaviour of hono when reloading the app in watch mode and different context handling than in koa. 

Anyway I think this fix lgtm :-) let's hope graphile just got another occasional contributor!

benjie:
You’re right @malyzeli I tied those behaviours together 🤔

PR #2809, opened 2025-11-12
Enable last update author and time in Docusaurus config

mrmntte:
Assuming `publish-websites.sh` is the code to deploy the Grafast documentation, and that during the build process there is [access to the commit history for extracting the dates](https://github.com/facebook/docusaurus/issues/2798#issuecomment-636602951), these simple changes should display the date next to "Edit this page".


benjie (approved):
Awesome; thanks! I've removed the author because it doesn't seem as important, but have added the update time to our other docs sites too :raised_hands:

PR #2820, opened 2025-11-19
Experiment with json_to_recordset

benjie:
The `json_array_elements` code is really verbose. Maybe `json_to_recordset` will be more straightforward both for humans to read and for PostgreSQL to parse (and maybe even execute?)


benjie:
Put on hold because I need to stop fiddling and get V5 out. Can change this later.

PR #2827, opened 2025-11-21
website: Grafast glossary editorial

jemgillam:
## Description

- Terms in two sections: GraphQL terms and Grafast specific terms
- Terms in alphabetical order
- When the link is to an anchor on the same page, I have bolded it.


jemgillam:
The diff isn't great, but you should be able to look at each commit easily

PR #2911, opened 2026-01-27
Use .ts file extensions in source code

benjie:
Helps move us towards being able to use Node's native type stripping.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -9929,11 +9929,11 @@ __metadata:
   linkType: hard
 
 "baseline-browser-mapping@npm:^2.8.25":
-  version: 2.8.31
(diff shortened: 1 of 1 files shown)


grafast/grafast/src/steps/access.ts
```diff
@@ -72,13 +72,15 @@ export class AccessStep<TData> extends UnbatchedStep<TData> {
   allowMultipleOptimizations = true;
   public readonly path: (string | number | symbol)[];
   private readonly hasSymbols: boolean;
+  public readonly fallback?: any;
```
jemgillam, line 75:
```suggestion
  public readonly fallback: any;
```

PR #2937, opened 2026-02-04
Ensure we have no dangling promises

benjie:
We don't want to hit `unhandledPromiseRejection` errors, so let's lint for them.

Here we use `void` if we're (fairly) certain the relevant code won't cause an unhandled promise rejection - typically because we already have a `try/catch` inside the `async` function, or we handle the promise elsewhere.

The AI did the groundwork, I then rolled a lot of it back and neatened it up. I'm unsure of the changes made to the h3 adaptor...

### DIFF
--- a/.changeset/crisp-ghosts-relate.md
+++ b/.changeset/crisp-ghosts-relate.md
@@ -1,0 +1,14 @@
+---
+"graphile-build-pg": patch
+"@grafserv/persisted": patch
+"ruru-components": patch
+"graphile-config": patch
+"@dataplan/pg": patch
+"grafserv": patch
+"grafast": patch
+"ruru": patch
+"postgraphile": patch
+---
+
+Eliminate dangling promises, reducing chance of process exit due to unhandled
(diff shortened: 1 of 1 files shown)


grafast/grafserv/src/servers/h3/v1/index.ts
```diff
@@ -287,24 +287,23 @@ export class H3Grafserv extends GrafservBase {
...
       },
       message(peer, message) {
-        clients.get(peer)?.handleMessage?.(message.text());
+        return clients.get(peer)?.handleMessage?.(message.text());
       },
       close(peer, details) {
-        clients.get(peer)?.closed?.(details.code, details.reason);
+        const client = clients.get(peer);
         clients.delete(peer);
+        return client?.closed?.(details.code, details.reason);
```
benjie, line 306: @Dodobibi I think these changes are safe/desired? (Looks like it's restoring it closer back to what you originally proposed, now the TypeScript dance is no longer required.)

PR #2947, opened 2026-02-10
Layer plan test and flag handling improvements

### LINKED ISSUE #2501: Check over the flagUnion logic
In particular we need to ensure it's correctly calculated and we've not missed updating it anywhere.

benjaie:
- Fixes #2501

Fixes issues with `null` in lists in abstract positions.

Also fixes issues with `PromiseRejectionHandledWarning` due to delayed handling of errors in lists in some circumstances.

### DIFF
--- a/grafast/grafast/__tests__/combined-layer-plan-test.ts
+++ b/grafast/grafast/__tests__/combined-layer-plan-test.ts
@@ -13,6 +13,7 @@ import {
   InterfacePlan,
   lambda,
   makeGrafastSchema,
+  Step,
 } from "../dist/index.js";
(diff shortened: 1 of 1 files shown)


grafast/grafast/src/engine/LayerPlan.ts
```diff
@@ -855,6 +856,9 @@ export class LayerPlan<TReason extends LayerPlanReason = LayerPlanReason> {
                   flags = flags | FLAG_NULL;
                 }
               }
+              if (flags !== NO_FLAGS) {
+                flagUnion |= flags;
+              }
```
benjaie: Might be faster to just do 
```suggestion
              flagUnion |= flags;
```

PR #2955, opened 2026-02-18
Update missing index detection query

dargmuesli:
## Description

This:

```sql
where indrelid = conrelid
and conkey = indkey
or (array_length(indkey, 1) > 1 and indkey @> conkey)
```

is parsed by PostgreSQL as:

```sql
(
  indrelid = conrelid
  and conkey = indkey
)
OR
(
  array_length(indkey, 1) > 1
  and indkey @> conkey
)
```

Which means any multi-column index anywhere that happens to contain the FK columns will satisfy the EXISTS, even if it’s on a different table. Result: NOT EXISTS becomes false for every FK → empty set. The fix should be to group the index conditions under the same table constraint.

## Performance impact

unknown

## Security impact

unknown

## Checklist

- [ ] I've added tests for the new feature, and `yarn test` passes. (docs only)

### DIFF
--- a/postgraphile/website/postgraphile/postgresql-indexes.md
+++ b/postgraphile/website/postgraphile/postgresql-indexes.md
@@ -44,7 +44,8 @@ with indexed_tables as (
       i.relname as index_name,
       string_agg(a.attname::text, ', ' order by array_position(ix.indkey, a.attnum)) as column_names,
       ix.indrelid,
-      array_agg(a.attnum order by array_position(ix.indkey, a.attnum))::smallint[] as indkey
+      -- cast int2vector to smallint[]
+      string_to_array(ix.indkey::text, ' ')::smallint[] as indkey
   from pg_class i
   join pg_index ix on i.oid = ix.indrelid
   join pg_class t on ix.indrelid = t.oid


postgraphile/website/postgraphile/postgresql-indexes.md
```diff
@@ -42,9 +42,8 @@ with indexed_tables as (
       ns.nspname,
       t.relname as table_name,
       i.relname as index_name,
-      array_to_string(array_agg(a.attname), ', ') as column_names,
-      ix.indrelid,
-      string_to_array(ix.indkey::text, ' ')::smallint[] as indkey
+      array_agg(a.attnum order by array_position(ix.indkey, a.attnum))::smallint[] as indkey,
+      ix.indrelid
```
benjie: I think the column names may have been deleted by mistake? They're typically the more useful part of this

```suggestion
      string_agg(a.attname::text, ', ' order by array_position(ix.indkey, a.attnum)) as column_names,
      ix.indrelid,
      array_agg(a.attnum order by array_position(ix.indkey, a.attnum))::smallint[] as indkey
```

postgraphile/website/postgraphile/postgresql-indexes.md
```diff
@@ -77,8 +76,10 @@ and not exists(
   select 1
   from indexed_tables
   where indrelid = conrelid
-  and conkey = indkey
-  or (array_length(indkey, 1) > 1 and indkey @> conkey)
+  and (
+    conkey = indkey
+    or (array_length(indkey, 1) > array_length(conkey, 1) and indkey @> conkey)
```
benjie: With some extensions, the `@>` will produce:

```
ERROR:  42725: operator is not unique: smallint[] @> smallint[]
```

So we'll address that too:

```suggestion
    or (
      array_length(indkey, 1) > array_length(conkey, 1)
      and indkey::int[] @> conkey::int[]
    )
```

postgraphile/website/postgraphile/postgresql-indexes.md
```diff
@@ -42,9 +42,9 @@ with indexed_tables as (
       ns.nspname,
       t.relname as table_name,
       i.relname as index_name,
-      array_to_string(array_agg(a.attname), ', ') as column_names,
+      string_agg(a.attname::text, ', ' order by array_position(ix.indkey, a.attnum)) as column_names,
       ix.indrelid,
-      string_to_array(ix.indkey::text, ' ')::smallint[] as indkey
+      array_agg(a.attnum order by array_position(ix.indkey, a.attnum))::smallint[] as indkey
```
benjie: Actually your expression here breaks the logic for complex indexes, where the value here would be `0` (which won't match an attnum), and also makes this expression dependent on the `where` clause.

We just want to cast the `int2vector` to a `smallint[]`, and this is a compressed way to do so; restoring the old code:

```suggestion
      -- cast int2vector to smallint[]
      string_to_array(ix.indkey::text, ' ')::smallint[] as indkey
```

benjie (approved):

PR #2965, opened 2026-02-24
Export shrinkage: more plan tidying & EXPORTABLE_ARRAY_CLONE

benjie:


grafast/dataplan-pg/src/datasource.ts
```diff
@@ -1163,18 +1163,12 @@ export function makeRegistry<
...
-          }))
-        : rawConfig.parameters,
-    };
-    const resource = new PgResource(registry, resourceConfig) as any;
+    addExecutor(rawConfig.executor);
+    addCodec(rawConfig.codec);
+    rawConfig.parameters?.forEach((p: PgResourceParameter) =>
+      addCodec(p.codec),
+    );
+    const resource = new PgResource(registry, rawConfig) as any;
```
benjie, line 1171: Really not sure why we used to create derivative objects for this... 🤔

PR #2973, opened 2026-03-03
pg-sql2: restore sql.TRUE, sql.FALSE and sql.NULL to ease v5 migration

benjie:
These are just copies of sql.true, sql.false and sql.null respectively.


.changeset/quick-lights-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+"pg-sql2": patch
```
benjie, line 3:
```suggestion
"graphile-utils": patch
"pg-sql2": patch
"postgraphile": patch
```

PR #2985, opened 2026-03-09
Extend documentation with a guide for plan resolvers (#244)

msotnikov:
## Description                                                                                                                                                                                                   
                                                                                                                                                                                                                 
  Addresses https://github.com/graphile/crystal-pre-merge/issues/244 — general recommendations when writing plan resolvers.                                                                                        
                                                            
  **Documentation:**                                                                                                                                                                                               
                                                            
  - Added a new **"Best practices"** page (`grafast/website/grafast/plan-resolvers/best-practices.md`) covering the four recommendations from the issue:
    1. **Deep argument extraction** — use `getRaw(["path", "to", "value"])` or `$`-prefixed destructuring instead of shallow extraction followed by lambda transforms
    2. **Prefer custom steps over `lambda`** — comparison table, guidance on when lambda is appropriate vs when to create a step class, worked example of a custom step
    3. **File-scoped lambda callbacks** — why inline anonymous functions defeat deduplication, with do/don't examples
    4. **Avoid `try`/`catch` in plan resolvers** — why imperative error handling doesn't fit the declarative model, with examples using `inhibitOnNull()` and `trap()` instead
  - Added a **"When to use something else"** section to `lambda.md` pointing users toward `loadOne`/`loadMany`, custom steps, and `sideEffect()` before they reach for lambda
  - Added a **tip callout** linking to the best practices page from the top of `plan-resolvers/index.mdx`

  **Runtime warning:**

  - In dev mode (`NODE_ENV=development`), `lambda()` now emits a `console.warn` when called with an anonymous (unnamed) callback function, since such callbacks cannot be deduplicated. The warning fires once per
  unique callback reference (tracked via `WeakSet`).

  ## Performance impact

  No impact in production. In development mode, a single `WeakSet` lookup is added per `lambda()` call (negligible).

  ## Security impact

  None.

  ## Checklist

### DIFF
--- a/grafast/website/grafast/plan-resolvers/best-practices.md
+++ b/grafast/website/grafast/plan-resolvers/best-practices.md
@@ -78,11 +78,11 @@ class.
 
 ### Comparison
 
-| | `lambda` | `loadOne` | Custom step |
-|---|---|---|
-| Batching | No &mdash; called once per value | Yes, batched and uniqued |  Yes with full control |
-| Deduplication | Only if callback is the same reference | Only if callback is the same reference | Full control via `deduplicate()` |
-| Optimization | None | Many automatic optimizations | Full control via `optimize()` / `finalize()` / `execute()` |
+|               | `lambda`                               | `loadOne`                              | Custom step                                                |
+| ------------- | -------------------------------------- | -------------------------------------- | ---------------------------------------------------------- |
+| Batching      | No &mdash; called once per value       | Yes, batched and uniqued               | Yes with full control                                      |
+| Deduplication | Only if callback is the same reference | Only if callback is the same reference | Full control via `deduplicate()`                           |
+| Optimization  | None                                   | Many automatic optimizations           | Full control via `optimize()` / `finalize()` / `execute()` |
 
 :::note[Batching may not be relevant to mutations/side effects]
 
@@ -95,7 +95,6 @@ a mutation field plan resolver, and no other steps should be needed.
 
 Side-effects should not happen in other (non-Mutation) plan resolvers,
 however any step can be marked as having side effects via:
-
 
 ### When to use loadOne/loadMany
 
@@ -193,8 +192,11 @@ export class GoogleDriveFileStep extends Step<GoogleDriveFile> {
 
   // Execute once for the entire batch
   async execute(details: ExecutionDetails) {
-    const { values: [fileIdEv], indexMap } = details;
-    const uniqueIds = [...new Set(indexMap(i => fileIdEV.at(i)))];
+    const {
+      values: [fileIdEv],
+      indexMap,
+    } = details;
+    const uniqueIds = [...new Set(indexMap((i) => fileIdEV.at(i)))];
 
     // One HTTP request for the whole batch
     const url = new URL("https://www.googleapis.com/drive/v3/files");
@@ -370,9 +372,7 @@ details on when and how to use these.
 
 ## Summary
 
-| Recommendation | Why |
-|---|---|
-| [Extract arguments deeply](#extract-arguments-deeply) | Fewer intermediate steps, better optimization |
-| [Use `loadOne`/`loadMany` for I/O; reserve `lambda` for trivial sync transforms](#choose-the-right-step-type) | Batching, deduplication, optimization |
-| [File-scoped callbacks](#define-callbacks-at-file-scope) | Enables deduplication |
-| [No plan resolver `try`/`catch`](#dont-use-trycatch-in-plan-resolvers) | Plan resolvers are declarative; use `maskError` or flow control steps |
+- [Extract arguments deeply](#extract-arguments-deeply) for fewer intermediate steps and better optimization
+- [Use `loadOne`/`loadMany` for I/O](#choose-the-right-step-type) to ensure you're making the most of batching, deduplication, and optimization; `lambda()` should only be used for trivial sync transforms
+- [Use file-scoped callbacks](#define-callbacks-at-file-scope) so steps can be deduplicated
+- [Avoid plan resolver `try`/`catch`](#dont-use-trycatch-in-plan-resolvers) because plan resolvers are declarative; use `maskError` or flow control steps


grafast/grafast/src/steps/lambda.ts
```diff
@@ -72,6 +75,15 @@ function lambda<const TInMultistep extends Multistep, TOut>(
...
   }
+  if (isDev && !fn.name && !warnedCallbacks.has(fn)) {
+    warnedCallbacks.add(fn);
+    console.warn(
+      `lambda() was called with an anonymous (inline) callback function. ` +
+        `This prevents deduplication. Define the callback at file scope or ` +
+        `give it a name for better optimization. ` +
+        `See: https://grafast.org/grafast/standard-steps/lambda#define-callback-in-top-scope`,
+    );
+  }
```
benjie: I'm not super keen on this; I'd rather handle it with a lint rule later. The main reason is that just not having a name is not sufficient to know that it was inline; for example we recommend `EXPORTABLE(...)` usage everywhere; however:

```ts
const { EXPORTABLE } = require('graphile-export')
const TWO = 2
const addTwo = EXPORTABLE((TWO) => (n) => n + TWO, [TWO])
console.log(addTwo.name)
```

This'll output `''` since there's no name for this function.

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,245 @@
...
+function bookCount_plan($parent, fieldArgs) {
+  const { $search, $filter } = fieldArgs;
+  const { $author, $publishedAfter } = $filter;
+  // ...
+}
+```
+
+Both `.getRaw()` with a path array and the `$`-prefixed destructuring give
+Gra*fast* direct visibility into exactly which leaf values you need, allowing it
+to skip unnecessary work and optimize the plan more aggressively.
```
benjie: This was a major concern before the "eliminate eval" logic came in. It's no longer _as_ important, but probably worth keeping as a best practice anyway - it may become more significant again in future optimizations.

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Prefer custom steps over `lambda`
+
+[`lambda()`](../standard-steps/lambda.md) is an escape hatch &mdash; it
+processes values **one at a time** rather than in batches. This is fine for
+trivial synchronous transforms (string concatenation, simple math), but for
+anything more complex you should create a custom step class.
```
benjie: No need to jump to custom class; `loadOne()`/`loadMany()` are the bread and butter steps - most people should just use those (at least to start with).

```suggestion
anything more complex you should use a batch step such as `loadOne()`,
`loadMany()`, or a custom class.
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Why custom steps are better
+
+| | `lambda` | Custom step |
+|---|---|---|
+| Batching | No &mdash; called once per value | Yes &mdash; `execute()` receives the full batch |
+| Deduplication | Only if callback is the same reference | Full control via `deduplicate()` |
+| Optimization | None | Can implement `optimize()` |
+| Side effects | Not supported (use `sideEffect()`) | Full control via `hasSideEffects` |
```
benjie: I've added `loadOne()` to make it clear what the trade-offs are between lambda vs loadOne vs full DIY. `loadOne()` is generally the sweet-spot.

I've also separated out the side effects point.

```suggestion
| | `lambda` | `loadOne` | Custom step |
|---|---|---|
| Batching | No &mdash; called once per value | Yes, batched and uniqued |  Yes with full control |
| Deduplication | Only if callback is the same reference | Only if callback is the same reference | Full control via `deduplicate()` |
| Optimization | None | Many automatic optimizations | Full control via `optimize()` / `finalize()` / `execute()` |

:::note[Batching may not be relevant to mutations/side effects]

According to
[the GraphQL spec](https://spec.graphql.org/September2025/#sel-GANVLDCB-BBqFxyV),
side effects may only occur in `Mutation` fields, and these fields are
executed serially. This gives no opportunity for batching. Typically
`sideEffect()` (like `lambda()`, but with side effects) is suitable for
a mutation field plan resolver, and no other steps should be needed.

Side-effects should not happen in other (non-Mutation) plan resolvers,
however any step can be marked as having side effects via:

```ts
$step.hasSideEffects = true;
```

This should generally be avoided for the reasons above.

:::
```

File: grafast/website/grafast/plan-resolvers/best-practices.md [hunk 5]
```diff
@@ -0,0 +1,245 @@
...
+- Simple math: `lambda($n, (n) => n + 1, true)`
+- Trivial data mapping that doesn't benefit from batching
+
+### When to create a custom step
+
+- The transform involves I/O or async work &mdash; use
+  [`loadOne()`](../standard-steps/loadOne.md) /
+  [`loadMany()`](../standard-steps/loadMany.md) instead
+- You want deduplication (e.g. multiple fields perform the same transform)
+- The logic is non-trivial or would benefit from batching
```
benjie:
```suggestion
### When to use loadOne/loadMany

Use [`loadOne()`](../standard-steps/loadOne.md) to loading a single record for each input, or [`loadMany()`](../standard-steps/loadMany.md) to load a collection of records for each input, when:

- you have async work (except mutations),
- you have I/O work (except mutations), or
- the code would benefit from batching.

### When to create a custom step

Custom steps can be used for any purpose, typically you'll want
to build your own step classes if:

- You want to expose your own helper APIs (e.g. custom methods on your step)
- You want full control over execution (e.g. if loadOne/loadMany's optimizations don't fit your needs)
- You want full control over deduplication (reducing redundant work)
- You want full control over plan optimization (in particular eliminating over- and under- fetching by communicating with other steps)
- You want to do custom work one time only for your step (custom `finalize()`)
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,245 @@
...
+
+  unbatchedExecute(_extra: UnbatchedExecutionExtra, firstName: string, lastName: string) {
+    return `${firstName} ${lastName}`;
+  }
+}
+
+export function fullName($first: ExecutableStep<string>, $last: ExecutableStep<string>) {
+  return new FullNameStep($first, $last);
+}
+```
```
benjie, line 235: Let's not use an UnbatchedStep as an example, because... it's explicitly unbatched. It's very rare a user would ever need an unbatched step (they could just use `lambda`). Instead; maybe use a batch-capable API that also allows you to determine which features to fetch - e.g.

```
GET https://www.googleapis.com/drive/v3/files
  ?q=id in ('1abc','2def','3ghi')
  &fields=files(id,name,owners(emailAddress))
```

This would demonstrate the value of building your own step class with helper methods and custom execution.

Ref:

- https://developers.google.com/workspace/drive/api/reference/rest/v3/files/list
- https://developers.google.com/workspace/drive/api/guides/fields-parameter?utm_source=chatgpt.com

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Define lambda callbacks at file scope
+
+If you do use `lambda`, **always define the callback at file/module scope**
+(or import it from another file) rather than inline. Gra*fast* deduplicates
+lambda steps by comparing the callback reference &mdash; inline arrow functions
+create a new reference on every call, defeating deduplication.
```
benjie: Encourage this for _all_ functions; in particular:

- `lambda`
- `loadOne`
- `loadMany`
- `applyInput`

Less importantly:

- `each`
- `filter`
- `groupBy`
- `partitionByIndex`
- `sideEffect`

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,245 @@
...
+  },
+};
+```
+
+## Don't use `try`/`catch` in plan resolvers
+
+Plan resolvers run at **plan-time**, not execution-time. They build a
+declarative graph of steps &mdash; think of them like React component render
+functions where steps are like hooks. Using `try`/`catch` introduces imperative
+control flow that doesn't fit this model.
```
benjie:
```suggestion
Plan resolvers run at **plan-time**, before any input values are known and
before any data has been fetched. They build a declarative graph of steps
that will be executed later. Since they run at plan-time, `try`/`catch`
will only catch planning errors (which shouldn't really happen!) - it will
not catch execution-time errors (i.e. errors resulting from
fetching/manipulating real data).
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Why it doesn't work
+
+- Plan resolvers don't execute your data-fetching logic &mdash; they only
+  **describe** it. A `try` block around step creation doesn't catch runtime
+  data errors because those errors happen later, during execution.
+- Wrapping step creation in `try`/`catch` can mask plan-time programming
+  errors that should be fixed, not caught.
+- It suggests a misunderstanding of the plan/execute separation.
```
benjie, line 316: I really like this last bullet point.

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,245 @@
...
+  try {
+    const $authorId = $post.get("authorId");
+    return loadOne($authorId, batchGetAuthorById);
+  } catch (e) {
+    return constant(null);
+  }
+}
+```
+
+### Do: use flow control steps
```
benjie: I think most users should only use flow control steps in rare occasions.

````suggestion
### Do: use `maskError` or similar to process errors

GraphQL is designed to continue in the face of errors, allowing for
"partial success"; however, you may wish to relabel an error when
presenting it to a user. To do so, use Grafserv's `maskError`
functionality, or similar methods that come with your server of choice.

### If necessary: use flow control steps
````

File: grafast/website/grafast/plan-resolvers/best-practices.md [hunk 11]
```diff
@@ -0,0 +1,245 @@
...
+
+function post_author_plan($post) {
+  const $authorId = $post.get("authorId");
+
+  // Guard against null authorId — skip the load entirely
+  const $guardedId = inhibitOnNull($authorId);
+
+  // Load the author; if it errors, convert to null
+  const $author = loadOne($guardedId, batchGetAuthorById);
+  return trap($author, TRAP_ERROR);
```
benjie: Be more explicit about what we're doing here:

```suggestion
  return trap($author, TRAP_ERROR, { valueForError: "NULL" });
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+The key flow control steps are:
+
+- [`inhibitOnNull()`](../standard-steps/inhibitOnNull.mdx) &mdash; suppresses
+  downstream work when a value is `null`
+- [`assertNotNull()`](../standard-steps/assertNotNull.mdx) &mdash; turns
+  `null` into a `SafeError` visible to clients
+- [`trap()`](../standard-steps/trap.mdx) &mdash; recovers inhibited or errored
+  values back into ordinary data (e.g. `null` or an empty list)
```
benjie:
```suggestion
- [`trap()`](../standard-steps/trap.mdx) &mdash; recovers inhibited or errored
  values back into ordinary data (e.g. `null`, an empty list, or the error as
  a simple "data value" rather than an exception)
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,245 @@
...
+
+See [Thinking in plans: Flow control](../flow.mdx#flow-control) for more
+details on when and how to use these.
+
+## Summary
+
+| Recommendation | Why |
+|---|---|
+| [Extract arguments deeply](#extract-arguments-deeply) | Fewer intermediate steps, better optimization |
+| [Prefer custom steps over `lambda`](#prefer-custom-steps-over-lambda) | Batching, deduplication, optimization |
```
benjie:
```suggestion
| [Avoid `lambda`, except for inexpensive synchronous work](#prefer-custom-steps-over-lambda) | Batching, deduplication, optimization |
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,245 @@
...
+| [File-scoped lambda callbacks](#define-lambda-callbacks-at-file-scope) | Enables deduplication |
```
benjie:
```suggestion
| [File-scoped callbacks](#define-lambda-callbacks-at-file-scope) | Enables deduplication |
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,245 @@
...
+| [No `try`/`catch`](#dont-use-trycatch-in-plan-resolvers) | Plan resolvers are declarative; use flow control steps |
```
benjie:
```suggestion
| [No plan resolver `try`/`catch`](#dont-use-trycatch-in-plan-resolvers) | Plan resolvers are declarative; use `maskError` or flow control steps |
```

grafast/website/grafast/plan-resolvers/index.mdx
```diff
@@ -25,6 +25,15 @@ as steps to be populated at execution time for each request.
 
 :::
 
+:::tip[Best practices]
+
+For recommendations on writing efficient and correct plan resolvers &mdash;
+including deep argument extraction, when to use custom steps vs `lambda`,
+and how to handle errors declaratively &mdash; see the
```
benjie:
```suggestion
including deep argument extraction; when to use custom steps, `loadOne()`, `loadMany()`, or `lambda()`;
and how to handle errors declaratively &mdash; see the
```

grafast/website/grafast/standard-steps/lambda.md
```diff
@@ -80,6 +80,22 @@ not do this unless you are certain!
...
+- **Non-trivial transforms that appear in multiple fields** &rarr; create a
+  [custom step class](../step-classes.mdx) with `deduplicate()` support
+- **Side effects** &rarr; use [`sideEffect()`](/grafast/standard-steps/sideEffect)
+
+`lambda` is best reserved for trivial, synchronous, pure transforms such as
+string concatenation or simple arithmetic.
+
+See [Plan resolver best practices](../plan-resolvers/best-practices.md) for
+more guidance.
+
```
benjie: We already have two notes on this page warning about lambda not batching; perhaps this should be folded into those?

msotnikov:
Thank you for the detailed comments. I've accepted all the suggestions and fixed the remaining. I also adjusted the focus to provide a more neutral overview.

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,369 @@
...
+  const { $author, $publishedAfter } = $filter;
+  // ...
+}
+```
+
+Both `.getRaw()` with a path array and the `$`-prefixed destructuring give
+Gra*fast* direct visibility into exactly which leaf values you need. While
+current optimizations (such as the "eliminate eval" pass) reduce the impact of
+shallow extraction, deep extraction remains a good habit — it makes your intent
+explicit and may become more significant again in future optimizations.
```
benjie:
```suggestion
Both `.getRaw()` with a path array and the `$`-prefixed destructuring give
Gra*fast* direct visibility into exactly which leaf values you need. While
current optimizations make limited use of this, deep extraction remains a
good habit — it makes your intent explicit and may become more significant
in future optimizations.
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,369 @@
...
+side effects may only occur in `Mutation` fields, and these fields are
+executed serially. This gives no opportunity for batching. Typically
+`sideEffect()` (like `lambda()`, but with side effects) is suitable for
+a mutation field plan resolver, and no other steps should be needed.
+
+Side-effects should not happen in other (non-Mutation) plan resolvers,
+however any step can be marked as having side effects via:
+
+```ts
+$step.hasSideEffects = true;
```
benjie:
````suggestion
```ts
$step.hasSideEffects = true;
```

This should generally be avoided for the reasons above.

:::
````

File: grafast/website/grafast/plan-resolvers/best-practices.md [hunk 20]
```diff
@@ -0,0 +1,369 @@
...
+
+Use [`loadOne()`](../standard-steps/loadOne.md) to load a single record for each input, or [`loadMany()`](../standard-steps/loadMany.md) to load a collection of records for each input, when:
+
+- you have async work (except mutations),
+- you have I/O work (except mutations), or
+- the code would benefit from batching.
+
+### When `lambda` is appropriate
+
+- Concatenating strings: `lambda([$first, $last], ([f, l]) => \`${f} ${l}\`, true)`
```
benjie: Simpler escaping:

- Concatenating strings: ``lambda([$first, $last], ([f, l]) => `${f} ${l}`, true)``

```suggestion
- Concatenating strings: ``lambda([$first, $last], ([f, l]) => `${f} ${l}`, true)``
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,369 @@
...
+API into a custom step that batches multiple file-ID lookups into a single HTTP
+request and uses the
+[`fields` parameter](https://developers.google.com/workspace/drive/api/guides/fields-parameter)
+to fetch only the data the GraphQL query actually needs.
+
+```ts
+import { ExecutableStep, ExecutionExtra, access } from "grafast";
+
+/** Loads Google Drive file metadata, batching multiple IDs into one API call. */
+export class GoogleDriveFileStep extends ExecutableStep<GoogleDriveFile> {
```
benjie: `ExecutableStep` is deprecated, it's just an alias for `Step` now.

```suggestion
import { Step, ExecutionExtra, access } from "grafast";

/** Loads Google Drive file metadata, batching multiple IDs into one API call. */
export class GoogleDriveFileStep extends Step<GoogleDriveFile> {
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,369 @@
...
+  static $$export = {
+    moduleName: "my-app",
+    exportName: "GoogleDriveFileStep",
+  };
```
benjie, line 152: This probably warrants its own explanation

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,369 @@
...
+  }
+
+  /**
+   * Helper: declare that downstream steps need a particular field.
+   * Returns an access step that reads the field from the result.
+   */
+  getField(name: string): ExecutableStep {
+    this.fieldPaths.add(name);
+    return access(this, name);
+  }
```
benjie: Should probably use `get` for this as is convention:

https://grafast.org/grafast/step-classes/#get


```suggestion
  get(name: string): Step {
    this.fieldPaths.add(name);
    return access(this, name);
  }
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,369 @@
...
+
+  /**
+   * Helper: declare that downstream steps need a nested field.
+   */
+  getNestedField(parent: string, child: string): ExecutableStep {
```
benjie:
```suggestion
  getNestedField(parent: string, child: string): Step {
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,369 @@
...
+  // Deduplicate steps that request the same file ID
+  deduplicate(peers: GoogleDriveFileStep[]): GoogleDriveFileStep[] {
+    return peers;
+  }
+
+  // Merge requested fields from deduplicated peers
+  deduplicatedWith(peers: GoogleDriveFileStep[]): void {
+    for (const peer of peers) {
+      for (const field of peer.fieldPaths) {
+        this.fieldPaths.add(field);
```
benjie: We've been deleted, so _they_ need to know to get our stuff.

```suggestion
      for (const field of this.fieldPaths) {
        peer.fieldPaths.add(field);
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,369 @@
...
+    }
+  }
+
+  // Execute once for the entire batch
+  async execute(
+    _extra: ExecutionExtra,
+    [fileIds]: [string[]],
+  ): Promise<(GoogleDriveFile | null)[]> {
+    const uniqueIds = [...new Set(fileIds)];
+    const fields = `files(${[...this.fieldPaths].join(",")})`;
```
benjie: The pattern of `execute` you have here hasn't existed for a while.

```suggestion
  async execute(details: ExecutionDetails) {
    const { values: [fileIdEv], indexMap } = details;
    const uniqueIds = [...new Set(indexMap(i => fileIdEV.at(i)))];
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,369 @@
...
+    const url = new URL("https://www.googleapis.com/drive/v3/files");
+    url.searchParams.set("q", uniqueIds.map((id) => `'${id}'`).join(" or "));
+    url.searchParams.set("fields", fields);
+
+    const response = await fetch(url);
+    const { files } = await response.json();
+
+    // Build a lookup map and return results in the original order
+    const byId = new Map(files.map((f: GoogleDriveFile) => [f.id, f]));
+    return fileIds.map((id) => byId.get(id) ?? null);
```
benjie:
```suggestion
    return indexMap.map((i) => {
      const fileId = fileIdEV.at(i);
      return byId.get(fileId) ?? null;
    });
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,369 @@
...
+  }
+}
+
+export function googleDriveFile($fileId: ExecutableStep<string>) {
```
benjie:
```suggestion
export function googleDriveFile($fileId: Step<string>) {
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,369 @@
...
+
+Usage in a plan resolver:
+
+```ts
+function file_plan($parent) {
+  const $fileId = $parent.get("driveFileId");
+  const $file = googleDriveFile($fileId);
+
+  // Only the fields actually requested by the GraphQL query are fetched
+  const $name = $file.getField("name");
```
benjie:
```suggestion
  const $name = $file.get("name");
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,369 @@
...
+  // ...
+}
+```
+
+## Define callbacks at file scope
+
+Many step functions accept a callback. **Always define these callbacks at
+file/module scope** (or import them from another file) rather than inline.
+Gra*fast* deduplicates steps by comparing the callback reference &mdash; inline
+arrow functions create a new reference on every call, defeating deduplication.
```
benjie:
```suggestion
functions create a new reference on every call, defeating deduplication.
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,369 @@
...
+Named functions also produce more readable debug output and `explain` plans.
+
+This applies to the following functions, in order of importance:
```
benjie:
```suggestion
This applies to the following functions:
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,369 @@
...
+
+**Most important** &mdash; commonly used in nearly every schema:
```
benjie:
```suggestion
**Most important** &mdash; lack of deduplication here can be expensive:
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,369 @@
...
+
+- [`lambda()`](../standard-steps/lambda.md)
+- [`loadOne()`](../standard-steps/loadOne.md)
+- [`loadMany()`](../standard-steps/loadMany.md)
+- [`applyInput()`](../standard-steps/applyInput.md)
+
+**Also recommended** &mdash; less common, but the same principle applies:
```
benjie:
```suggestion
**Also recommended** &mdash; less critical, but the same principle applies:
```

grafast/website/grafast/plan-resolvers/index.mdx
```diff
@@ -25,6 +25,15 @@ as steps to be populated at execution time for each request.
...
 
+:::tip[Best practices]
+
+For recommendations on writing efficient and correct plan resolvers &mdash;
+including deep argument extraction; when to use custom steps, `loadOne()`, `loadMany()`, or `lambda()`;
+and how to handle errors declaratively &mdash; see the
+[Best practices](./best-practices.md) guide.
+
+:::
+
```
benjie: Let's move this lower

```suggestion
```

grafast/website/grafast/plan-resolvers/index.mdx
benjie, line 52:
```suggestion
resolver](#default-plan-resolver).

:::tip[Best practices]

For recommendations on writing efficient and correct plan resolvers &mdash;
including deep argument extraction; when to use custom steps, `loadOne()`, `loadMany()`, or `lambda()`;
and how to handle errors declaratively &mdash; see the
[Best practices](./best-practices.md) guide.

:::
```

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,372 @@
...
+
+This example wraps the
+[Google Drive files.list](https://developers.google.com/workspace/drive/api/reference/rest/v3/files/list)
+API into a custom step that batches multiple file-ID lookups into a single HTTP
+request and uses the
+[`fields` parameter](https://developers.google.com/workspace/drive/api/guides/fields-parameter)
+to fetch only the data the GraphQL query actually needs.
+
+```ts
+import { Step, ExecutionExtra, access } from "grafast";
```
benjie:
```suggestion
/* DISCLAIMER: this example has not been tested */

import { Step, ExecutionDetails, access } from "grafast";
```

benjie (dismissed):

grafast/website/grafast/plan-resolvers/best-practices.md
```diff
@@ -0,0 +1,383 @@
...
+    }
+  }
+
+  // Execute once for the entire batch
+  async execute(
+    _extra: ExecutionExtra,
+    [fileIds]: [string[]],
+  ): Promise<(GoogleDriveFile | null)[]> {
+    const uniqueIds = [...new Set(fileIds)];
+  async execute(details: ExecutionDetails) {
```
benjie:
```suggestion
  async execute(details: ExecutionDetails) {
```

PR #2997, opened 2026-03-16
Support overloaded computed column functions

### LINKED ISSUE #2972: Functions with same name but different parameter types are silently dropped (overload detection too aggressive)
When multiple functions share the same name but have different parameter types (valid PostgreSQL overloading), PgProceduresPlugin silently drops all overloads, not just duplicates.

**Example**

```
CREATE SCHEMA computed;

CREATE FUNCTION computed.code(sales.orders) RETURNS text AS $$ ... $$ LANGUAGE sql STABLE;
CREATE FUNCTION computed.code(work.jobs) RETURNS text AS $$ ... $$ LANGUAGE sql STABLE;
CREATE FUNCTION computed.code(partner.accounts) RETURNS text AS $$ ... $$ LANGUAGE plpgsql STABLE;
```

These are three distinct PostgreSQL functions (different OIDs, different parameter types). Each takes a different table's row type as its first argument,  making them valid computed column candidates for three different GraphQL types (SalesOrder, WorkJob, PartnerAccount).

Expected: All three appear as code fields on their respective GraphQL types.

Actual: None of them appear. All are silently dropped.

**Cause**

In PgProceduresPlugin, the overload check compares only pronamespace and proname, ignoring proargtypes:

```
  const overload = introspection.procs.find((p) =>
      p.pronamespace === pgProc.pronamespace &&
      p.proname === pgProc.proname &&
      p._id !== pgProc._id
  );
  if (overload) {
      return; // skips this function entirely
  }
```

Since each function finds one of the others as an "overload", all overloads are dropped.

**Why this matters**

This pattern is natural for computed columns: a function named `is_open` that works on `sales.orders`, `sales.orderlines`, and `inbound.deliverylines` is three logically separate computed columns. Each targets a different type and there's no naming conflict in the GraphQL schema.

**Possible fix**

  When the first parameter is a composite (row) type for different tables, these aren't true overloads from a GraphQL perspective, they produce fields on different types. The check could be relaxed to allow functions with the same name when their first parameter refers to a different table, or the resource name could include the table name to disambiguate (e.g., orders_code, jobs_code).

  **Workaround**

Rename each function to include the table name prefix: `computed.orders_code(sales.orders)`, `computed.jobs_code(work.jobs)`, etc. However, this is a lot less nice for SQL-usage of the database. What was previously:
```
SELECT orders.code FROM sales.orders
-- or
SELECT computed.code(order.*) FROM sales.orders
```
Must now become:
```
SELECT computed.orders_code(order.*) AS code FROM sales.orders
```

**Versions**

- postgraphile 5.0.0-rc.9
- graphile-build-pg 5.0.0-rc.7

GeertJohan:
## Description

Support overloaded computed column functions, where the first argument is a table.
Fixes https://github.com/graphile/crystal/issues/2972
Adds a test for this specific scenario.

I deliberately did not modify the PgV4BehaviorPlugin as I suspect users will not appreciate a sudden change in behavior. So this still requires smart tags, just like computed column functions that are NOT overloaded.

AI disclaimer: I used Claude to investigate the codebase and implement parts of this PR, but am actively trying to avoid slop. I personally read all the changes line-by-line, except for the snapshots, and validate the changes to the best of my ability.

Regardless of the use of AI; this is my first PR on this project and first experience with the inner workings of graphile. Although I've been a user for a couple of years now and wrote some custom plugins, diving into a large project like this one always a little overwhelming at first. So please consider carefully and do not hesitate to just close my PR if it's bad, I'll understand.

About the snapshots, I'm not sure if all those changes are correct, this is what recreating the snapshots resulted in but I wonder if it's supposed to do that? It feels like the diff is way too large... I am not familiar enough (yet) with this project to asses this; help needed.

Since procedures need a unique name it probably makes sense to add the table the overloaded procedure is targeting.
For `FUNCTION code(finance.invoices)` we could go with either:
- `code_finance_invoices` or
- `finance_invoices_code`

I decided to go with the first since it has the same order of identifiers as the function definition (`a(b.c)` > `a_b_c`), and less likely to be confused/collide with the `finance.invoices_code(..)` approach of v4.

## Performance impact

Should be relatively low. I used manual loops instead of functional to keep allocations low.

## Security impact

unknown

## Checklist

  - Is dont think this is needed? This PR changes make that overloaded functions work the same as non-overloaded functions. It doesn't 'add' a new feature in that sense.
  - Does not exist.
  - Afaik not a breaking change, this only extends what was already possible with one computed column function. Users must explicitly +typeField on overloaded functions to use this change.

### DIFF
--- a/graphile-build/graphile-build-pg/src/plugins/PgProceduresPlugin.ts
+++ b/graphile-build/graphile-build-pg/src/plugins/PgProceduresPlugin.ts
@@ -606,7 +606,10 @@ export const PgProceduresPlugin: GraphileConfig.Plugin = {
       resourceOptionsByPgProcByService: new Map(),
     }),
     hooks: {
-      async pgIntrospection_proc({ helpers, resolvedPreset }, event) {
+      async pgIntrospection_proc(
+        { helpers, resolvedPreset, inflection },
+        event,
+      ) {
         const { entity: pgProc, serviceName } = event;
 
         const pgService = resolvedPreset.pgServices?.find(
@@ -648,17 +651,35 @@ export const PgProceduresPlugin: GraphileConfig.Plugin = {
           return;
         }
 
-        // We also don’t want procedures that have been defined in our namespace
-        // twice. This leads to duplicate fields in the API which throws an
-        // error. In the future we may support this case. For now though, it is
-        // too complex.
-        const overload = introspection.procs.find(
+        // We don't want procedures whose inflected resource name clashes
+        // with another overload; this would produce duplicate fields.
+        const name = inflection.functionResourceName({ serviceName, pgProc });
+        const forbiddenOverload = introspection.procs.find(
           (p) =>
             p.pronamespace === pgProc.pronamespace &&
             p.proname === pgProc.proname &&
-            p._id !== pgProc._id,
+            p._id !== pgProc._id &&
+            inflection.functionResourceName({
+              serviceName,
+              pgProc: p,
+            }) === name,
         );
-        if (overload) {
+        if (forbiddenOverload) {
+          // Warn if both functions target composite types; the user likely
+          // intended these as computed columns on different tables.
+          const thisFirstArg = pgProc.getArguments().find((a) => a.isIn);
+          const otherFirstArg = forbiddenOverload
+            .getArguments()
+            .find((a) => a.isIn);
+          if (
+            thisFirstArg?.type.typtype === "c" &&
+            otherFirstArg?.type.typtype === "c" &&
+            thisFirstArg.type._id !== otherFirstArg.type._id
+          ) {
+            console.warn(
+              `Skipping function '${namespace!.nspname}.${pgProc.proname}' because it has overloads and they generate the same name. Consider using 'PgFunctionOverloadsPreset' to factor argument types into resource names.`,
+            );
+          }
           return;
         }
 
--- a/.changeset/slow-hounds-impress.md
+++ b/.changeset/slow-hounds-impress.md
@@ -1,0 +1,11 @@
+---
+"graphile-build-pg": patch
+"postgraphile": patch
+---
+
+Add opt-in `PgFunctionOverloadsPreset` which factors the input argument types
+into function resource names (e.g. `code(pets)` becomes `code__pets`), enabling
+support for overloaded functions such as computed column functions targeting
+different tables. Without the preset, overloaded functions are skipped as
+before; a warning is now logged when the skipped overloads look like computed
+columns.
--- a/graphile-build/graphile-build-pg/src/index.ts
+++ b/graphile-build/graphile-build-pg/src/index.ts
@@ -16,6 +16,10 @@ export { PgEnumDomainsPlugin } from "./plugins/PgEnumDomainsPlugin.ts";
 export { PgEnumTablesPlugin } from "./plugins/PgEnumTablesPlugin.ts";
 export { PgFakeConstraintsPlugin } from "./plugins/PgFakeConstraintsPlugin.ts";
 export { PgFirstLastBeforeAfterArgsPlugin } from "./plugins/PgFirstLastBeforeAfterArgsPlugin.ts";
+export {
+  PgFunctionOverloadsPlugin,
+  PgFunctionOverloadsPreset,
+} from "./plugins/PgFunctionOverloadsPlugin.ts";
 export { PgIndexBehaviorsPlugin } from "./plugins/PgIndexBehaviorsPlugin.ts";
 export { PgInterfaceModeUnionAllRowsPlugin } from "./plugins/PgInterfaceModeUnionAllRowsPlugin.ts";
 export { PgIntrospectionPlugin } from "./plugins/PgIntrospectionPlugin.ts";
--- a/graphile-build/graphile-build-pg/src/plugins/PgFunctionOverloadsPlugin.ts
+++ b/graphile-build/graphile-build-pg/src/plugins/PgFunctionOverloadsPlugin.ts
@@ -1,0 +1,63 @@
+import "graphile-config";
+
+import type { PgProc } from "pg-introspection";
+
+import { version } from "../version.ts";
+
+declare global {
+  namespace GraphileBuild {
+    interface Inflection {
+      /**
+       * The signature to append to a function's resource name to
+       * disambiguate it from its overloads, based on the names of its input
+       * argument types.
+       */
+      functionResourceNameSignature(
+        this: Inflection,
+        details: {
+          serviceName: string;
+          pgProc: PgProc;
+        },
+      ): string;
+    }
+  }
+
+  namespace GraphileConfig {
+    interface Plugins {
+      PgFunctionOverloadsPlugin: true;
+    }
+  }
+}
+
+export const PgFunctionOverloadsPlugin: GraphileConfig.Plugin = {
+  name: "PgFunctionOverloadsPlugin",
+  description:
+    "Factors the input argument types into function resource names so that overloaded functions receive distinct names rather than being skipped",
+  version: version,
+
+  inflection: {
+    replace: {
+      functionResourceName(previous, options, details) {
+        if (!previous) {
+          throw new Error(`No functionResourceName inflector found!`);
+        }
+        return `${previous(details)}${this.functionResourceNameSignature(
+          details,
+        )}`;
+      },
+    },
+    add: {
+      functionResourceNameSignature(_options, { pgProc }) {
+        return pgProc
+          .getArguments()
+          .filter((a) => a.isIn)
+          .map((a) => "__" + a.type.typname)
+          .join("");
+      },
+    },
+  },
+};
+
+export const PgFunctionOverloadsPreset: GraphileConfig.Preset = {
+  plugins: [PgFunctionOverloadsPlugin],
+};
--- a/postgraphile/postgraphile/__tests__/function-overloads-schema.sql
+++ b/postgraphile/postgraphile/__tests__/function-overloads-schema.sql
@@ -1,0 +1,27 @@
+-- Test overloaded computed column functions targeting different tables
+drop schema if exists function_overloads, function_overloads_other_schema cascade;
+
+create schema function_overloads;
+create schema function_overloads_other_schema;
+create table function_overloads.pets (id serial primary key, name text);
+create table function_overloads.buildings (id serial primary key, address text);
+
+-- Two overloaded functions in the SAME schema as their target tables
+create function function_overloads.code(function_overloads.pets) returns text
+  as $$ select 'P' || $1.id::text; $$ language sql stable;
+create function function_overloads.code(function_overloads.buildings) returns text
+  as $$ select 'B' || $1.id::text; $$ language sql stable;
+comment on function function_overloads.code(function_overloads.pets)
+  is E'@behavior +typeField -queryField';
+comment on function function_overloads.code(function_overloads.buildings)
+  is E'@behavior +typeField -queryField';
+
+-- Cross-schema computed column functions (different schema from target tables)
+create function function_overloads_other_schema.age(function_overloads.pets) returns int
+  as $$ select 42; $$ language sql stable;
+create function function_overloads_other_schema.age(function_overloads.buildings) returns int
+  as $$ select 99; $$ language sql stable;
+comment on function function_overloads_other_schema.age(function_overloads.pets)
+  is E'@behavior +typeField -queryField';
+comment on function function_overloads_other_schema.age(function_overloads.buildings)
+  is E'@behavior +typeField -queryField';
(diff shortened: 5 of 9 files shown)


graphile-build/graphile-build-pg/src/plugins/PgProceduresPlugin.ts
```diff
@@ -214,10 +235,68 @@ export const PgProceduresPlugin: GraphileConfig.Plugin = {
...
+                      break;
+                    }
+                  }
+                  if (className && schemaName) {
+                    name = `${name}_${schemaName}_${className}`;
+                  }
+                }
+              }
+            }
+          }
```
benjie: This logic should be moved to the `functionResourceName` inflector; otherwise other parts of the code that use the inflector to determine the name of the function (e.g. so they can read it from the `pgRegistry`) will not get the same name and might extract the wrong function.

Further, renaming a function when it gets an overload makes me uncomfortable... ~~But I think we'll allow it because the alternatives seem worse.~~

Though having said this... the inflector shouldn't need to look through the rest of the registry to determine what something is called... so... Yeah this is a non-trivial issue.

~~If we're doing opt-in for this feature, I think just renaming all procedures so they're named by their signature (`${fn.name}_${params.map(param => param.type.name).join('_')}`) might be the best bet. Or at least all functions that look like computed columns.~~

Actually we already have a naming convention that encourages `${tableName}_${fieldName}` as a name - so I think we can fix this by just enforcing that:

- If this is a "computed column function" (stable/immutable, first argument is composite, belongs to same schema as composite type), then:
  - If the function name starts with the composite type name followed by underscore, use current logic;
  - Otherwise, _pretend_ the function was named such that it started with the composite type name followed by an underscore, and the follow current logic.

If we follow this, the vast majority of PostGraphile users won't be impacted at all, and only functions that don't follow our established naming conventions will be renamed, and they'll be renamed whether or not they are overloads which means adding an overload later won't break/rename them.

Yes, this feels much better.

graphile-build/graphile-build-pg/src/plugins/PgProceduresPlugin.ts
```diff
@@ -648,18 +727,65 @@ export const PgProceduresPlugin: GraphileConfig.Plugin = {
...
+              p.pronamespace === pgProc.pronamespace &&
+              p.proname === pgProc.proname &&
+              p._id !== pgProc._id
+            ) {
+              hasOverload = true;
+              break;
+            }
+          }
+          if (hasOverload) {
+            // This proc has overloads — check if it qualifies as a computed
```
benjie: Seems unnecessary to have rewritten this? Does it need a naked block? In general I'm not a fan of `let` either, `const` should be preferred where possible, and in this case the performance cost of `find` vs `for` isn't a major concern.

graphile-build/graphile-build-pg/src/plugins/PgProceduresPlugin.ts
```diff
@@ -648,18 +727,65 @@ export const PgProceduresPlugin: GraphileConfig.Plugin = {
...
+        // allow overloads where each targets a distinct composite type as its
+        // first argument (i.e. computed columns on different tables), since
+        // each will be exposed on a different GraphQL type.
+        {
+          let hasOverload = false;
+          for (const p of introspection.procs) {
+            if (
+              p.pronamespace === pgProc.pronamespace &&
+              p.proname === pgProc.proname &&
+              p._id !== pgProc._id
```
benjie: The issue we're really trying to avoid here is when the multiple (overloaded) functions end up with the same name in inflection. Maybe we should encode that specifically as the rule and skip the remaining complexity?

```suggestion
              p.pronamespace === pgProc.pronamespace &&
              p.proname === pgProc.proname &&
              p._id !== pgProc._id &&
              // Check if name conflicts:
              info.inflection.functionResourceName({
                serviceName,
                pgProc: p,
              }) === name
```

postgraphile/postgraphile/__tests__/kitchen-sink-schema.sql
```diff
@@ -2574,6 +2574,16 @@ create table cjk."期间" (
 
 --------------------------------------------------------------------------------
 
+-- Test overloaded computed column functions targeting different tables.
+create table a.pets (id serial primary key, name text);
+create table a.buildings (id serial primary key, address text);
+create function a.code(a.pets) returns text as $$ select 'P' || $1.id::text; $$ language sql stable;
+create function a.code(a.buildings) returns text as $$ select 'B' || $1.id::text; $$ language sql stable;
+comment on function a.code(a.pets) is E'@behavior +typeField';
+comment on function a.code(a.buildings) is E'@behavior +typeField';
```
benjie: The reason the snapshots have changed so much is you've added stuff to one of our most used schemas (`a`); please instead create a new schema ("function_overloads") and create these elements in there.

Also, you may need to change the behavior to `+typeField -queryField` - I can't remember. Check in the generated schema for unexpected changes (e.g. this function showing up twice or causing a clash).

benjie (changes requested):
Thanks for the contribution; here's some guidance on how to help move this forward. I think we can end up with a much smaller PR that still achieves your goals 🤞

postgraphile/postgraphile/__tests__/kitchen-sink-schema.sql
benjie: The `__tests__/schema/...` is where you'd add a schema test for this. You can use the defaultOptions test as a template, just change the schema name.

graphile-build/graphile-build-pg/src/plugins/PgProceduresPlugin.ts
GeertJohan: > belongs to same schema as composite type

Although for our use-case we actually place all computed columns in a schema specifically for that purpose. This creates nice explicit SQL, e.g.:

`SELECT id, computed.code(orders.*) FROM sales.orders;`

This works fine already, except for computed columns that are overloaded. e.g. if we would add a `computed.code(orderlines.*)`, then suddenly the `computed.code(orders.*)` computed column disappears.

So I would prefer to have cross-schema computed columns remain possible with this approach as well, I'm working on a diff for this PR right now :)

GeertJohan: I see this would mean existing functions that take a compound argument will change name.. That's not very nice as it's a breaking change. On the other hand the current behavior where `computed.code(orders.*)` works fine as long as there is no second `computed.code(?)` is not very nice either..

Should we accept the breaking change in favor of consistent behavior? Or perhaps be looking at a different solution/direction?

GeertJohan:
I have pushed changes to address a number of comments. It also moves towards the consistent prefix behavior, but that does break existing functions (see the thread).

In general I'm strictly against breaking changes, however since we are still pre 5.0 stable and since it makes behavior more consistent across the board, I think it may be a good direction. The breaking change will probably have a very limited impact since it's a rare feature?

GeertJohan:
As a result of these changes, the `query_compound_type_array` is now renamed to `compound_type_query_compound_type_array`. This is the breaking change we're talking about.

I've made the test succeed now. But if we think this is the good way forward then it may make sense to rename `query_compound_type_array` to `query_array` (which is then renamed to `compound_type_query_array` to make the tests a little easier to read.

postgraphile/postgraphile/__tests__/kitchen-sink-schema.sql
GeertJohan: Fixed, thanks for the help!

graphile-build/graphile-build-pg/src/plugins/PgProceduresPlugin.ts
benjie: We cannot accept a breaking change at this point; but it can be handled through a plugin/preset.

GeertJohan: Ah ofcourse, good idea!

I have reverted the existing tests to their original state so we don't have breaking changes any more.
Ive extracted a small part of the `functionResourceName` into an inflector which can be overridden by plugin/preset. Does it make sense to do it like this? Or should we move the whole `functionResourceName` to be plugin-overridable?

Do you want me to add a global plugin/preset that's available to the user or just leave that implementation to the user.

I did add a warning log when a computed column is thrown out when it is overloaded. I think it is warranted since, from the user's perspective, its confusing that a field on resource A suddenly disappears when an overload for resource B is added.

graphile-build/graphile-build-pg/src/plugins/PgProceduresPlugin.ts
```diff
@@ -41,6 +41,13 @@ declare global {
           pgProc: PgProc;
         },
       ): string;
+      functionResourceNameShouldPrefixCompositeType(
+        this: Inflection,
+        details: {
+          pgProc: PgProc;
+          firstArgCompositeType: PgType;
+        },
+      ): boolean;
```
benjie, line 50: Inflectors _always_ return `string`, so this isn't an inflector.

Instead, use something like `functionResourceNameCompositeTypePrefix(pgProc)` which would return the empty string `""` for the falsy path, and the prefix to use (e.g. `firstArg.type.typname + "_"`) for the truthy path. Then you would compose it in as normal:

```ts
        const computedPrefix = functionResourceNameCompositeTypePrefix(pgProc);
        return `${schemaPrefix}${computedPrefix}${pgProc.proname}`;
```

Also, `firstArgCompositeType` can be derived directly from `pgProc` and so should not be passed.

benjie (commented):
Thanks for the patience whilst we were working through the V5 release!

graphile-build/graphile-build-pg/src/plugins/PgProceduresPlugin.ts
GeertJohan: Oops, I've pushed the change.

graphile-build/graphile-build-pg/src/plugins/PgProceduresPlugin.ts
```diff
@@ -648,17 +692,39 @@ export const PgProceduresPlugin: GraphileConfig.Plugin = {
...
-            p._id !== pgProc._id,
+            p._id !== pgProc._id &&
+            inflection.functionResourceName({
+              serviceName,
+              pgProc: p,
+            }) ===
+              inflection.functionResourceName({
+                serviceName,
+                pgProc,
+              }),
```
benjie: This doesn't need to be recomputed each time, let's hoist it up into a `const` above `const overload =`

graphile-build/graphile-build-pg/src/plugins/PgProceduresPlugin.ts
```diff
@@ -648,17 +692,39 @@ export const PgProceduresPlugin: GraphileConfig.Plugin = {
...
 
-        // We also don’t want procedures that have been defined in our namespace
-        // twice. This leads to duplicate fields in the API which throws an
-        // error. In the future we may support this case. For now though, it is
-        // too complex.
+        // We don’t want procedures whose inflected resource name clashes
+        // with another overload — this would produce duplicate fields.
+        // Overloads targeting distinct composite types get unique names
+        // from the inflector (e.g. pets_code vs buildings_code).
         const overload = introspection.procs.find(
```
benjie: Probably rename this to `forbiddenOverload`?

graphile-build/graphile-build-pg/src/plugins/PgProceduresPlugin.ts
```diff
@@ -123,14 +140,38 @@ export const PgProceduresPlugin: GraphileConfig.Plugin = {
...
+        const firstArg = pgProc.getArguments().find((a) => a.isIn);
+        if (!firstArg || firstArg.type.typtype !== "c") {
+          return "";
+        }
+        // By default, only consider a function as a computed column if it
+        // belongs to the same schema as the composite type. Override this
+        // inflector to support cross-schema computed columns.
+        if (firstArg.type.typnamespace !== pgProc.pronamespace) {
+          return "";
+        }
```
benjie: I'm not keen that we mix "will this be a computed column" logic into the inflector for a resource. Whether it's a computed column or not is the result of behaviors, and inflection doesn't have access to that.

Might a simpler solution be to offer an extra preset that overrides `functionResourceName` and appends the types of the input arguments, unilaterally, enabling overloads throughout? This'll then have knock-on consequences that'll likely have to be dealt with, but pushing conflicts from the pgRegistry down into schema build time seems more sensible - the registry should be able to represent everything in the DB, even if the schema can't make use of it.

```ts
const PgFunctionOverloadsPreset: GraphileConfig.Preset = {
  inflection: {
    replace: {
      functionResourceName(previous, options, details) {
        if (!previous) throw new Error(`No functionResourceName inflector found!`);
        return `${previous(details)}${this.functionResourceNameSignature(details)}`;
      },
    },
    add: {
      functionResourceNameSignature(_options, { pgProc }) {
        return pgProc.getArguments().filter(a => a.isIn).map(a => "__" + a.type.typname).join("");
      },
    },
  },
};
```

graphile-build/graphile-build-pg/src/plugins/PgProceduresPlugin.ts
```diff
@@ -648,17 +692,39 @@ export const PgProceduresPlugin: GraphileConfig.Plugin = {
...
+          // inflector produced the same name (e.g. cross-schema overloads).
+          const thisFirstArg = pgProc.getArguments().find((a) => a.isIn);
+          const otherFirstArg = overload.getArguments().find((a) => a.isIn);
+          if (
+            thisFirstArg?.type.typtype === "c" &&
+            otherFirstArg?.type.typtype === "c" &&
+            thisFirstArg.type._id !== otherFirstArg.type._id
+          ) {
+            console.warn(
+              `Skipping function '${namespace!.nspname}.${pgProc.proname}' because its resource name clashes with an overload. Consider overriding the 'functionResourceNameCompositeTypePrefix' inflector to support cross-schema computed columns.`,
```
benjie:
```suggestion
              `Skipping function '${namespace!.nspname}.${pgProc.proname}' because it has overloads and they generate the same name. Consider using 'PgFunctionOverloadsPreset' to factor argument types into resource names.`,
```

benjie (commented):
Not the simplest of problems to solve 😅 

Let me know what you think of the ideas below.

GeertJohan:
Hi @benjie, I re-did the feature based on your feedback and based on the latest main (this branch was 474 commit behind :sweat_smile: ).

AI Disclaimer: I used Claude Fable 5 while making these changes.

PR #3025, opened 2026-04-10
Fix @skip/@include on fragment spreads in abstract positions

benjie:
Due to an oversight, `@skip` and `@include` when used on a named or inline fragment spread in an abstract position were previously ignored; this fixes the issue.

### DIFF
--- a/grafast/grafast/src/graphqlMergeSelectionSets.ts
+++ b/grafast/grafast/src/graphqlMergeSelectionSets.ts
@@ -6,6 +6,7 @@ import type {
 } from "graphql";
 import * as graphql from "graphql";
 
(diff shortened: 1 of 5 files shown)


grafast/grafast/src/graphqlMergeSelectionSets.ts
```diff
@@ -46,6 +47,9 @@ export function fieldSelectionsForType(
 ): ReadonlyArray<FieldNode> {
   for (let i = 0, l = selections.length; i < l; i++) {
     const selection = selections[i];
+    if (selectionIsSkipped(operationPlan, selection)) {
+      continue;
+    }
```
benjie, line 52: ^ This is the fix

PR #3029, opened 2026-04-14
feat: add multi-tenancy optimization capability

Zetazzz:


benjie (commented):
Feels like the LLM did a lot of work just to tell you what I would: use a single PostGraphile instance with unqualified identifiers and change the runtime schema via search_path.

If doing so does not fit your needs, I’d love to hear why.

PR #3034, opened 2026-04-21
Add "unknown" codec

benjaie:
When porting from V4, `sql.value(foo)` is typically replaced by `$pgSelect.placeholder($foo, TYPES.bar)`. The issue is, we don't always know what `bar` is. This PR makes it so we can use `TYPES.unknown` until the actual type is known, making it obvious to the user to return and refactor, whilst maintaining equivalence with V4 code.

### DIFF
--- a/grafast/dataplan-pg/src/codecs.ts
+++ b/grafast/dataplan-pg/src/codecs.ts
@@ -1107,6 +1107,7 @@ const reg = { hasNaturalOrdering: false };
  * names (those that would be found in the `pg_type` table).
  */
 export const TYPES = {
+  unknown: t<SQLRawValue>()("705", "unknown"), // unknown: 705
   void: t<void>()("2278", "void"), // void: 2278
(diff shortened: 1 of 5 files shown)


grafast/dataplan-pg/src/codecs.ts
```diff
@@ -1101,12 +1105,47 @@ const stripSubnet32 = {
...
+  }
+}
+
 /**
  * Built in PostgreSQL types that we support; note the keys are the "ergonomic"
  * names (like 'bigint'), but the values use the underlying PostgreSQL true
  * names (those that would be found in the `pg_type` table).
  */
 export const TYPES = {
+  unknown: t<SQLRawValue>()("705", "unknown", {
```
benjaie, line 1141:
```suggestion
  /** Experimental support for "unknown" data on input. @experimental */
  unknown: t<SQLRawValue>()("705", "unknown", {
```

benjie (approved):

PR #3131, opened 2026-08-07
docs: Edits for inhibition explanations

jemgillam:
## Description

Here are some edits to the Grafast documentation to make it clear that inhibition should not be relied on for security needs.

### DIFF
--- a/grafast/website/grafast/standard-steps/inhibitOnNull.mdx
+++ b/grafast/website/grafast/standard-steps/inhibitOnNull.mdx
@@ -37,9 +37,7 @@ adjusted downwards to match. Once the step has executed, the result has these
 indices re-populated (with an inhibited value) such that the batch size remains
 consistent with other steps in the same _layer plan_.
 
-With the current implementation, if all indices are inhibited then the step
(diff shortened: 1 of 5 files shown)


grafast/website/grafast/standard-steps/inhibitOnNull.mdx
```diff
@@ -28,24 +37,24 @@ adjusted downwards to match. Once the step has executed, the result has these
 indices re-populated (with an inhibited value) such that the batch size remains
 consistent with other steps in the same _layer plan_.
 
-Should all indices be inhibited, the step will not execute at all.
+With the current implementation, if all indices are inhibited then the step
+does not execute. This is an implementation detail, not a guarantee on which
+security or correctness may depend.
```
benjie: This one is actually correct. If all of the runtime values are inhibited then the step won't run. This doesn't necessarily correlate with the plan-time behavior looking like it _will_ inhibit or not.

```suggestion
Should all indices be inhibited, the step will not execute at all.
```

benjie (approved):
Thanks!

PR #3143, opened 2026-08-22
Respect `pgSettingsForIntrospection` for enum tables

nick-kang:
## Summary

`PgIntrospectionPlugin` applies `pgSettingsForIntrospection` during PostgreSQL
catalog introspection, but `PgEnumTablesPlugin` currently reads enum table values
without those settings.

This causes enum table discovery to run as the connection role rather than the
configured introspection role. It fails when the connection role intentionally
has no direct table access and must switch roles for introspection.

This change:

- applies `pgSettingsForIntrospection` when reading enum table values
- applies the same settings to the diagnostic query after a failed read
- adds a focused regression test
- adds a patch changeset for `graphile-build-pg`

## Testing

- `yarn build`
- `yarn workspace graphile-build-pg build`
- `yarn workspace graphile-build-pg test --runInBand`
- ESLint and Prettier checks for the changed files

### DIFF
--- a/.changeset/calm-enums-introspect.md
+++ b/.changeset/calm-enums-introspect.md
@@ -1,5 +1,6 @@
 ---
 "graphile-build-pg": patch
+"postgraphile": patch
 ---
 
 Apply `pgSettingsForIntrospection` when reading enum table values.
--- a/.changeset/calm-enums-introspect.md
+++ b/.changeset/calm-enums-introspect.md
@@ -1,0 +1,5 @@
+---
+"graphile-build-pg": patch
+---
+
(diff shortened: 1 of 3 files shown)


benjie (dismissed):
Thanks!

.changeset/calm-enums-introspect.md
```diff
@@ -0,0 +1,5 @@
+---
+"graphile-build-pg": patch
```
benjie, line 2:
```suggestion
"graphile-build-pg": patch
"postgraphile": patch
```

PR #3144, opened 2026-08-26
Improve support for generated types

benjie:
A single Node instance might run more than one PostGraphile instances; and similarly multiple schemas might exist in the same TypeScript scope (e.g. in your editor). For the types to be reliable and consistent, we need them to be scoped to each schema.

This PR makes it so that the generated types need to be 'scoped' to be used. The scope generally defaults to `"default"`, so people only dealing with a single PostGraphile instance get an easier time.


grafast/grafast/src/interfaces.ts
```diff
@@ -234,11 +234,10 @@ export type FieldArgs<TObj extends BaseGraphQLArguments = any> = {
...
     ([unknown] extends [TObj[key]]
       ? { [subkey in string as `$${subkey}`]: Step<any> }
-      : TObj[key] extends Record<string, any>
+      : NonNullable<TObj[key]> extends Record<string, any>
         ? {
-            [subkey in keyof TObj[key] & string as `$${subkey}`]: Step<
-              TObj[key][subkey]
-            >;
+            [subkey in keyof NonNullable<TObj[key]> &
+              string as `$${subkey}`]: Step<NonNullable<TObj[key]>[subkey]>;
```
benjie: This should maybe factor in the nullability of TObj[key]

PR #3146, opened 2026-08-28
Make inlining smarter and allow users to override behavior

benjie:
`$pgSelect.setInliningStrategy("preferSubquery")` etc


grafast/dataplan-pg/src/steps/pgSelect.ts
```diff
@@ -1397,12 +1437,12 @@ export class PgSelectStep<
         sql.isEquivalent(a, b, options);
 
       // Check trusted matches
-      if (p.trusted !== this.trusted) {
```
benjie, line 1400: Bugfix

PR #3168, opened 2026-09-11
support `@argNdescription` to document  arguments in GraphQL

### LINKED ISSUE #1419: Allow do document function parameters via smart-comments
### Feature description

Support to document the function parameters on graphql.

* currently it only lists the parameters
* defaults are not shown in the documentation
* no difference between required parameters (those not having a default) and optional parameters (those having a default)

### Motivating example

currently it seems there is no way to document function parameters using smart comments

```
create or replace function create_chart_event(
        title text,
        chart_size int,
        submission_start timestamptz default null,
    )
    returns foo as
$$
    ...
$$ language sql volatile;

comment on function create_chart_event is
  E'Creates a chart event where users can vote on something';

```

it would help users of the api  if we could document each parameter.

eg:

```
comment on parameter create_chart_event.submission_start is
  E'Datetime users can start submitting their votes, defaults to now if not given';
```

unfortunately it seems postgres does not support comments on function parameters.

so something like this might work (similar to [jsonfield](https://www.graphile.org/postgraphile/smart-comments/#smart-comment-spec)):

```
comment on function create_chart_event is
  E'Creates a chart event where users can vote on something\n'
  '@param submission_start Datetime users can start submitting their votes, defaults to now if not given';
```

also, in the above example title and chart_size should be recognized as required inputs whereas submission_start is optional

```
type CreateChartEventInput {
  title: String!
  chartSize: Int!

  """
  Datetime users can start submitting their votes, defaults to now if not given
  """
  submissionStart: Datetime
}
```

### Supporting development

I [tick all that apply]:

- [ ] am interested in building this feature myself
- [ ] am interested in collaborating on building this feature
- [x] am willing to help testing this feature before it's released
- [ ] am willing to write a test-driven test suite for this feature (before it exists)
- [ ] am a [Graphile sponsor](https://www.graphile.org/sponsor/) ❤️
- [ ] have an active [support or consultancy contract](https://www.graphile.org/support/) with Graphile

tsnobip:
Fixes #1419

## Performance impact

Impact should be low, it just adds support for a new smart tag.

## Security impact

Same

## Checklist

### DIFF
--- a/.changeset/describe-function-arguments.md
+++ b/.changeset/describe-function-arguments.md
@@ -1,0 +1,7 @@
+---
+"graphile-build-pg": minor
+---
+
+Add `PgArgumentDescriptionsPlugin`, loaded by default, to apply
+`@argNdescription` smart tags to SQL function arguments and mutation input
+fields.
--- a/graphile-build/graphile-build-pg/src/plugins/PgArgumentDescriptionsPlugin.ts
+++ b/graphile-build/graphile-build-pg/src/plugins/PgArgumentDescriptionsPlugin.ts
@@ -1,0 +1,42 @@
+import "graphile-config";
+
+import { version } from "../version.ts";
+
+declare global {
+  namespace GraphileConfig {
+    interface Plugins {
+      PgArgumentDescriptionsPlugin: true;
+    }
+  }
+}
+
+export const PgArgumentDescriptionsPlugin: GraphileConfig.Plugin = {
+  name: "PgArgumentDescriptionsPlugin",
+  description:
+    "Applies zero-based @argNdescription tags to SQL function inputs.",
+  version,
+  gather: {
+    hooks: {
+      pgProcedures_PgResourceOptions(_info, { pgProc, resourceOptions }) {
+        const { parameters } = resourceOptions;
+        if (!parameters) return;
+        const { tags } = pgProc.getTagsAndDescription();
+        const types = pgProc.proallargtypes ?? pgProc.proargtypes ?? [];
+        let inputIndex = 0;
+        types.forEach((_type, index) => {
+          const mode = pgProc.proargmodes?.[index] ?? "i";
+          if (mode !== "i" && mode !== "b") return;
+          const param = parameters[inputIndex++];
+          const tag = tags[`arg${index}description`];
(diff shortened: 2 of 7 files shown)


.changeset/describe-function-arguments.md
```diff
@@ -0,0 +1,7 @@
+---
+"graphile-build-pg": minor
```
benjie, line 2: Lets make this show up in PostGraphile's changelog explicitly

```suggestion
"graphile-build-pg": minor
"postgraphile": minor
```

.changeset/describe-function-arguments.md
```diff
@@ -0,0 +1,7 @@
...
+---
+
+Add `PgArgumentDescriptionsPlugin`, loaded by default, to apply
+`@argNdescription` smart tags to SQL function arguments and mutation input
+fields.
```
benjie, line 7: I'm not sure the specific plugin is particularly important to most users, it's the functionality they care about:

```suggestion
Introduce `@argNDescription` smart tag to allow documentation of PostgreSQL
function arguments, where N is the zero-based index of the argument.
```

graphile-build/graphile-build-pg/src/plugins/PgArgumentDescriptionsPlugin.ts
```diff
@@ -0,0 +1,42 @@
...
+          const mode = pgProc.proargmodes?.[index] ?? "i";
+          if (mode !== "i" && mode !== "b") return;
+          const param = parameters[inputIndex++];
+          const tag = tags[`arg${index}description`];
+          const description = Array.isArray(tag) ? tag.join("\n") : tag;
+          if (typeof description === "string") {
+            param.extensions = {
+              ...param.extensions,
+              argDescription: description,
+            };
```
benjie, line 36: We might as well hoist 'description' to be a first-class citizen.

```suggestion
            param.description = argDescription;
```

(This will require `PgResourceParameter` to gain a `description?: string` property.)

benjie (changes requested):
Nice; thanks for taking this on! Since this is a 'minor', please target branch `next` (all the PostGraphile 5.2 stuff). The main change is to move description to an explicit PgResourceParameter field rather than an extension.

PR #3174, opened 2026-09-16
Type functional resource execution results

benjaie:
## Description

Add result-shape typing for functional PostgreSQL resources.

`PgResource.execute()`, `executePositional()`, and `executeNamed()` previously exposed a broad `Step` result even when a resource’s cardinality was known. Typed registries can now preserve the concrete execution shape:

- single-result functions return `PgSelectSingleStep`
- set-returning functions and functions returning `foo[]` return `PgSelectStep`
- partitioned `setof foo[]` functions return `PgPartitionedSelectStep`

The result kind is carried through `PgResourceOptions` and applied when constructing `PgRegistry["pgResources"]`, without narrowing graphile-build-pg’s runtime procedure-discovery types.

## Performance impact

None; this is type-level only.

## Security impact

None.

## Checklist


.changeset/typed-resources-execute.md
```diff
@@ -0,0 +1,5 @@
+---
+"@dataplan/pg": patch
+---
+
+Expose the specific plan shape returned by generated functional resources when executing them.
```
benjie:
```suggestion
`PgResource.execute()` now infers result type from underlying function signature. This has added two more generics to `PgResource` and `PgResourceOptions`, so you may need to update your code if you are explicitly stating these generics.
```

grafast/dataplan-pg/src/datasource.ts
```diff
@@ -1070,6 +1148,7 @@ export class PgResource<
 }
 exportAs("@dataplan/pg", PgResource, "PgResource");
 
+/** A resource with the result shape of its functional execution preserved. */
```
benjie:
```suggestion
```

benjie (changes requested):
Pretty close

benjie (dismissed):
Ace 👍

benjie (approved):

